PR #4, opened 2019-05-03
Minimal build setup

fjetter:
This should set up the initial travis build pipeline. Integration tests with pandas,dask,arrow master, etc will follow later


README.md
```diff
@@ -1,6 +1,7 @@
 kartothek
 =========
 
+[![Build Status](https://travis-ci.org/fjetter/kartothek.svg?branch=master)](https://travis-ci.org/fjetter/kartothek)
```
lr4d: not `fjetter`

setup.py
```diff
@@ -21,8 +21,13 @@ def setup_package():
         classifiers=[
             "Development Status :: 5 - Production/Stable",
             "Programming Language :: Python",
+            'Programming Language :: Python :: 3',
+            'Programming Language :: Python :: 3.6',
```
xhochy: What about 3.7?

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: python
+
+python:
+  - 3.6
```
xhochy: Can you add `3.7` to the matrix?

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+
+install:
+  - pip install --upgrade -r requirements.in
```
xhochy: Is there a reason to make this three different calls all with `--upgrade`?  Furthermore, it is good practice to create a `virtualenv` and not use the system python environment eventhough you only require a single env.

setup.py
fjetter: The title says _minimal_ ;)

I'm currently setting up a proper build matrix. I'll update this once it's done

.travis.yml
fjetter: Well, according to https://docs.travis-ci.com/user/languages/python/#travis-ci-uses-isolated-virtualenvs I am running in a dedicated freshly created `virtualenv` and *not* in the system env. That being said, I figured that I use the simple way until we hit any issues to keep our build setup as simple as possible until we hit any issues

.travis.yml
fjetter: done

fjetter:
There seems to be an issue with `dask==1.2.1` and `pyarrow==0.13.0`. Couldn't reproduce locally, yet but I'll have a look this afternoon and file an issue in whatever tracker which seems to be appropraite

lr4d:
While I am all for improving the build setup, I think it'd be nice to stick with _minimal_ here, as we'll want to check that other PR's pass the tests soon.

fjetter:
Turns out the `dask` `pyarrow` incompatibility is intentional, see https://github.com/dask/dask/pull/4695

I mark the test as xfail int his circumstance since our functionality is not impacted by this

ci/check_format.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
xhochy: What about using `pre-commit run -a`. This generates a much nicer output with the same effect.

fjetter: Good point, didn't know about this. Will remove `check_format.sh` in favor of `pre-commit run -a`. This will also help with synchronization between the files

fjetter:
@xhochy I added a deploy section and exchanged the formatting. I'd say this suffices for now. Changing to our own venv, etc is probably necessary once we build pyarrow, pandas, etc but for now we require working builds :)

.travis.yml
```diff
@@ -0,0 +1,41 @@
...
+  - pytest
+
+deploy:
+  provider: pypi
+  user: fjetter
+  password:
+    secure: akyeNjwXBsalgbqtPJkcEVaUzjXCvc78UZavjF3e8MuPefoGLzGnPdjh1vf/HX1msERem3WxKrzyqfmp7AMypXnzA5QuiZD81Zp55ypXmULBfm2zAkWCwdiej4mCAEBeOWrIihFPM7zBUNFKBG4/FzRYAlA1BJR7jSlHe4wZpWrfZ4hwTEbwH6J87YgDtiA7D9MCxewct5lecUMuLNVvaSRxt2rH1U6JyaExAtSJdMZhVqh8KNKfumxSoHrewDId6n2ITUKrks/TV4Ygm22+bIZ9dvpDcYpQq3Y+P56mTrHs6Eam1EHOxF11U/Vt4CLPuTr5eFMbRzEOnP+i+KbrIS6onWdLGQB98gtVrnIYYLwr+kLBowkVayyZTysw9FZO6niF0FaX+gw0XYhiqBX2pJPWcQDHHmfnRYrtFsofwgTFs5ux+cOhCiOC+PwPnfuUWzeiGgLrg5VQB5Gz2Hw2NRdUxd7l5zBTWyJqreKFCIGu+PEPY2abZeA1DKgdqd67nVm++d9PDDgufqRkooG4Y8UOr+7CIkV04nPtvIo4FpOI85D4DDBE3wWU8x2eEYQETH0kwNQkITmwaThsqW1ZVzqRi3dcFt1RX6gCMnziE/lcfw7vT2L7BgVbihF/KR7FDxi+6/9c7BYAEaUbsQbunpaIH5QdRN5xLJbWIgt4KD8=
+  on:
+    tags: true
+  distributions: bdist_wheel
```
xhochy: We need the `sdist` for `conda` packages.
```suggestion
  distributions: "sdist bdist_wheel"
```

fjetter:
The travis integration seems to be stuck for the PR build. Merging anyhow

PR #9, opened 2019-05-07
Extend core eager functionality

lr4d:
- Implemented `io.eager.update_dataset_from_dataframes`. 
Renamed `tests/io.eager.update` to `tests/io.eager.commit` and created new `tests/io.eager.update` file, which tests the `update_dataset_from_dataframes` function. 
Replaced a list comparison to set comparison in `io.testing.update`

- Implemented `io.eager.garbage_collect_dataset` (slight modification to `io_components.gc.dispatch_files_to_gc`).

- Implemented `io.eager.index.build_dataset_indices` and created `io.testing.index`

- Misc. improvements


kartothek/io/testing/index.py
```diff
@@ -0,0 +1,119 @@
...
+def index_dct_regulator(indices_dict):
+    """
+    Output
+    `kartothek.core.dataset.DatasetMetadataBase.indices.index_dct`
+    in standard format.
+    """
+    return [(k, sorted(v)) for k, v in sorted(indices_dict.items())]
+
+
+def index_dcts_are_equal(indices1, indices2):
```
fjetter: Use a `assert obj1 == obj2` instead of `assert are_objects_equal(obj1, obj2)` since `assert` may print better error messages

kartothek/io/testing/index.py
```diff
@@ -0,0 +1,119 @@
...
+
+from kartothek.core.factory import DatasetFactory
+from kartothek.core.index import ExplicitSecondaryIndex
+from kartothek.io.eager import (
+    read_dataset_as_metapartitions,
+    store_dataframes_as_dataset,
+)
+
+
+def index_dct_regulator(indices_dict):
```
fjetter: I would expect that the index object equal implementation would handle this already. Isn't this the case? What's missing? Maybe it makes sense to extend the index class with a better `equals` method

lr4d: We are comparing the `index_dct` attribute there. Given that the `ExplicitSecondaryIndex` object may look something like this:
```
ipdb> dataset_factory.indices                                                                     
ExplicitSecondaryIndex(index_storage_key=dataset_uuid/indices/p/2019-05-07T13%3A44%3A31.731906.by-dataset-index.parquet, column=p, dtype=int64, creation_time=2019-05-07 13:44:37.746877, index_dct=[1, 2, 3])
```
Attributes such as `creation_time` and `index_storage_key` are impossible to infer. Thus, I only compare `index_dct`.
We could either discard these time-based attributes in the index `__eq__` method, or (better) change the definition of the `index_dct` attribute to support unordered comparisons

lr4d: The most simple solution may be to just implement a getter method for `IndexBase` which returns the `index_dct` value in the same way as `index_dct_regulator`

fjetter: I'm hesitant to introduce a getter which sorts the attributes by default. These dictionaries and the lists as values may become quite large and sorting the entire dict is quite expensive

fjetter: Please change this to an `assert_index_dct_equal` s.t. we may benefit from better assertion messages. Sorting all dict values should be sufficient and allow python/pytest to use dict assertion messages which are, I believe, better readable than list assertion messages.

A simply way to do this could be done using https://toolz.readthedocs.io/en/latest/api.html#toolz.dicttoolz.valmap

```python
from toolz.dicttoolz import valmap
def assert_index_dct_equal(dict1, dict2):
    dict1 = valmap(dict1, sorted)
    dict2 = valmap(dict2, sorted)
    assert dict1 == dict2
```

kartothek/io_components/gc.py
```diff
@@ -43,8 +43,11 @@ def dispatch_files_to_gc(dataset_uuid, store_factory, chunk_size, factory):
 
     files_to_remove = list(remove_index_files | remove_table_files)
 
-    for i in range(0, len(files_to_remove), chunk_size):
-        yield files_to_remove[i : i + chunk_size]
+    if chunk_size is None:
```
fjetter: nitpick: you could set `chunk_size=len(files_to_remove)`. This way the `if` clause would be less verbose

kartothek/io/eager.py
```diff
@@ -612,3 +620,126 @@ def write_single_partition(
...
+        ds_factory=factory,
+        default_metadata_version=default_metadata_version,
+        partition_on=partition_on,
+        load_dynamic_metadata=load_dynamic_metadata,
+    )
+
+    secondary_indices = _ensure_compatible_indices(ds_factory, secondary_indices)
+
+    mp = parse_input_to_metapartition(
+        df_list, metadata_version=default_metadata_version
```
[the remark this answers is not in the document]
lr4d: Implemented this test in the latest commit

tests/conftest.py
```diff
@@ -97,6 +101,51 @@ def mock_uuid(mocker):
...
+
+    # Mock `kartothek.core.utils.verify_metadata_version`
+    def patched__verify_metadata_version(metadata_version):
+        return metadata_version
+
+    mocker.patch(
+        "kartothek.core.utils.verify_metadata_version", patched__verify_metadata_version
+    )
+    mocker.patch(
+        "kartothek.io_components.metapartition.verify_metadata_version",
```
fjetter: To avoid having to patch the same function multiple times, a common practice is to define a private function which can be the mock target.
```python
# utils.py
def _verify_metadata_version(...):
    """
    This is factored out to be an easier target for mocking
    """
def verify_metadata_version(...):
    return _verify_metadata_version(*args, **kwargs)

# conftest.py
def mock_version(mocker):
    mocker.patch(
	"kartothek.core.utils._verify_metadata_version", patched__verify_metadata_version
    )
    return mock_version
```

This way you can mock `_verify_metadata_version` once and this will easily propagate to all users of the actual function

lr4d: Done

PR #12, opened 2019-05-08
Add coverage report

fjetter:
This adds a coverage report and publishes on codecov.io

example: https://codecov.io/gh/fjetter/kartothek/branch/add_coverage


fjetter:
Coverage looks pretty bad at first glance but looking closer the worst thing is that we still have some dead code we didn't properly deprecated/remove. Core functionality which is actually used is covered pretty well

PR #15, opened 2019-05-09
Fix warnings and other quality-of-life improvements.

crepererum:
See individual commits for details.

### DIFF
--- a/kartothek/serialization/_generic.py
+++ b/kartothek/serialization/_generic.py
@@ -239,6 +239,12 @@ def filter_array_like(
     elif is_list_like(value) and len(value) and isinstance(value[0], date_type_to_cast):
         value = [pd.Timestamp(val).to_datetime64() for val in value]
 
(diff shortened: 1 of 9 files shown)


kartothek/serialization/_generic.py
```diff
@@ -239,6 +239,12 @@ def filter_array_like(
     elif is_list_like(value) and len(value) and isinstance(value[0], date_type_to_cast):
         value = [pd.Timestamp(val).to_datetime64() for val in value]
 
+    # NULL types might not be preserved well, so try to cast floats (pandas default type) to the value type
+    if np.issubdtype(array_like.dtype, np.floating) and np.isnan(array_like).all():
```
fjetter: Is this what was causing the warning of the numpy equals warning? Is this something we should add an explicit test for? I'm having issues estimating the impact of this

crepererum: The issue here are CSVs that have incorrect types when being recovered. So empty columns won't be `object` but `float`, which causes the warning.

PR #22, opened 2019-05-13
move config in `.isort.cfg` to `pyproject.toml` under the section `[tool.isort]`

lr4d:
Since https://github.com/timothycrosley/isort/issues/763 has been fixed


lr4d:
Will reformat this, wanted to take a look at the build status

lr4d:
It is necessary to install `isort[pyproject]` instead of plain `isort` to support `pyproject.toml` files (the `additional_dependecies` in `.pre-commit-config.yaml` serves the same purpose).
For the same reason, the version of `isort` had to be bumped.

Issue #26, opened 2019-05-14
Add type annotations


crepererum:
# Abstract
We use types in our documentation but it would be better to provide proper annotations. 

# Example
From:
```python
def add(a, b):
    """
    Adds to numbers.

    Parameters
    ----------
    a: Union[float, int]
        First number to add.
    b: Union[float, int]
        Second number to add.

    Returns
    -------
    c: float
        Sum of both.
    """
    ...
```

To:
```python
from typing import Union


def add(a: Union[float, int], b: Union[float, int]) -> float:
    """
    Adds to numbers.

    Parameters
    ----------
    a:
        First number to add.
    b:
        Second number to add.

    Returns
    -------
    c:
        Sum of both.
    """
    ...
```

Please note the colons and new lines after the parameter and return value names, otherwise, sphinx docs break (sphinx will treat your docs as types).

fjetter:
What should the scope of this issue be? set up the infrastructure or set up the annotations? If it's the latter we should restrict the scope, e.g. public `core` or public `io`, etc.

crepererum:
The infrastructure is set up in the linked PR. I would not like to limit the scope. The annotations are helpful externally (for other libs and for your IDE) and internally (consistency checks). The assigned person does not has to do everything at once once though. Sure, it's more a long term thing.

PR #27, opened 2019-05-14
set up mypy

crepererum:
Ref #26

### DIFF
--- a/kartothek/io/testing/read.py
+++ b/kartothek/io/testing/read.py
@@ -148,24 +148,24 @@ def _perform_read_test(
             assert isinstance(res, MetaPartition)
         result = [mp.data for mp in result]
 
-        def sort_by(dct):
-            return dct["core"].P.iloc[0]
+        def sort_by(obj):
+            return obj["core"].P.iloc[0]
 
     elif output_type == "table":
 
         assert isinstance(result[0], pd.DataFrame)
         assert "P" in result[0]
 
-        def sort_by(df):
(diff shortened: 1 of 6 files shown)


mypy.ini
```diff
@@ -0,0 +1,4 @@
+[mypy]
```
fjetter: A lot of these configs look useful. Did you have a look at these? If it spams to much we shouldn't use it https://mypy.readthedocs.io/en/stable/command_line.html#configuring-warnings

kartothek/io/testing/read.py
```diff
@@ -148,24 +149,24 @@ def _perform_read_test(
             assert isinstance(res, MetaPartition)
         result = [mp.data for mp in result]
 
-        def sort_by(dct):
-            return dct["core"].P.iloc[0]
+        def sort_by(obj: Any) -> Any:
```
fjetter: Don't have any experience with type annotations, yet. Does `Any` offer any value opposed to _not_ annotating?

crepererum: fixed. Mypy was complaining that the different variants of `sort_by` have different signatures and I've thought that this is because of the types. Turns out, they default to `Any` and the issue was that the parameters are named differently.

mypy.ini
crepererum: I think we can increase the warning thresholds in subsequent PRs. I think many of them need tiny code changes and I wanted to keep the very first setup basic / non-invasive.

Issue #28, opened 2019-05-15
Fill out type system docs


crepererum:
Placeholder for myself :)

PR #29, opened 2019-05-16
Arrow-based dict-type normalization

crepererum:
Discovered while working on #28.

Remove the hack we had for pre-0.10 pyarrow versions and implement
categorical-stripping in a proper way.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,19 +6,19 @@ Changelog
 Version 3.1.0 (2019-XX-XX)
 ==========================
 
+- fix ``getargspec`` ``DeprecationWarning``
+- fix ``FutureWarning`` in ``filter_array_like``
+- remove ``funcsigs`` requirement
 - Implement reference ``io.eager`` implementation, adding the functions:
     + ``io.eager.garbage_collect_dataset``
     + ``io.eager.index.build_dataset_indices``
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -2,14 +2,19 @@
...
-Version 3.0.1 (2019-XX-XX)
+Version 3.1.0 (2019-XX-XX)
 ==========================
 
 - fix ``getargspec`` ``DeprecationWarning``
 - fix ``FutureWarning`` in ``filter_array_like``
 - remove ``funcsigs`` requirement
 - fix ``_apply_partition_key_predicates`` ``FutureWarning``
 
+**Semi-breaking:**
```
fjetter: First of all, there is no such thing as a semi breaking change :) I think we should write up something about what we actually guarantee to be stable across releases and what not (e.g. specfication is tracked using metadata version, not the package version)

That being said. Is it actually breaking at all? Isn't the make_meta calls normalize_type for all columns so we just shifted the logic.

crepererum: @fjetter in our use case, it is not breaking but `normalize_type` is public API and it normalizes more types now so strictly speaking, this is a breaking change

fjetter: I'm just having issues with the term semi-breaking but that's not a showstopper.

PR #32, opened 2019-05-17
type system docs

### LINKED ISSUE #28: Fill out type system docs

crepererum:
Your Sunday-read is online :D

Closes #28


docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+- **efficiency:** Data written once should not be altered if not really necessary.
+- **stability:** Kartothek is made for production use. It should provide stable, predicable outputs and must not crash
+  or lead other libraries to behave in an expected way.
+- **compatibility:** Kartothek is not the only software in the stack and must play nicely with others (also see `Related
+  Type Systems`_).
+
+
+Base Type System
+----------------
+Since we think that `Arrow`_ is a solid, future-proof choice for in-memory dataframes will hopefully offer a sane
```
fjetter: for in-memory dataframes `<something is missing here>` will hopefully

fjetter: We should think about whether or not we want to refer to Apache Arrow instead of plain Arrow

cc @xhochy what's your official stance to topics like these?

docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+
+.. important::
+
+   `NaN payloads`_ are not handled and should not be used. The `IEEE 754`_ declares them as optinal and hardware and
+   software may wipe them anyway, so portable code cannot make use this data.
+
+For each of these categories, we can represent all values of ``float{16, 32}`` by using ``float64``. So we normalize all
+floating point types to ``float64``.
+
+Decimal
```
fjetter: Do we offer any decimal support?

docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+~~~~~~~
+Decimals have a given precision and scale and used to store fixed-point floats like money.
+
+The decimal type ``decimal128[P, S]`` (``P`` denoting the precision and ``S`` the scale) is not touched, since the
+exact precision and scale has a semantic meaning.
+
+Date
```
fjetter: We might want to add to this section, that there is no numpy array of this type which makes it slow for some use cases (since we're giving a very thorough intro to types overall)

docs/spec/type_system.rst
xhochy: The official stance is that you should use `Apache Arrow` but common usage is that you use this in headlines and once in your text and all following mentions are then simply `Arrow`.

docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+and ``O`` flags if the index is ordered or not.
+
+Since categoricals are, in our opinion, a pure optimization and do not alter the nature of the data, we treat
+dictionary-encoded data like the values they encode. So ``dictionary[T1, I1, O1]`` is compatible with ``T2`` if ``T1``
+and ``T2`` are compatible. This also means that it is compatible with ``dictionary[T2, I2, O2]``. Note that the ordered
+flag and the index data type are ignored. So the values in the example shown above are treated like ``string``.
+
+Normalization
+-------------
+Following the outlined guidelines, we can write down the following normalization rule set:
```
fjetter: Not sure if here is the best place but is a link to the arrow<->pandas type mapping helpful here? https://arrow.apache.org/docs/python/pandas.html#type-differences

docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+
+where :math:`\text{expbias} = 2^{\#\text{exponentbits}}-1`.
+
+.. important::
+
+   `Signaling NaN`_ values are discouraged and should not be used!
+
+.. important::
+
+   `NaN payloads`_ are not handled and should not be used. The `IEEE 754`_ declares them as optinal and hardware and
```
lr4d: typo at `optinal`

docs/spec/type_system.rst
```diff
@@ -4,5 +4,471 @@
...
+-------------
+Following the outlined guidelines, we can write down the following normalization rule set:
+
++------------------+------------------------------------+-----------------------------------------------------------+
+| Type Class       | Normalization                      | Examples                                                  |
++==================+====================================+===========================================================+
+| signed integer   | ``int{8, 16, 32, 64} -> int64``    | | ``norm(int8) = int64``                                  |
+|                  |                                    | | ``norm(int64) = int64``                                 |
++------------------+------------------------------------+-----------------------------------------------------------+
+| unsigned integer | ``uint{8, 16, 32, 64} -> uint64``  | | ``norm(int8) = int64``                                  |
```
lr4d: examples are `int`s

docs/spec/type_system.rst
crepererum: good point. I've double-checked the semantic and updated the section. Would advice against using them.

docs/spec/type_system.rst
crepererum: fixed both

docs/spec/type_system.rst
crepererum: added as note to the "base type system" section

crepererum:
I'll sqash the doc-content commits before merging.

docs/spec/type_system.rst
```diff
@@ -150,8 +150,38 @@ Decimal
 ~~~~~~~
 Decimals have a given precision and scale and used to store fixed-point floats like money.
 
-The decimal type ``decimal128[P, S]`` (``P`` denoting the precision and ``S`` the scale) is not touched, since the
-exact precision and scale has a semantic meaning.
+There is a single decimal type ``decimal128[P, S]`` where ``P`` measures the total precision in digits and ``S``
+measures the scale in digits (therefore :math:`P \ge S`).
+
+>>> from decimal import Context, Decimal
```
fjetter: Is this properly rendered as code?

fjetter:
Have a look at the decimal code example. Looks like it's not set up for code rendering. Otherwise LGTM

Thanks for putting in this work!

docs/spec/type_system.rst
fjetter: I realized that all code sections are not specially marked. Is this sphinx magic or an actual bug?

crepererum: Sphinx magic. Doctests don't require a marker.

crepererum: Screen Shot 2019-05-22 at 14 25 53

PR #39, opened 2019-05-22
Ensure column names are read as `str`

lr4d:
Pandas schema metadata saved using python2 stores column names as type `bytes` given that that is how Python2 implements strings by default.
When this schema metadata is loaded using Python3, `kartothek.core.common_metadata.empty_dataframe_from_schema`retrieves these column names as if they were `bytes` objects. However, the column names in Parquet are still strings, so when kartothek  attempts to retrieve these `bytes` object columns it return nulls as there is no data stored for `bytes` columns. 
AFAIK this only affects `io.eager.read_table`.

For `pyarrow>=0.13`, we can create an empty pandas dataframe without using this pandas metadata. By doing so, kartothek remains compatible with schemas created in Python2.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ Version 3.1.0 (2019-XX-XX)
 - fix ``_apply_partition_key_predicates`` ``FutureWarning``
 - serialize :class:`~kartothek.core.index.ExplicitSecondaryIndex` to parquet
 - improve messages for schema violation errors
+- Ensure binary column names are read as type ``str``:
+
+    - Ensure dataframe columns are of type ``str`` in :func:`~kartothek.core.common_metadata.empty_dataframe_from_schema`
+    - Testing: create :func:`~kartothek.io.testing.read.test_binary_column_metadata` which checks column names stored as
+      ``bytes`` objects are read as type ``str``
 
 **Breaking:**
 
--- a/kartothek/core/common_metadata.py
+++ b/kartothek/core/common_metadata.py
@@ -15,7 +15,7 @@ import simplejson
 
 from kartothek.core import naming
 from kartothek.core._compat import load_json
-from kartothek.core.utils import _ensure_string_type
+from kartothek.core.utils import ensure_string_type
 
 from ._compat import ARROW_LARGER_EQ_0130
 
@@ -789,7 +789,7 @@ def empty_dataframe_from_schema(schema, columns=None, date_as_object=False):
         )
         df = table.to_pandas(date_as_object=date_as_object)
 
-    df.columns = df.columns.map(_ensure_string_type)
+    df.columns = df.columns.map(ensure_string_type)
     if columns is not None:
         df = df[columns]
 
--- a/kartothek/core/common_metadata.py
+++ b/kartothek/core/common_metadata.py
@@ -15,6 +15,7 @@ import simplejson
 
 from kartothek.core import naming
 from kartothek.core._compat import load_json
+from kartothek.core.utils import _ensure_string_type
 
 from ._compat import ARROW_LARGER_EQ_0130
 
@@ -787,6 +788,8 @@ def empty_dataframe_from_schema(schema, columns=None, date_as_object=False):
             arrays=arrays, names=names, metadata=schema.metadata
         )
         df = table.to_pandas(date_as_object=date_as_object)
+
+    df.columns = df.columns.map(_ensure_string_type)
     if columns is not None:
         df = df[columns]
 
--- a/kartothek/core/utils.py
+++ b/kartothek/core/utils.py
@@ -24,3 +24,10 @@ def _verify_metadata_version(metadata_version):
 
 def verify_metadata_version(*args, **kwargs):
     return _verify_metadata_version(*args, **kwargs)
+
+
+def _ensure_string_type(obj):
+    if isinstance(obj, bytes):
+        return obj.decode()
+    else:
+        return str(obj)
--- a/kartothek/core/utils.py
+++ b/kartothek/core/utils.py
@@ -26,7 +26,22 @@ def verify_metadata_version(*args, **kwargs):
     return _verify_metadata_version(*args, **kwargs)
 
 
-def _ensure_string_type(obj):
+def ensure_string_type(obj):
+    """
+    Parse object passed to the function to `str`.
+
+    If the object is of type `bytes`, it is decoded, otherwise a generic string representation of the object is
+    returned.
+
+    Parameters
+    ----------
+    obj: Any
+        object which is to be parsed to `str`
+
+    Returns
+    -------
+    str_obj: String
+    """
     if isinstance(obj, bytes):
         return obj.decode()
     else:
--- a/kartothek/io/testing/read.py
+++ b/kartothek/io/testing/read.py
@@ -539,3 +539,31 @@ def test_datetime_predicate_with_dates_as_object(
         predicates=[[("c", ">=", datetime.datetime(2000, 1, 1))]],
         dates_as_object=True,
     )
+
+
+def test_binary_column_metadata(store_factory, bound_load_dataframes):
+    table_name = "core"
+    df = {
+        "label": "part1",
+        "data": [(table_name, pd.DataFrame({b"int_col": [1], "🙈".encode(): [2]}))],
+    }
+
+    store_dataframes_as_dataset(
+        dfs=[df], store=store_factory, dataset_uuid="dataset_uuid"
+    )
+
+    result = bound_load_dataframes(
+        dataset_uuid="dataset_uuid", store=store_factory, tables=table_name
+    )
+
+    probe = result[0]
+    if isinstance(probe, MetaPartition):
+        result_dfs = [mp.data[table_name] for mp in result]
+    elif isinstance(probe, dict):
+        result_dfs = [mp[table_name] for mp in result]
+    else:
+        result_dfs = result
+    df = pd.concat(result_dfs).reset_index(drop=True)
+
+    # Assert column names are of type `str`, instead of `bytes` objects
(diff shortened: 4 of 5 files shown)


fjetter:
We encountered this issue for a dataset which was originally created in PY2 and read in PY3. While this is not necessarily a common use case, especially not with the recent py3-only releases, I'm still wondering if we should drop the pandas metadata entirely when storing the schema. It is important when reading actual parquet files but I don't see value in storing this information in the `_common_metadata` 

cc @crepererum @xhochy

CHANGES.rst
```diff
@@ -15,6 +15,7 @@ Version 3.1.0 (2019-XX-XX)
     + ``io.eager.update_dataset_from_dataframes``
 - fix ``_apply_partition_key_predicates`` ``FutureWarning``
 - serialize :class:`~kartothek.core.index.ExplicitSecondaryIndex` to parquet
+- remove pandas schema metadata during empty dataframe creation for `pyarrow>=0.13`
```
fjetter: Try to express the change in behavior / the fixed issue in the changelog. the changelog is there to help the user understand what's going on without knowing the implementation. Without knowing the issue even I wouldn't know what the implications of removing this metadata would be.

xhochy:
> We encountered this issue for a dataset which was originally created in PY2 and read in PY3. While this is not necessarily a common use case, especially not with the recent py3-only releases, I'm still wondering if we should drop the pandas metadata entirely when storing the schema. It is important when reading actual parquet files but I don't see value in storing this information in the `_common_metadata`
> 
> cc @crepererum @xhochy

We might still need this metadata to reconstruct the pandas indices. I know that you're not a big fan of that but it is still used to my knowledge.

crepererum:
I agree that we should not drop the metadata because of the the Py2->3 migration. The columns ARE incompatible since under Py2, the columns can contain any byte sequence while under Python 3 they should be Unicode. So datasets under Python2 must be converted manually. Same btw. goes for `str` columns under Python2.

fjetter:
@xhochy I'm talking about the `_common_metadata` only. It is, by definition, a table w/out any rows, i.e. there is no index. For the actual parquet files this is a different story. There, the metadata is needed. For instance, the latest arrow stores a `pandas.RangeIndex` only in the metadata and no longer as a `__index_level_0__` field. We learned this the hard way when migrating to the new arrow version...

@crepererum it is not necessarily about py2/3 migrations. I'm just wondering if the metadata serves any actual value for the `_common_metadata` (not talking about file level metadata) and we might be better of (future compatibility) by not storing it at all.

The column types are the only attribute, I am aware of, we are storing in the pandas metadata of `_common_metadata` which has an impact on a completely empty table/dataframe. afaik, we do ensure, however, that columns which are loaded by MetaPartition *always* have columns of type string
https://github.com/JDASoftwareGroup/kartothek/blob/e8d597ccc5e65e3116b3e23303d450742cba16eb/kartothek/io_components/metapartition.py#L717
https://github.com/JDASoftwareGroup/kartothek/blob/e8d597ccc5e65e3116b3e23303d450742cba16eb/kartothek/io_components/metapartition.py#L1495

fjetter:
And yes, you are right, I am not a big fan of the pandas metadata :)

crepererum:
Might be off-topic, but how about the following proposal for the next metadata version (since it is breaking): We only accept dataframes w/ integer or range indices and always do a `reset_index` before storing the data to parquet. This would not create any additional columns anymore. Furthermore, we drop the whole pandas-metadata section. This would also ease the way to interact with other stacks (like Java, R, Julia, Rust, ...).

fjetter:
@crepererum Yes, slightly off-topic but generally I agree. Mostly because it's hard for me to imagine how a single-partition index generalizes to the distributed nature of kartothek but I guess we'll need to discuss the details before we settle on the details (another off-topic: we should figure out where we share design documents to discuss stuff like this)

crepererum:
I've opened #41 for this discussion.

lr4d:
Note: this also affects `kartothek.io.dask.delayed.read_table_as_delayed`

kartothek/core/utils.py
```diff
@@ -24,3 +24,10 @@ def _verify_metadata_version(metadata_version):
 
 def verify_metadata_version(*args, **kwargs):
     return _verify_metadata_version(*args, **kwargs)
+
+
+def _ensure_native_string_type(obj):
```
fjetter: Please rename this to `_ensure_string_type`. The _native_ part was lost a long time agi

CHANGES.rst
```diff
@@ -15,6 +15,14 @@ Version 3.1.0 (2019-XX-XX)
     + ``io.eager.update_dataset_from_dataframes``
 - fix ``_apply_partition_key_predicates`` ``FutureWarning``
 - serialize :class:`~kartothek.core.index.ExplicitSecondaryIndex` to parquet
+- Provide compatibility for dataset schemas created in Python2:
```
fjetter: I wouldn't promote this as a py2 compatibility feature. This might open the doors for support we explicitly don't want to give.
We have a small feature within kartothek that we ensure that columns always have `string` type and not `binary`. This PR rather fixes this for edge cases where the metadata was stored with columns of binary type (which **coincidentally** was the case in Py2)

kartothek/core/common_metadata.py
```diff
@@ -730,7 +731,12 @@ def empty_dataframe_from_schema(schema, columns=None, date_as_object=False):
     """
 
     if ARROW_LARGER_EQ_0130:
-        df = schema.internal().empty_table().to_pandas(date_as_object=date_as_object)
+        df = (
+            schema.remove_metadata()
```
fjetter: Can you remove this again? I think I'd prefer using the `_ensure_string_type` mapping for both code paths. This reduces the complexity since both paths use the same logic

kartothek/io/testing/read.py
```diff
@@ -539,3 +541,46 @@ def _f(b_c):
         predicates=[[("c", ">=", datetime.datetime(2000, 1, 1))]],
         dates_as_object=True,
     )
+
+
+@pytest.fixture
+def py2_reference_store():
```
fjetter: Same here regarding the py2 compat. Please rephrase this to `binary_column_metadata_store` (or something similar)

reference-data/kartothek-compat/py2-dataset/src/create_dataset.py
```diff
@@ -0,0 +1,29 @@
...
+
+def get_storefactory(store_config):
+    return partial(storefact.get_store, **store_config)
+
+
+uuid = "dataset_uuid"
+store = get_storefactory({"type": "hfs", "path": "../data"})
+df = pd.DataFrame(
+    {
+        "int_col": [1, 2, 3],
```
fjetter: You should be able to achieve the same thing by using `b"int_col"` w/out going back to py2. Kartothek isn't py2 compatible anymore. This is a big coincidence that this code could even be executed.

reference-data/kartothek-compat/py2-dataset/src/create_dataset.py
```diff
@@ -0,0 +1,29 @@
...
+from functools import partial
+
+import pandas as pd
+import storefact
+
+import kartothek
+from kartothek.io.eager.write import store_dataframes_as_dataset  # noqa
+
+assert sys.version_info.major == 2  # Show we're on Python 2
+assert LooseVersion(kartothek.__version__) < "3"  # Tested using `2.31.4`
```
fjetter: There is no public `2.31.4`

fjetter: I think we should not check in this file anyhow. Try creating the test data during runtime by defining the cols as `b"col_name"`

lr4d: Do you consider it more appropriate to simulate the previously-incompatible data, instead of using a dataset with identical characteristics as the one in the original bug report?

fjetter: We cannot base anything on non-public releases.

lr4d: Fair enough

CHANGES.rst
```diff
@@ -18,6 +18,10 @@ Version 3.1.0 (2019-XX-XX)
 - fix ``_apply_partition_key_predicates`` ``FutureWarning``
 - serialize :class:`~kartothek.core.index.ExplicitSecondaryIndex` to parquet
 - improve messages for schema violation errors
+- Ensure binary column names are read as type ``str``:
```
crepererum: Please use `-` for the nested list and add an empty line between between the parent and the nested list. Try to build the docs to see why this is required.

kartothek/io/testing/read.py
```diff
@@ -539,3 +539,31 @@ def _f(b_c):
         predicates=[[("c", ">=", datetime.datetime(2000, 1, 1))]],
         dates_as_object=True,
     )
+
+
+def test_binary_column_metadata(store_factory, bound_load_dataframes):
+    table_name = "core"
+    df = {
+        "label": "part1",
+        "data": [(table_name, pd.DataFrame({b"int_col": [1, 2, 3]}))],
```
crepererum: could you also add a unicode test case here?

kartothek/core/utils.py
```diff
@@ -24,3 +24,10 @@ def _verify_metadata_version(metadata_version):
 
 def verify_metadata_version(*args, **kwargs):
     return _verify_metadata_version(*args, **kwargs)
+
+
+def _ensure_string_type(obj):
```
crepererum: this is a public function. Functions starting with `_` are module-private, e.g. should not be used outside of this module. Also, some docstrings might help.

PR #42, opened 2019-05-22
Fix #16 

### LINKED ISSUE #16: Adding an index to an already existing dataset using update results in a corrupt index
When adding new partitions to an already existing dataset and specifying an index that does not exist in the dataset, that index is silently created, but references only the newly added partitions.

Expected behavior would be to raise an exception.

lr4d:
I have implemented a test to check whether the issue #16 takes place as described.

This test checks that kartothek raises an exception when a user tries to update a dataset and create a new index on that dataset, where the original dataset was created without an index.

### DIFF
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -1494,7 +1494,9 @@ def partition_labels_from_mps(mps):
     return partition_labels
 
 
-def parse_input_to_metapartition(obj, metadata_version=None):
+def parse_input_to_metapartition(
+    obj, metadata_version=None, expected_secondary_indices=False
+):
     """
     Parses given user input and returns a MetaPartition
 
@@ -1550,6 +1552,11 @@ def parse_input_to_metapartition(obj, metadata_version=None):
     obj : Union[Dict, pd.DataFrame, kartothek.io_components.metapartition.MetaPartition]
     metadata_version : int, optional
         The kartothek dataset specification version
+    expected_secondary_indices : Optional[Union[Iterable[str], Literal[False]]]
+        Iterable of strings containing expected columns on which indices are created. An empty iterable indicates no
+        indices are expected.
+        The default is `False`, which, indicates no checking will be done (`None` behaves the same way).
+        This is only used in mode "Dictionary with partition information".
 
     Raises
     ------
@@ -1568,12 +1575,16 @@ def parse_input_to_metapartition(obj, metadata_version=None):
             return MetaPartition(label=None, metadata_version=metadata_version)
         first_element = obj.pop()
         mp = parse_input_to_metapartition(
-            obj=first_element, metadata_version=metadata_version
+            obj=first_element,
+            metadata_version=metadata_version,
+            expected_secondary_indices=expected_secondary_indices,
         )
         for mp_in in obj:
             mp = mp.add_metapartition(
                 parse_input_to_metapartition(
-                    obj=mp_in, metadata_version=metadata_version
+                    obj=mp_in,
+                    metadata_version=metadata_version,
+                    expected_secondary_indices=expected_secondary_indices,
                 )
             )
     elif isinstance(obj, dict):
@@ -1581,11 +1592,19 @@ def parse_input_to_metapartition(obj, metadata_version=None):
             data = dict(obj["data"])
         else:
             data = obj["data"]
+
+        indices = obj.get("indices", {})
+        if expected_secondary_indices not in (False, None):
+            # Validate explicit input of indices
+            _ensure_valid_indices(
+                secondary_indices=expected_secondary_indices, mp_indices=indices
+            )
+
         mp = MetaPartition(
             # TODO: Deterministic hash for the input?
             label=obj.get("label", gen_uuid()),
             data=data,
-            indices=obj.get("indices", {}),
+            indices=indices,
             metadata_version=metadata_version,
         )
     elif isinstance(obj, pd.DataFrame):
@@ -1600,3 +1619,24 @@ def parse_input_to_metapartition(obj, metadata_version=None):
         raise ValueError("Unexpected type: {}".format(type(obj)))
 
     return mp
+
+
+def _ensure_valid_indices(secondary_indices, mp_indices):
+    if not secondary_indices:
+        if mp_indices:
+            raise ValueError(
+                "Incorrect indices provided for dataset.\n"
+                f"Expected index columns: {secondary_indices}"
+                f"Provided index: {mp_indices}"
+            )
+    else:
+        secondary_indices = set(secondary_indices)
+        # If the dataset has `secondary_indices` defined, then these indices will be build later so there is no need to
+        # ensure that they are also defined here (on a partition level).
+        # Hence,  we just check that no new indices are defined on the partition level.
+        if not secondary_indices.issuperset(mp_indices.keys()):
+            raise ValueError(
+                "Incorrect indices provided for dataset.\n"
+                f"Expected index columns: {secondary_indices}"
+                f"Provided index: {mp_indices}"
+            )
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -1510,7 +1510,8 @@ def parse_input_to_metapartition(
                         is generated using :func:`kartothek.core.uuid.gen_uuid`.
             * **data** - A dict or list of tuples. The keys represent the table name \
                         and the values are the actual payload data as a pandas.DataFrame.
-            * **indices** - A dictionary to describe the dataset indices. All \
+            * **indices** - Deprecated, see the keyword argument `secondary_indices` to create indices.
+                            A dictionary to describe the dataset indices. All \
                             partition level indices are finally merged using \
                             :func:`kartothek.io_components.metapartition.MetaPartition.merge_indices` \
                             into a single dataset index
@@ -1594,6 +1595,12 @@ def parse_input_to_metapartition(
             data = obj["data"]
 
         indices = obj.get("indices", {})
+        if indices:
+            warnings.warn(
+                "The explicit input of indices using the `indices` key is deprecated."
+                'Use the `secondary_indices` keyword argument of "write" and "update" functions instead.',
+                DeprecationWarning,
+            )
         if expected_secondary_indices not in (False, None):
             # Validate explicit input of indices
             _ensure_valid_indices(
--- a/kartothek/io_components/utils.py
+++ b/kartothek/io_components/utils.py
@@ -119,7 +119,10 @@ def _ensure_compatible_indices(dataset, secondary_indices):
             )
         return ds_secondary_indices
     else:
-        return secondary_indices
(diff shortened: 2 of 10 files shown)


lr4d:
After being referenced to some code which may have caused the issue, I will implement some variations of this test to see if I get it to fail

lr4d:
Found the bug. This should fix the issue

kartothek/io_components/metapartition.py
```diff
@@ -1600,8 +1602,42 @@ def parse_input_to_metapartition(obj, metadata_version=None):
...
+                "Incorrect indices provided for dataset.\n"
+                f"Expected index columns: {secondary_indices}"
+                f"Provided index: {mp_indices}"
+            )
+    else:
+        # If the dataset has `secondary_indices` defined, then these indices will be build later so there is no need to
+        # ensure that they are also defined here (on a partition level).
+        # Hence,  we just check that no new indices are defined on the partition level.
+        secondary_indices = set(secondary_indices)
+        if not secondary_indices.issuperset(mp_indices.keys()):
```
fjetter: If the keyword argument is used, I'd like to forbid the explicit input via the dictionary. I'd like to get rid of the explicit input in the next major release anyhow

kartothek/io_components/metapartition.py
```diff
@@ -1600,8 +1602,42 @@ def parse_input_to_metapartition(obj, metadata_version=None):
...
-        return obj
+        mp = obj
     else:
         raise ValueError("Unexpected type: {}".format(type(obj)))
 
+    # Check whether index columns of `mp` are equal to `assert_secondary_indices` to prevent
+    # metadata errors
+    if assert_partition_level_indices is not False:
+        if len(mp.metapartitions) > 1:  # nested metapartition
+            for _mp in mp.metapartitions:
```
fjetter: 1. Please don't loop over the metapartitions attribute explicitly but rather loop over the `MetaPartition` instance itself. There is a MetaPartitionIterator which takes care of this for you
2. Instead of looping here you should verify the index in the above code branch in `isinstance(obj, dict)`. Otherwise we'll verify this in every recursion call and for input which doesn't even support indices (e.g. DF only)

kartothek/io_components/metapartition.py
```diff
@@ -1499,7 +1499,9 @@ def _ensure_native_string_type(obj):
         return str(obj)
 
 
-def parse_input_to_metapartition(obj, metadata_version=None):
+def parse_input_to_metapartition(
+    obj, metadata_version=None, assert_partition_level_indices=False
```
fjetter: What's a `partition_level` index? We should probably stick to a nomenclature and we're usually referring to these as `secondary_indices`. Is there a reason why you changed this?

lr4d: I personally made a difference here between how indices are created, either using the`secondary_indices` kwarg or using what I termed "partition-level" indices: created using the `indices` key as in
```python
            {
                "label": "cluster_2",
                "data": [("core", df2)],
                "indices": {
                    "P": ExplicitSecondaryIndex(
                        "P", {v: ["cluster_2"] for v in values_p2}
                    )
                },
            },
```
Thus, I tried to emphasize that the the indices being checked in this commit are not those of the `secondary_indices` kwarg, but the indices of the kind defined in the code snippet. 
Anyways, I agree with the nomenclature point, so if you think this should be called `assert_secondary_indices` I will rename it

kartothek/io_components/metapartition.py
lr4d: Rookie mistake, my bad

kartothek/io_components/metapartition.py
```diff
@@ -1573,24 +1580,46 @@ def parse_input_to_metapartition(obj, metadata_version=None):
...
             data = obj["data"]
+
+        indices = obj.get("indices", {})
+        if indices:
+            warnings.warn(
+                "The explicit input of indices using the `indices` key is deprecated."
+                'Use the `secondary_indices` keyword argument of "write" and "update" functions instead.',
+                DeprecationWarning,
+            )
+        if hasattr(expected_secondary_indices, "__iter__"):
```
fjetter: Ideally, you shouldn't check for the `__iter__` attribute but be rather explicit about what you expect.
For instance a dict would be acceptable for you check as well (although I'd consider this bad input)
```python
In [1]: hasattr({}, "__iter__")
Out[1]: True
```

lr4d: I "expect" a set, since it cannot contain duplicates and will be used as a set anyways. But I thought "duck typing" would make it simple here for the users so they need not to worry about casting a list or tuple to set when making this call.

lr4d: I will change `normalize_args` so that `secondary_indices`  gets normalized to a `set`. I think that is a solution which allows for strict type parsing and flexible user input

lr4d: To remain consistent in our kwarg parsing, it would make sense that `partition_on` also supports a `set` as input (`secondary_indices` supports both `list` and `set`). I did not want to do this directly for performance reasons so I have opened #56

kartothek/io_components/docs.py
```diff
@@ -119,8 +119,8 @@
         kinds of predicates that are possible using boolean logic.
 """,
     "secondary_indices": """
-    secondary_indices: List[str]
-        A list of columns for which a secondary index should be calculated.
+    secondary_indices: Set[str]
```
fjetter: Please don't change this as
1. Technically speaking this would be a breaking change
2. We do not accept/enforce sets as input anywhere else
3. We neither validate nor cast anything to sets atm

kartothek/io_components/metapartition.py
```diff
@@ -1568,24 +1575,46 @@ def parse_input_to_metapartition(obj, metadata_version=None):
...
             data = obj["data"]
+
+        indices = obj.get("indices", {})
+        if indices:
+            warnings.warn(
+                "The explicit input of indices using the `indices` key is deprecated."
+                'Use the `secondary_indices` keyword argument of "write" and "update" functions instead.',
+                DeprecationWarning,
+            )
+        if isinstance(expected_secondary_indices, set):
```
fjetter: we should at least accept lists as well. While a `set` is mathematically speaking correct for the input, I'd argue that `list`s are a more common data type and I wouldn't want to restrict the input to `set`s. That makes the interface a bit cumbersome.

lr4d: `list`s (and tuples) are normalized to `set`s after the latest commit. We can discuss this later in more detail

kartothek/io_components/utils.py
```diff
@@ -157,14 +160,19 @@ def validate_partition_keys(
...
+_ARGS_TO_TYPE = {"partition_on": list, "delete_scope": list, "secondary_indices": set}
 
 
 def normalize_arg(arg_name, old_value):
     def _make_list(*_args):
         return list(_args)
 
-    type_to_normalize = {list: _make_list}
+    def _make_set(x):
+        if isinstance(x, (str, bytes)):
```
fjetter: This introduces different behavior for lists and sets. If we catch this edge case for strings/bytes we should do the same for lists. afaik, we're doing this for partition_on already but I'm not sure where and if this is consistently done

lr4d: Yes, the same happens for lists.
The thing is, any object you pass  (e.g. the tuple `(0,1)`) to the `_make_list` normalization function, the return value is a list with that object as the element, e.g. `[(0, 1)]`.
I wanted to avoid that in the `_make_set` normalization, so that lists are normalized to a proper set, to not cause a breaking change.
The underlying issue is that the normalization in `normalize_arg` is not strict nor really clear, as stated in #56 . It could probably use a re-write, but that would be out of scope for this PR.

kartothek/io_components/metapartition.py
fjetter: This isn't normalized but explicit input, isn't it?

kartothek/io_components/metapartition.py
```diff
@@ -1550,6 +1553,10 @@ def parse_input_to_metapartition(obj, metadata_version=None):
     obj : Union[Dict, pd.DataFrame, kartothek.io_components.metapartition.MetaPartition]
     metadata_version : int, optional
         The kartothek dataset specification version
+    expected_secondary_indices : Union[Set[str], Literal[False]], optional
```
fjetter: Any reason why to use `False` instead of `None` as the "don't verify anything" sentinel?

kartothek/io_components/metapartition.py
fjetter: You could make this section a bit easier top level and push the complexity down below

```
if expected_secondary_indices is not None:
    _ensure_valid_indices(secondary_indices=expected_secondary_indices, mp_indices=indices)

def _ensure_valid_indices():
    if secondary indices is None and mp_indices is not None:
        raise Exception
    else:
        # either case the secondary indices to a set or assert it to be a set
        if secondary_indices.issuperset(mp_indices.keys()):
            # perform validation
```
I would generally assume that a function should verify it's input values not the other way round.

kartothek/io_components/metapartition.py
lr4d: IMHO, `None` may be intepreted as "ensure there are no secondary indices", whereas `False` seems more clearly to represent "don't verify anything"

kartothek/io_components/docs.py
lr4d: I dropped the commit

kartothek/io_components/metapartition.py
lr4d: Thanks

kartothek/io_components/metapartition.py
lr4d: Both `None` and `False` are allowed now as "don't verify anything".

kartothek/io_components/metapartition.py
```diff
@@ -1600,3 +1626,24 @@ def parse_input_to_metapartition(obj, metadata_version=None):
         raise ValueError("Unexpected type: {}".format(type(obj)))
 
     return mp
+
+
+def _ensure_valid_indices(secondary_indices, mp_indices):
+    if not secondary_indices:
+        if mp_indices:
+            raise ValueError(
```
crepererum: Big thanks for writhing nice, detailed exception messages 👍

kartothek/io_components/metapartition.py
```diff
@@ -1550,6 +1553,11 @@ def parse_input_to_metapartition(obj, metadata_version=None):
     obj : Union[Dict, pd.DataFrame, kartothek.io_components.metapartition.MetaPartition]
     metadata_version : int, optional
         The kartothek dataset specification version
+    expected_secondary_indices : Union[Iterable[str], Literal[False], None], optional
```
crepererum: please leave out the optional part. It is already clear from the union type. You could also use `Optional[Union[Iterable[str], Literal[False]]]` if you prefer.

kartothek/io_components/metapartition.py
lr4d: My pleasure

PR #50, opened 2019-05-27
improve schema origin handling

crepererum:
For a very common usecase where different computation branches (e.g.
dask items, iterator elements) produce different partitions, let's try
to be more helpful for the user and tell them which part of the dataset
are not compatible.

### DIFF
--- a/kartothek/core/common_metadata.py
+++ b/kartothek/core/common_metadata.py
@@ -4,7 +4,9 @@
 import difflib
 import logging
 import pprint
-from copy import deepcopy
+from copy import copy, deepcopy
+from functools import reduce
+from typing import Set, Union
 
 import pandas as pd
 import pyarrow as pa
@@ -24,13 +26,33 @@ class SchemaWrapper(object):
     Wrapper object for pyarrow.Schema, since their pickle implementation is broken.
     """
 
-    def __init__(self, schema, origin):
-        if not isinstance(origin, str):
-            raise TypeError("Schema origin must either be None or a string.")
+    def __init__(self, schema, origin: Union[str, Set[str]]):
+        if isinstance(origin, str):
+            origin = {origin}
+        elif isinstance(origin, set):
+            origin = copy(origin)
+        if not all(isinstance(s, str) for s in origin):
+            raise TypeError("Schema origin elements must be strings.")
 
         self.__schema = schema
         self.__origin = origin
         self._schema_compat()
+
+    def with_origin(self, origin: Union[str, Set[str]]) -> "SchemaWrapper":
+        """
+        Create new SchemaWrapper with given origin.
+
+        Parameters
+        ----------
+        origin:
+            New origin.
+
+        Returns
+        -------
+        schema:
+            New schema.
+        """
+        return SchemaWrapper(self.__schema, origin)
 
     def _schema_compat(self):
         # https://issues.apache.org/jira/browse/ARROW-5104
@@ -81,8 +103,8 @@ class SchemaWrapper(object):
         return self.__schema
 
     @property
-    def origin(self):
-        return self.__origin
+    def origin(self) -> Set[str]:
+        return copy(self.__origin)
 
     def __repr__(self):
         return self.__schema.__repr__()
@@ -128,7 +150,8 @@ class SchemaWrapper(object):
 
     def remove_metadata(self):
         return SchemaWrapper(
-            self.__schema.remove_metadata(), self.__origin + "__no_metadata"
+            self.__schema.remove_metadata(),
+            {s + "__no_metadata" for s in self.__origin},
         )
 
     remove_metadata.__doc__ = pa.Schema.remove_metadata.__doc__
@@ -615,6 +638,18 @@ def validate_compatible(schemas, ignore_pandas=False):
             current, null_columns
         ).remove_metadata()
 
+        def _fmt_origin(origin):
+            origin = sorted(origin)
+            # dask cuts of exception messages at 1k chars:
+            #   https://github.com/dask/distributed/blob/6e0c0a6b90b1d3c/distributed/core.py#L964
+            # therefore, we cut the the maximum length
+            max_len = 200
+            inner_msg = ", ".join(origin)
+            ellipsis = "..."
+            if len(inner_msg) > max_len + len(ellipsis):
+                inner_msg = inner_msg[:max_len] + ellipsis
+            return "{{{}}}".format(inner_msg)
+
         if reference_to_compare != current_to_compare:
             schema_diff = _diff_schemas(reference, current)
             exception_message = """Schema violation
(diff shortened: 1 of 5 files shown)


kartothek/io_components/write.py
```diff
@@ -72,7 +72,8 @@ def persist_common_metadata(partition_list, update_dataset, store, dataset_uuid)
             )
 
     result = {}
-    for table, schemas in tm_dct.items():
+    for table in sorted(tm_dct.keys()):
+        schemas = sorted(tm_dct[table], key=lambda s: s.origin)
```
fjetter: why sort it?

kartothek/io_components/metapartition.py
```diff
@@ -1249,14 +1249,31 @@ def partition_on(self, partition_on):
...
-        new_mp = self.as_sentinel().copy(partition_keys=partition_on)
 
         if isinstance(partition_on, str):
             partition_on = [partition_on]
         partition_on = self._ensure_compatible_partitioning(partition_on)
 
         new_data = self._partition_data(partition_on)
 
+        # find a nice origin for the schema
+        if new_data:
```
fjetter: Why do we need this complicated logic? The partitioning shouldn't change the schema and the label / uuid of the original schema should be enough to track the partitions down.

If this is necessary, please add a comment what it is supposed to be doing. The split/zip/join/if is a bit confusing
Consider refactoring it into it's own function.

crepererum:
> can you give an example where this might be relevant?
please look at the new test case which illustrates the behavior (imo) quite well.

kartothek/io_components/write.py
crepererum: reproducible error messages. Same input, same error, which I think is nice for users (and tests). I'll add a comment.

kartothek/io_components/metapartition.py
fjetter: Regarding your top level comment
> please look at the new test case which illustrates the behavior (imo) quite well.

The new test case yields a violation

```
Schema violation
	
	Origin schema: core/P=2/uuid2
	Origin reference: core/P=1/uuid1
	
	Diff:
```

but there is a one-to-one mapping between P values and UUIDs in this example so it doesn't help me to understand.

I don't understand why we have to go through this label transformation logic here. To be exact, I don't even understand what this is doing. `new_data`s keys should be `P=1/uuid1` for the first partition and `P=2/uuid1` for the second partition. In this test case, the new labels are exactly what you would want to print. Is this logic about the missing tables in the keys?
In a more elaborate test case where the partition_on split actually splits the data (e.g. with P values 1 and 2) we would receive
`P=1/uuid1`
and
`P=2/uuid1`
which are again what you probably want to have (excluding tables)

1. Have you thought about attaching the origin to a given schema during MetaPartition initialization? The origin seems to be always connected to the label of the `MetaPartition` and by implementing the attachment of the origin there we don't have to go through this transformation. Plus this might safe us from inconsistencies in any other transformations where one might change labels in the future.
2. My original question is rather: why does the full label add any more value than e.g. `uuid1`? `uuid1` should determine the entirety of the input data

crepererum: UUIDs are random IDs by default (UUID4), the tests just hardcodes them for better error message assertion. So they are not that useful to determine which parts of your dataset have conflicting schemas. Sure, you can open a storage explorer, dig through the files with the UUIDs (which you only find by listing the whole dataset store, because no central metadata file will exist due to the canceled write transaction or the central metadata file does not contain the new files due to a canceled update transaction). I would very much call this a bad UX and therefore a bug.

So what I am trying here is to give the user more hints where the data is. You can only have 1 origin per MetaPartition instance, but normally, this is good enough, because for most usages 1 MetaPartition only spawns a tiny part of the dataset, often even exactly on partition-on combination. In case (see test) it spawns more (Q in P-Q in the test), we can still try to give a more helpful origin.

fjetter: I got the UX part. Just wanted to make sure I didn't miss anything.

Back to point 1)
Wouldn't is be possible to do something like:

```python
class MetaPartition:
    def __init__(..., label=None, table_meta=None, ...):
        self.label = label or gen_uuid()
        self.table_meta = {
            table: schema.with_origin(self.label) 
            for table, schema in table_meta.items()
        }
```

crepererum: This would work if the partition-on only creates a single new partition, otherwise `MetaPartition.table_meta` is 1 instance for all MPs within the nested/container MP. If you look at the test-case, this is exactly what happens for Q. You get 2 Q-values for 1 nested MP (containing 2 sub-MPs each). Another solution could be to store not 1 table_meta for all (sub)MPs within the container/nested MP, but to keep all of them. In that case, we could also have an origin for all these schemas 🤔

fjetter: Ok, took my a while but I finally understand what the test case and the code is supposed to be doing... (I hope, correct me if I'm wrong)
The exception shows `core/P=2/uuid2` but the partition label/path would be `core/P=2/Q=1/uuid2` *and* `core/P=2/Q=2/uuid2`, wouldn't it? I.e. the Q part is filtered out?

If this is the case, we should probably add all physical origins to the schemawrapper. As you said youself, iterating over the blob is *not* recommended and `core/P=2/<>/uuid2` is not sufficient to look up the file.
So we'd rather require something like

```
schema_wrapper.add_origins(sorted(new_data.keys()))
```

Yes, No?

crepererum: > The exception shows core/P=2/uuid2 but the partition label/path would be core/P=2/Q=1/uuid2 and core/P=2/Q=2/uuid2, wouldn't it? I.e. the Q part is filtered out?

Exactly. Sorry for not writing more docs for the specific test case.

> If this is the case, we should probably add all physical origins to the schemawrapper. 

Nice idea, didn't think of that. Will do so. :)

fjetter: If this is the case my proposed solution with `__init__` would of course not work out of the box. One could add a loop over the `table_meta` dict in `add_metapartition`, though (No preference, whatever is easier and consistent)

kartothek/core/common_metadata.py
```diff
@@ -616,6 +638,15 @@ def validate_compatible(schemas, ignore_pandas=False):
             current, null_columns
         ).remove_metadata()
 
+        def _fmt_origin(origin):
+            origin = sorted(origin)
+            # cut origin after a certain length since dask & others cut overlong exception messages
+            max_elements = 5
+            if len(origin) > max_elements:
```
fjetter: I know this is nitpicking but considering that dask cuts off at 1k chars, see https://github.com/dask/distributed/blob/6e0c0a6b90b1d3c3f686f0c968e9cf3d0c354413/distributed/core.py#L964
showing 5 origins may be even too much and we should rather cut regarding the length of the strings. Consider a realistic example (table name, two partition cols and a hex UUID)
```
table_name/a_string_column=a_long_name/a_date_column=2018-09-12/ef1aa9f0ce3e49e496ea93e889e812bc
```
we have 96 chars for a single origin. With 5 origins times two (reference and schema), the diff is completely cut off already.

Maybe we should rather do smth like
```python
lenghts = list(map(len, origins))
if sum(lengths) > 200:
# cut off but ensure there is at least one
else:
# return everything
```

On a side note: Should we log the entire exception? Good idea / horrible idea? I don't have much experience with loggers who receive *big* log messages

fjetter:
@crepererum Please also add a short changelog entry, thanks

PR #60, opened 2019-06-04
Doc strings and type annotations

fjetter:
This includes two changes:

1. I passed once over the docs using `flake8-docstrings` which is quite strict regarding the format and phrasing of the docstrings. I don't think we should enable it (yet) since it is quite strict but here are a few fixes for three modules I had a look at

2. For the same modules I added Type annotations. It was pretty straight forward with the exceptions of two things
a) I couldn't figure out how one could define the return type of a method to `type(self)` unless explicitly written as a string of the class name (classes are only registered once the entire body is parsed) Not sure if there is a prettier solution
b) `IndexBase` relied on the check `index_dct is None` to indicate whether or not the index is loaded already. I replace this with an explicit flag since that messed up the type checks heavily for essentially every method. i.e. a index which isn't loaded will still have an (empty) `index_dct` now, we could catch this by implementing it as a property and check for `is_loaded` and otherwise raise.


crepererum:
> a) I couldn't figure out how one could define the return type of a method to type(self) unless explicitly written as a string of the class name (classes are only registered once the entire body is parsed) Not sure if there is a prettier solution

Please go for the string solution. It is the default work-around also understood by mypy. The restriction was [lifted in python 3.7 as an opt-int](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep563), but since this is not our minimum requirement, we have to stick with the string-based solution.

> b) IndexBase relied on the check index_dct is None to indicate whether or not the index is loaded already. I replace this with an explicit flag...

Since we're not optimizing C/C++ code here, I think explicit flags are fine. It is easier to reason about and the memory overhead is IMHO negligible.

crepererum:
I think it would be great to use flake8-docstrings at one point. I personally think that too many docs are just broken out there and that this can safely considered to be a bug.

fjetter:
> Since we're not optimizing C/C++ code here, I think explicit flags are fine. It is easier to reason about and the memory overhead is IMHO negligible.

My issue is not about memory overhead or performance 😂 

I'm rather wondering about the behavior since `None` previously flagged a not loaded index (which, thinking about it, doesn't make too much sense for the base class anyhow but rather only for the Secondary index)
Since tests break everywhere my suspicion was right... will need to investigate

lr4d:
I have seen string types defined as `String`, `string` and `str`. 
Are these interchangeable?

crepererum:
@lr4d `str` is the only correct annotation.

Side note: The weird forms like `Set`, `List`, `Tuple` etc. are only required because `set`, `list`, `tuple` and Co are generic types and the python syntax does not allow you to specify their concrete type (like "list of integers"). So the workaround the Python community created is to overload the `[]`-operator for certain special objects in `typing` so you can write stuff like `List[int]`.

fjetter:
@crepererum do you know if there is a sphinx extension which renders the type info in the doc strings nicely? I get that shifting to the `typing` style documentation, like `List[str]` is very explicit and in alignment with the type annotation but strictly speaking numpy doc style would list this as `list of str`

crepererum:
@fjetter you're right. Seems like numpy is not following proper python typing. 😭 This Python-100-standards-thing drives me crazy. Then I would say: we're not following NumPy style in that case, because I don't think we should write type annotations AND types into the docstrings, especially because the docstring types aren't checked in any way.

kartothek/core/dataset.py
```diff
@@ -155,7 +148,7 @@ def exists(uuid, store):
         return key in store
 
     @staticmethod
-    def storage_keys(uuid, store):
+    def storage_keys(uuid: str, store: KeyValueStore):
```
lr4d: return value? I see it's defined in the docstring atm

kartothek/core/dataset.py
```diff
@@ -180,35 +173,43 @@ def storage_keys(uuid, store):
...
+    def to_dict(self) -> Dict:
         dct = OrderedDict(
             [
                 (naming.METADATA_VERSION_KEY, self.metadata_version),
                 (naming.UUID_KEY, self.uuid),
             ]
         )
         if self.indices:
-            dct["indices"] = {k: v.to_dict() for k, v in self.indices.items()}
+            dct["indices"] = {
```
lr4d: what's the reason for changing the logic here?

fjetter: Previously the `to_dict` method returned either the dict or the storage key. That's obviously not type safe and this is where it breaks. To fis this I needed to introduce this complicated dict comprehension

PR #62, opened 2019-06-07
re-write getting started guide

lr4d:
Major re-write
Added:
- Update section
- GC
- Partition/Indices

Improved:
- Write
- Read
- Storage

With input from @kagharpure . He may be able to apply requested changes if necessary.

### DIFF
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -37,7 +37,7 @@ of written data didn't require reading through an entire dataset to be able to
 identify and access the required subset. This is where *explicitly* partitioning by
 table columns helps.
 
-To see explicit partitioning in action, lets set up some data and a storage location
+To see explicit partitioning in action, let's set up some data and a storage location
 first and store the data there with ``kartothek``:
 
 .. ipython:: python
@@ -210,7 +210,7 @@ provides update functions that generally have the prefix `update_dataset` in the
 For example, :func:`~kartothek.io.eager.update_dataset_from_dataframes` is the update
 function for the ``eager`` backend.
 
-To see updating in action, lets set up some data and a storage location first and store
+To see updating in action, let's set up some data and a storage location first and store
 the data there with ``kartothek``:
 
 .. ipython:: python
@@ -261,7 +261,7 @@ The way dataset updates work is that new partitions are added to a dataset
 as long as they have the same tables as the existing partitions. A `different`
 table **cannot** be introduced into an existing dataset with an update.
 
-To illustrate this point better, lets first create a dataset with two tables:
+To illustrate this point better, let's first create a dataset with two tables:
 
 .. ipython:: python
 
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -392,9 +392,10 @@ function is called on the dataset.
 If a dataset is updated on a regular basis, it may be useful to run garbage collection
 periodically to decrease unnecessary storage use.
 
-An example of garbage collection is shown below. A file named ``trash.parquet`` is
-created in storage but untracked by kartothek. When garbage collection is called, the
-file is removed.
+An example of garbage collection is shown below. Suppose a file named
+``E=train/F=x/d513c388.parquet`` might have been referenced before a deletion or
+compaction operation. This file remains in storage but is untracked by kartothek.
+When garbage collection is called, the file is removed.
 
 .. ipython:: python
 
@@ -402,7 +403,7 @@ file is removed.
 
     store = store_factory()
     # Put corrupt parquet file in storage for dataset "a_unique_dataset_identifier"
-    store.put("a_unique_dataset_identifier/table/trash.parquet", b"trash")
+    store.put("a_unique_dataset_identifier/table/E=train/F=x/d513c388.parquet", b"trash")
     files_before = set(store.keys())
 
     garbage_collect_dataset(store=store_factory, dataset_uuid="a_unique_dataset_identifier")
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -81,7 +81,7 @@ Of interest here is ``dm.partitions``:
 
 .. ipython:: python
 
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 We can see that partitions have been stored in a way which indicates the
@@ -101,7 +101,7 @@ be specified as a list:
         [df, duplicate_df],
         partition_on=["E", "F"],
     )
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Note that, since 2 dataframes have been provided as input to the function, there are
@@ -130,7 +130,7 @@ For example:
         partition_on="B",
     )
 
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Because partitions are physical in nature, it is not possible to modify the
@@ -207,7 +207,7 @@ the data there with ``kartothek``:
 .. ipython:: python
 
     dm = store_dataframes_as_dataset(store_factory, "a_unique_dataset_identifier", df)
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Now, we create ``another_df`` with the same schema as our intial dataframe
@@ -231,7 +231,7 @@ Now, we create ``another_df`` with the same schema as our intial dataframe
     dm = update_dataset_from_dataframes(
         [another_df], store=store_factory, dataset_uuid="a_unique_dataset_identifier"
     )
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Looking at ``dm.partitions``, we can see that another partition has
@@ -274,7 +274,7 @@ To illustrate this point better, lets first create a dataset with two tables:
         dfs=[{"data": {"table1": df, "table2": df2}}],
     )
     dm.tables
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Below is an example where we update the existing dataset ``another_unique_dataset_identifier``
@@ -300,7 +300,7 @@ with new data for ``table1`` and ``table2``:
         dataset_uuid="another_unique_dataset_identifier",
     )
     dm.tables
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 Trying to update only a subset of tables throws a ``ValueError``:
@@ -338,7 +338,7 @@ keyword argument as shown in the example below:
         partition_on="E",
         delete_scope=[{"E": "train"}],
     )
-    dm.partitions
+    sorted(dm.partitions.keys())
 
 
 As we can see, we specified using a dictionary that data where the column ``E`` has the
@@ -367,7 +367,7 @@ list but have to be specified instead as individual dictionaries, i.e.
         delete_scope=[{"E": "train", "F": "foo"}, {"E": "test", "F": "bar"}],
     )
 
-    dm.partitions  # `E=train/F=foo` and `E=test/F=bar` are deleted
+    sorted(dm.partitions.keys())  # `E=train/F=foo` and `E=test/F=bar` are deleted
 
 
 
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -20,10 +20,10 @@ particular applications.
 Partitioning and Indexing
 =========================
 
-
 ``kartothek`` is designed primarily for storing large datasets consistently and
 accessing them efficiently. To achieve this, it provides two useful functionalities:
 partitioning and secondary indices.
+
 
 Partitioning
 ------------
@@ -33,11 +33,12 @@ partitions, which in the underlying key-value store translates to writing new fi
 to the storage layer.
 
 From the perspective of efficient access, it would be helpful if accessing a subset
-of written data didn't require reading through an entire dataset to be able to identify
-and access the required subset. This is where partitioning by table columns helps.
+of written data didn't require reading through an entire dataset to be able to
+identify and access the required subset. This is where *explicitly* partitioning by
+table columns helps.
 
-To see partitioning in action, lets set up some data and a storage location first and store
-the data there with ``kartothek``:
+To see explicit partitioning in action, lets set up some data and a storage location
+first and store the data there with ``kartothek``:
 
 .. ipython:: python
 
@@ -65,16 +66,18 @@ the data there with ``kartothek``:
     )
     df
 
-``kartothek`` allows users to partition their data by the
-values of table columns such that, for a given input partition,
-all the rows with the same value of the column all get
-written to the same partition. To do this, we use the ``partition_on`` keyword argument:
+
+``kartothek`` allows users to explicitly partition their data by the values of table
+columns such that, for a given input partition, all the rows with the same value of the
+column all get written to the same partition. To do this, we use the
+``partition_on`` keyword argument:
 
 .. ipython:: python
 
     dm = store_dataframes_as_dataset(
         store_factory, "partitioned_dataset", df, partition_on="E"
     )
+
 
 Of interest here is ``dm.partitions``:
 
@@ -86,8 +89,8 @@ Of interest here is ``dm.partitions``:
 We can see that partitions have been stored in a way which indicates the
 specific value for the column on which partitioning has been performed.
 
-Partitioning can also be performed on multiple columns; in this case, columns should
-be specified as a list:
+Partitioning can also be performed on multiple columns; in this case, columns
+should be specified as a list:
 
 .. ipython:: python
 
@@ -105,12 +108,13 @@ be specified as a list:
 
 Note that, since 2 dataframes have been provided as input to the function, there are
 4 different files created, even though only 2 different combinations of values of E and
-F are found, ``E=test/F=foo`` and ``E=train/F=foo`` .
-(However, these 4 physical partitions can be read as just the 2 logical partitions by
-using the argument ``concat_partitions_on_primary_index=True`` at reading time).
+F are found, ``E=test/F=foo`` and ``E=train/F=foo`` (However, these 4 physical partitions
+can be read as just the 2 logical partitions by using the argument
+``concat_partitions_on_primary_index=True`` at reading time).
 
-For datasets consisting of multiple tables, partitioning on
-columns only works if the column exists in both tables and is of the same data type.
+For datasets consisting of multiple tables, explicit partitioning on columns can only be
+performed if the column exists in both tables and is of the same data type: guaranteeing
+that their types are the same is part of schema validation in ``kartothek``.
 
 For example:
 
@@ -132,21 +136,26 @@ For example:
     sorted(dm.partitions.keys())
 
 
-When data is appended to a dataset, ``kartothek`` guarantees it has the proper schema
-and partitioning.
+As noted above, when data is appended to a dataset, ``kartothek`` guarantees it has
+the proper schema and partitioning.
 
 .. note:: Every partition must have data for every table. An empty dataframe in this
           context is also considered as data.
 
+
 Secondary Indices
 -----------------
 
-The ability to build and maintain secondary indices are an additional ability
-provided by ``kartothek``. Secondary indices are `similar` to partitions in the
-sense that they allow faster access to subsets of data. The main difference
-between them is that while partitioning actually creates separate partitions based
-on column values, secondary indices are simply python dictionaries mapping column
-values and the partitions that rows with them can be found in.
+The ability to build and maintain 'secondary' indices are an additional ability
+provided by ``kartothek``. In general, an index is a data structure used to improve
+the speed of read queries. In the context of ``kartothek`` an index is a data structure
+that contains a mapping of every unique value of a given column to references to all the
+partitions where this value occurs.
+
+Explicitly partitioning on a column internally creates a :class:`~kartothek.core.index.PartitionIndex`.
+While this index has a one-to-one mapping of column values to partition references,
+secondary indices have the advantage of being able to contain one-to-many mappings of
+column values to partition references.
 
 .. note::
 
@@ -155,6 +164,7 @@ values and the partitions that rows with them can be found in.
     on a per-partition basis and accordingly data inputs are expected to be generators.
     Although using other iterables such as lists also works, doing so is counter
     to the intent of the ``iter`` backend (lists would be appropriate in ``eager``).
+
 
 Writing a dataset with a secondary index:
 
@@ -185,9 +195,11 @@ As can be seen from the example above, both ``partition_on`` and ``secondary_ind
 can be specified together. Multiple ``secondary_indices`` can also be added as a list of
 strings.
 
-In general, secondary indices behave like partitions in terms of when and how they can
-and cannot be created. However, when using ``partition_on`` the order of the columns
-provided is important, whereas it is ignored for ``secondary_indices``.
+In general, secondary indices behave like explicit partitions in terms of when and how they
+can and cannot be created. However, when using ``partition_on`` the order of the columns
+provided is important, whereas it is ignored for ``secondary_indices``: this is because the
+underlying processes differ and creation of the :class:`~kartothek.core.index.PartitionIndex`
+is merely a side-effect of using ``partition_on``.
 
 
 Updating an existing dataset
@@ -342,13 +354,23 @@ As we can see, we specified using a dictionary that data where the column ``E`` 
 value ``train`` should be removed. Looking at the partitions after the update, we see that
 the partition ``E=train`` has been removed.
 
-.. note:: We defined ``delete_scope`` over a value of ``E``. ``E`` also happens to be a
-    column that we partitioned by. This is because using ``delete_scope`` uses the same
-    underlying logic as the predicate-based filtering mentioned in :ref:`getting_started`.
+.. note:: We defined ``delete_scope`` over a value of ``E`` which is also the column that
+    we partitioned on: ``delete_scope`` only works *as expected* on indexed columns.
 
-    Attempting to use ``delete_scope`` will *also* work on datasets not previously
-    partitioned on any column(s); in this case however the effect will simply be to remove
-    **all** previous partitions and replace them with the ones in the update.
+    Furthermore it *should only* be used on partitioned columns due to their one-to-one mapping;
+    without the guarantee of one-to-one mappings, using ``delete_scope`` could have unwanted
+    effects like accidentally removing data with different values.
+
+
+.. admonition:: Using ``delete_scope`` on data not partitioned by columns
+
+    Attempting to use ``delete_scope`` *will also* work on datasets not previously partitioned
+    on any column(s); however the effect will simply be to remove **all** previous partitions
+    and replace them with the ones in the update and therfore **should not** be used in such cases.
+
+    If the intention of the user is to delete existing partitions, using :func:`kartothek.io.eager.delete_dataset`
+    would be a much better and safer way to go about doing so.
+
 
 When  using ``delete_scope``, multiple values for the same column cannot be defined as a
 list but have to be specified instead as individual dictionaries, i.e.
@@ -367,9 +389,6 @@ list but have to be specified instead as individual dictionaries, i.e.
     sorted(dm.partitions.keys())  # `E=train/F=foo` and `E=test/F=bar` are deleted
 
 
-
-
-
 Garbage collection
 ==================
 
@@ -382,7 +401,6 @@ appear in the output.
 
 Similarly, when deleting a partition, ``kartothek`` only removes the reference of that file
 from the metadata.
-
 
 These temporary files will remain in storage until a ``kartothek``  garbage collection
 function is called on the dataset.
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -325,9 +325,9 @@ Trying to update only a subset of tables throws a ``ValueError``:
     Tables of existing dataset: ['table1', 'table2']
 
 
-Adding data to existing datasets is not the only functionality achievable within an update
-and it can even be used to remove or replace data. To do this we use the ``delete_scope``
-keyword argument as shown in the example below:
+Adding data to an existing dataset is not the only functionality achievable within an update
+operation, such an operation can also be used to remove or overwrite data.
+To do so we use the ``delete_scope`` keyword argument as shown in the example below:
 
 .. ipython:: python
 
@@ -379,7 +379,7 @@ Garbage collection
 When ``kartothek`` is executing an operation, it makes sure to not
 commit changes to the dataset until the operation has been succesfully completed. If a
 write operation does not succeed for any reason, although there may be new files written
-to storage, those files will not used by the dataset as they will not be referenced in
+to storage, those files will not be used by the dataset as they will not be referenced in
 the ``kartothek`` metadata. Thus, when the user reads the dataset, no new data will
 appear in the output.
 
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -157,6 +157,9 @@ For example:
 As noted above, when data is appended to a dataset, ``kartothek`` guarantees it has
 the proper schema and partitioning.
 
+The order of columns provided in ``partition_on`` is important, as the partition
+structure would be different if the columns are in a different order.
+
 .. note:: Every partition must have data for every table. An empty dataframe in this
           context is also considered as data.
 
@@ -164,26 +167,16 @@ the proper schema and partitioning.
 Secondary Indices
 -----------------
 
-The ability to build and maintain 'secondary' indices are an additional ability
-provided by ``kartothek``. In general, an index is a data structure used to improve
+The ability to build and maintain `inverted indices <https://en.wikipedia.org/wiki/Inverted_index>`_
+are an additional feature provided by ``kartothek``.
+In general, an index is a data structure used to improve
 the speed of read queries. In the context of ``kartothek`` an index is a data structure
 that contains a mapping of every unique value of a given column to references to all the
 partitions where this value occurs.
 
-Explicitly partitioning on a column internally creates a :class:`~kartothek.core.index.PartitionIndex`.
 While this index has a one-to-one mapping of column values to partition references,
 secondary indices have the advantage of being able to contain one-to-many mappings of
 column values to partition references.
-
-.. note::
-
-    The examples we've looked at so far have mostly used functions from the ``eager``
-    backend. As noted earlier, the ``iter`` backend executes operations on the dataset
-    using an iterator/generator interface (data input is expected to be provided as a
-    generator).
-    Although using other iterables such as lists also works, doing so is counter
-    to the intent of the ``iter`` backend (lists would be appropriate in ``eager``).
-
 
 Writing a dataset with a secondary index:
 
@@ -214,11 +207,7 @@ As can be seen from the example above, both ``partition_on`` and ``secondary_ind
 can be specified together. Multiple ``secondary_indices`` can also be added as a list of
 strings.
 
-In general, secondary indices behave like explicit partitions in terms of when and how they
-can and cannot be created. However, when using ``partition_on`` the order of the columns
-provided is important, whereas it is ignored for ``secondary_indices``: this is because the
-underlying processes differ and creation of the :class:`~kartothek.core.index.PartitionIndex`
-is merely a side-effect of using ``partition_on``.
+Contrary to ``partition_on``, the order of columns is ignored for ``secondary_indices``.
 
 
 Updating existing data
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -25,13 +25,6 @@ support `ExtendedKeyspaceMixin
 <https://github.com/mbr/simplekv/search?q=%22class+ExtendedKeyspaceMixin%22&unscoped_q=%22class+ExtendedKeyspaceMixin%22>`_
 (this is necessary so that ``/`` can be used in the storage key name).
 
-`storefact`_ offers support for several stores to ``kartothek``, these can be created using the
-function :func:`~storefact.get_store_from_url` with one of the following prefixes:
-
-- ``hfs``: Local filesystem
-- ``hazure``: AzureBlockBlobStorage
-- ``hs3``:  BotoStore (Amazon S3)
-
 For more information, take a look out at the `storefact documentation
 <https://storefact.readthedocs.io/en/latest/reference/storefact.html>`_.
 
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -373,7 +373,7 @@ As we can see, we specified using a dictionary that data where the column ``E`` 
 value ``train`` should be removed. Looking at the partitions after the update, we see that
 the partition ``E=train`` has been removed.
 
-.. note:: we defined ``delete_scope`` over a value of ``E``. ``E`` also happens to be a
+.. note:: We defined ``delete_scope`` over a value of ``E``. ``E`` also happens to be a
     column that we partitioned by. This is because using ``delete_scope`` uses the same
     underlying logic as the predicate-based filtering mentioned in :ref:`getting_started`.
 
@@ -382,7 +382,7 @@ the partition ``E=train`` has been removed.
     **all** previous partitions and replace them with the ones in the update.
 
 When  using ``delete_scope``, multiple values for the same column cannot be defined as a
-list but have to be specified instead as individual dictionaries, ie.
+list but have to be specified instead as individual dictionaries, i.e.
 ``[{"E": ["test", "train"]}]`` will not work but ``[{"E": "test"}, {"E": "train"}]`` will.
 
 .. ipython:: python
--- a/docs/further_useful_features.rst
+++ b/docs/further_useful_features.rst
@@ -227,22 +227,26 @@ underlying processes differ and creation of the :class:`~kartothek.core.index.Pa
 is merely a side-effect of using ``partition_on``.
 
 
-Updating an existing dataset
-============================
+Updating existing data
+======================
 
-It's possible to update datasets by adding new physical partitions to them, ``kartothek``
+It's possible to update existing data (formally referred to in ``kartothek`` as a dataset)
+by adding new physical partitions to them and deleting or replacing old partitions. ``kartothek``
 provides update functions that generally have the prefix `update_dataset` in their names.
 For example, :func:`~kartothek.io.eager.update_dataset_from_dataframes` is the update
 function for the ``eager`` backend.
 
-To see updating in action, let's set up some data and a storage location first and store
-the data there with ``kartothek``:
+To see updating in action, let's first set up a storage location first and store some data there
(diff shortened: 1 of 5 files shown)


kagharpure:
Yes, will work on any changes required.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
-The most interesting ones for this guide are ``tables`` and ``partitions``.
-Each dataset can have multiple tables, each table is a collection of files
-that all have the same schema. These files are called ``partitions`` in
-``kartothek``.
+For this guide, two attributes that are noteworthy are ``tables`` and ``partitions``:
 
-As we neither have explicitly defined the name of the table nor the name
+- Each dataset has one or more ``tables``, where each table represents a particular subset of
+  data, this data is stored as a collection of dataframes/files which have the same schema.
+- Data is written to storage in batches (for ``eager``, there is only a single batch),
```
fjetter: I wouldn't map the term `Batch` to a `Partition`, primarily because the batches as seen by the user are often different to the actual partitions. The ktk input is a batch as seen by the user but the partition is an implementation detail. In particular there are multiple possibilities for the mapping between the two:
- One input batch -> one partition
- One input batch -> multiple partitions
- Many input batches -> many partitions (e.g. when using dask shuffle)

While you are explaining a portion of the ambiguity here, the entirety is a bit more complex and beyond a getting started guide.
In this section you're introducing the term `Partition` which is a physical shard of the data, possibly for multiple tables. The origin of the data can have multiple source and I would simply state that this shard is not necessarily a one-to-one mapping to the input. The `partition_on` is a particular example where this is not the case. Instead of describing it's usage, I would forward to the section about the partitioning which describes an example where there is no one-to-one mapping.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+
+
+For each table, ``kartothek`` also tracks the schema of the columns.
+Unless specified explicitly on write, it is inferred from the passed data.
+On writing additional data to a dataset, we will also check that the schema
+of the new data matches the schema of the existing data.
+
+.. admonition:: Passing multiple partitions to a dataset during write/update
+
+    To store multiple dataframes into a dataset (i.e. multiple `partitions`), it is possible
```
fjetter: This is also ambiguous. Storing multiple dataframes into a dataset could be misunderstood as writing _different_ dataframes into a table. You should be explicit that _multiple partitions_ means _multiple data chunks_ with the same schema

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+    to pass an iterator of dataframes, the exact format will depend on the I/O backend used.
+
+    If passing an iterator of dataframes, and table names are not specified, ``kartothek``
```
fjetter: This is also a detail which is not necessarily required for the getting started guide, is it? This is essentially a summary of the `parse_input_to_metapartition` function. It might be worth linking that function here there the actual user interface is documented for now

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+                    "core-table": pd.DataFrame({"col1": ["y"]}),
+                    "aux-table": pd.DataFrame({"f": [3.2]}),
+                },
+            },
+       ]
+
+       store_dataframes_as_dataset(store_factory, dataset_uuid="two-tables", dfs=dfs)
+
+
+Reading dataset from storage
```
fjetter: Reading data from storage

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+=============================
+
+After we have written the data, we may want to read it back in again. For this we can
+use :func:`kartothek.io.eager.read_table`. This method returns the complete
+table of the dataset as a pandas DataFrame (since there is only a single table in this
```
fjetter: The comment in the braces seems unnecessary

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+represent the `tables` of the dataset. For example,
+
+.. ipython:: python
+
+    from kartothek.io.iter import read_dataset_as_dataframes__iterator
+
+    for partition_index, df_dict in enumerate(
+            read_dataset_as_dataframes__iterator(dataset_uuid="two-tables", store=store_factory)
+        ):
+            print(f"Partition #{partition_index}")
```
fjetter: Is this print rendered by sphinx somehow?

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+            for table_name, table_df in df_dict.items():
+                print(f"Table: {table_name}. Data: \n{table_df}")
+
+
+.. admonition:: Filtering the dataset using predicates
+
+    It is possible to perform reading queries similar to an SQL ``WHERE`` statement using
```
fjetter: I wouldn't reference SQL here. We're nowhere near having full SQL whereclause support. I'd simply mentioning that we allow predicates in disjunctive normal form. If possible, we perform partition pruning and predicate pushdown to speed up queries.
In detail, this is a complex topic and requires in-depth documentation beyond a getting started guide

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+    .. ipython:: python
+
+        # Read only values table `aux-table` where `f` < 2.5
+        read_table(
+            "two-tables", store_factory, table="aux-table", predicates=[[("f", "<", 2.5)]]
+        )
+
+
+
+Updating an existing dataset
```
fjetter: I'm wondering if this shouldn't be on it's own page. For a getting started guide this is already quite long.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+
+Looking at ``dm.partitions``, we can see that another partition has
+been added.
+
+If we read the data again, we can see that the ``another_df`` has been appended to the
+previous contents.
+
+.. ipython:: python
+
+    df_again = read_table("a_unique_dataset_identifier", store_factory, table="table")
```
fjetter: Maybe something like `df_updated`

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+        read_table(
+            "two-tables", store_factory, table="aux-table", predicates=[[("f", "<", 2.5)]]
+        )
+
+
+
+Updating an existing dataset
+============================
+
+It's possible to update datasets by adding new physical partitions to them, ``kartothek`` provides
```
fjetter: It's not only possible to update a dataset by appending but it is also possible to remove/replace partitions using the `delete_scope` argument. It would be nice to have an example for how this works

kagharpure:
@fjetter - thanks for the comments, lots of good feedback there; will start working on it. 

On a separate note, are you still in the middle of reviewing? There are no comments from you on the partitioning and garbage collection sections yet.

fjetter:
@kagharpure as I said, please split the rewrite of the old section from the new section. It's too much to review. I stopped at some point

docs/getting_started.rst
```diff
@@ -1,28 +1,36 @@
+===============
 Getting started
 ===============
 
-``kartothek`` manages datasets that consist of files that contain tables.
-When working with these tables as a Python user, we will use pandas DataFrames
-as the user-facing type. We typically expect that the dataset contents are
-large, often too large to be held in a single machine but for demonstration
-purposes, we use a small DataFrame with a mixed set of types.
+``kartothek`` offers a metadata definition to handle these datasets efficiently.
```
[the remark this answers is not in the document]
kagharpure: Fair, adding grammatical fix.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
-The next module level ``eager`` describes the scheduling backend.
-``eager`` runs all execution immediately and on the local machine.
-There is also ``iter`` that supports reading the dataset on
-a per-partition base. For larger dataset, ``dask`` can be used
-to work on datasets in parallel or even in a cluster by using
-``distributed`` as the backend for ``dask``.
+Writing dataset to storage
+===========================
+
+Now that we have our data and the storage location, we can persist the dataset.
```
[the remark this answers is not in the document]
kagharpure: This refers to the `DataFrame` `df` created above. Calling that out explicitly in the rewrite.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+To do so, we will use :func:`kartothek.io.eager.store_dataframes_as_dataset`
+to store a ``DataFrame`` we already have in memory.
```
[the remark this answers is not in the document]
kagharpure: > Ideally I'd hope that a user can execute the statements one by one when opening an IPython shell
Yes, it has to be so because the ipython directive within the doc itself persists the session across cells, so building the docs would throw an error if that weren't the case.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
-      store, "a_unique_dataset_identifier", df, metadata_version=4
-   )
-   dm
+    from kartothek.io.eager import store_dataframes_as_dataset
+    dm = store_dataframes_as_dataset(
+        store_factory,
+        "a_unique_dataset_identifier",
+        df,
+    )
+    dm
```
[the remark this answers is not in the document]
kagharpure: Yes, it is used later on.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+This class holds information about the structure and schema of the dataset.
 
-The most interesting ones for this guide are ``tables`` and ``partitions``.
-Each dataset can have multiple tables, each table is a collection of files
-that all have the same schema. These files are called ``partitions`` in
-``kartothek``.
+For this guide, two attributes that are noteworthy are ``tables`` and ``partitions``:
 
-As we neither have explicitly defined the name of the table nor the name
+- Each dataset has one or more ``tables``, where each table represents a particular subset of
```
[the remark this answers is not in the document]
kagharpure: Good point. Trying to make this less ambiguous and more straightforward.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+  data, this data is stored as a collection of dataframes/files which have the same schema.
```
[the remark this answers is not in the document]
kagharpure: Duly removed.

docs/getting_started.rst
kagharpure: Fair point. Have attempted to simplify.

docs/getting_started.rst
kagharpure: Personally I am okay with the ambiguity exactly because it does encapsulate a bunch of possibilities. Thus the reason for this being within an admonition. The bigger idea being to convey that 'of course you can pass in multiple dataframes in one call to a write function but if and when you do so please be aware of how `kartothek` behaves in such cases so that you are not caught off-guard'.

docs/getting_started.rst
kagharpure: Agreed. The reason we're calling this out explicitly upfront is because it is what happens and can be a point of confusion for users when they try to pass in multiple dataframes with different schemas anonymously, ie. without defining table names against those dataframes. Will call this out a bit more explicitly and link to `parse_input_to_metapartition`.

docs/getting_started.rst
```diff
@@ -31,105 +39,467 @@ example to construct such a store for the local filesystem.
...
+
+.. admonition:: Passing multiple partitions to a dataset during write/update
+
+    To store multiple dataframes into a dataset (i.e. multiple `partitions`), it is possible
+    to pass an iterator of dataframes, the exact format will depend on the I/O backend used.
+
+    If passing an iterator of dataframes, and table names are not specified, ``kartothek``
+    assumes these dataframes are different partitions with a single table.
+
+A ``ValueError`` will be thrown if there is a mismatch in the schema. For example,
```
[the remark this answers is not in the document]
kagharpure: Added, along with an additional illustrative example.

docs/getting_started.rst
kagharpure: Yes.

docs/getting_started.rst
kagharpure: Breaking out updating, partitioning and GC to a second doc, `Further useful features in kartothek`.

docs/getting_started.rst
kagharpure: Will leave out the SQL reference and simplify this a bit.

docs/getting_started.rst
kagharpure: Good point. Will add an example for this.

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+
+.. note:: Under the hood, partitions are structurally identical to each other and each partition
+    is made up of a collection of files containing the subset of data of each table
+    belonging to that partition.
+
+Secondary Indices
+-----------------
+
+The ability to build and maintain secondary indices are an additional ability
+provided by ``kartothek``. Secondary indices are `similar` to partitions in the
```
fjetter: _partitions_ are a physical collection of files. The _partitioning_ schema you've been describing above are *not* partitions. Every dataset is made of _partitions_ even if I do not put any sense to it.

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+sense that they allow faster access to subsets of data. The main difference
+between them is that while partitioning actually creates separate partitions based
+on column values, secondary indices are simply python dictionaries mapping column
```
fjetter: Technically speaking the `PartitionIndex` which is created above is also just a dictionary in the end.

Instead of describing the data type you should describe what the index is all about: For every unique value of a given column over the entire dataset, we store all references to the partitions where this value occurs.

What makes the secondary index special opposed to the `PartitionIndex` above is that it allows for N:1 mappings (N unique values in one partition) while the PartitionIndex is a 1:1 relation

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+            }
+        )
+        for i in range(1, 6)
+    )
+
+    dm = store_dataframes_as_dataset__iter(
+        df_gen,
+        store_factory,
+        "secondarily_indexed",
+        partition_on = "date",
```
fjetter: whitespaces between operator. Should be `partition_on="date"`

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+        secondary_indices = "X"
+    )
+    dm
+
+    dm = dm.load_all_indices(store_factory())
+    dm.secondary_indices['X'].index_dct[0]  # Show files where `X == 0`
```
fjetter: instead of accessing the index dictionary directly, I'd recommend to

```
dm.indices["X"].eval_operator("==", 0)
```

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+
+
+As can be seen from the example above, both ``partition_on`` and ``secondary_indices``
+can be specified together. Multiple ``secondary_indices`` can also be added as a list of
+strings.
+
+In general, secondary indices behave like partitions in terms of when and how they can
```
fjetter: Again, careful about the term _partitions_ here

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+and cannot be created. However, when using ``partition_on`` the order of the columns
```
fjetter: This is because it is not the same process. The Index is merely a side effect for `partition_on`

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+       ....:        dataset_uuid="another_unique_dataset_identifier"
+       ....:        )
+       ....:
+    ---------------------------------------------------------------------------
+    ValueError: Input partitions for update have different tables than dataset:
+    Input partition tables: {'table2'}
+    Tables of existing dataset: ['table1', 'table2']
+
+
+Adding data to existing datasets is not the only functionality achievable within an update
```
fjetter: Grammar:
Adding data to _an_ existing dataset~s~ is not the only functionality ...

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+    dm.partitions
+
+
+As we can see, we specified using a dictionary that data where the column ``E`` has the
+value ``train`` should be removed. Looking at the partitions after the update, we see that
+the partition ``E=train`` has been removed.
+
+.. note:: We defined ``delete_scope`` over a value of ``E``. ``E`` also happens to be a
+    column that we partitioned by. This is because using ``delete_scope`` uses the same
+    underlying logic as the predicate-based filtering mentioned in :ref:`getting_started`.
```
fjetter: Not true. This is *not* the predicate logic but a rather simplified input schema. I would reference it this way

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+        delete_scope=[{"E": "train"}]
+    )
+    dm.partitions
+
+
+As we can see, we specified using a dictionary that data where the column ``E`` has the
+value ``train`` should be removed. Looking at the partitions after the update, we see that
+the partition ``E=train`` has been removed.
+
+.. note:: We defined ``delete_scope`` over a value of ``E``. ``E`` also happens to be a
```
fjetter: _happens_ implies that it is by accident. There are two very important things in here:
1. delete scope *only* works on indexed columns
2. it *should* only be used on partitioned columns due to the 1:1 mapping I mentioned before. Without the guarantee of 1:1 we would accidentally remove data with different values

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+    column that we partitioned by. This is because using ``delete_scope`` uses the same
+    underlying logic as the predicate-based filtering mentioned in :ref:`getting_started`.
+
+    Attempting to use ``delete_scope`` will *also* work on datasets not previously
```
fjetter: This sounds as if it was a feature. The behavior you describe rather sounds like: Be careful! If you do things wrong you delete your entire data!

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+        }
+    )
+    df2
+
+    dm = store_dataframes_as_dataset(
+        store_factory,
+        "another_unique_dataset_identifier",
+        dfs = [
+            {
+                "data":
```
fjetter: I'd like to see a section here mentioning that the usage of the explicit label is discouraged here since for the update to properly work using copy-on-write principles we require the filenames to be unique. This is why kartothek uses UUIDs for the labels by default

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+
+
+
+Garbage collection
+==================
+
+When ``kartothek`` is executing an operation, it makes sure to not
+commit changes to the dataset until the operation has been succesfully completed. If a
+write operation does not succeed for any reason, although there may be new files written
+to storage, those files will not used by the dataset as they will not be referenced in
```
fjetter: those files will note *be* used by the dataset

Instead of _used_ I'd use the term _referenced_

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+Similarly, when deleting a partition, ``kartothek`` only removes the reference of that file
+from the metadata.
+
+
+These temporary files will remain in storage until a ``kartothek``  garbage collection
+function is called on the dataset.
+If a dataset is updated on a regular basis, it may be useful to run garbage collection
+periodically to decrease unnecessary storage use.
+
+An example of garbage collection is shown below. A file named ``trash.parquet`` is
```
fjetter: I think from a user perspective the `trash.parquet` is not necessarily intuitive. Think of using a partitioned dataset schema where the trash looks like `colA=1/label.parquet` and mention that this file might have been referenced before an update, delete or compaction operation

docs/getting_started.rst
```diff
@@ -110,19 +110,16 @@ For this guide, two attributes that are noteworthy are ``tables`` and ``partitio
 
 - Each dataset has one or more ``tables``, where each table is a logical collection of data,
   bound together by a common schema.
-- Partitions are the mutually exclusive and collectively exhaustive physical `pieces` or
-  `shards` of data that constitute the contents of a dataset. Data is written to storage in
-  batches (for ``eager``, there is only a single batch) and in the simplest case, such a
-  batch is synonymous with a ``partition`` in ``kartothek``. In practice however, the mapping
-  between partitions and batches is almost never one-to-one. See the section on partitioning
-  for further details: :ref:`partitioning_section`.
+- Partitions are the mutually exclusive and collectively exhaustive physical `pieces` of data
```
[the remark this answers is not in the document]
kagharpure: Oops, sorry, wasn't aware that exclusivity isn't enforced, assumed that was the case. Will just leave it at collectively exhaustive.

docs/further_useful_features.rst
kagharpure: Ah, sorry, I inferred from our conversation last week that predicate-based filtering and delete-scope both use the same underlying mechanism (ie. `dataset.query`).

docs/further_useful_features.rst
lr4d: It may be useful to add a pre-commit hook to run black on docs using https://github.com/asottile/blacken-docs in the future

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+    duplicate_df = df.copy()
+    duplicate_df.F = "bar"
+
+    dm = store_dataframes_as_dataset(
+        store_factory,
+        "another_partitioned_dataset",
+        [df, duplicate_df],
+        partition_on = ['E', 'F'],
+    )
+    dm.partitions
```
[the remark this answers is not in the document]
lr4d: Replaced with `sorted(dm.partitions.keys())`

docs/further_useful_features.rst
lr4d: Also add a comment about future deprecation of explicit labels?

docs/getting_started.rst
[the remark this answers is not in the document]
lr4d: Replaced this with a much simpler sentence

docs/getting_started.rst
```diff
@@ -197,28 +194,25 @@ to :func:`~kartothek.io.eager.store_dataframes_as_dataset` throws ``ValueError``
...
 
     Sometimes it may be useful to write multiple dataframes with different schemas into
     a single dataset. This can be achieved by creating a dataset with multiple tables.
 
-    In this example, we create a dataset with two partitions ``partition1`` and ``partition2``
-    (represented by the dictionary objects inside the list). For each partition, there exist
-    two tables: ``core-table`` and ``aux-table``. The schemas of the tables are identical
-    across partitions.
+    In this example, we create a dataset with two tables: ``core-table`` and ``aux-table``.
+    The schemas of the tables are identical across partitions (each dictionary in the
```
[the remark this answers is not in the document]
lr4d: Highlighted just before this section. Also clarified `parse_input_to_metapartition` sentence

docs/further_useful_features.rst
lr4d: _referenced_ already _used_ 8 words later

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+            "C": pd.Series(1, index=list(range(4)), dtype="float32"),
+            "D": np.array([3] * 4, dtype="int32"),
+            "E": pd.Categorical(["test", "train", "test", "train"]),
+            "F": "foo",
+        }
+    )
+    df
+
+``kartothek`` allows users to (physically) partition their data by the
+values of table columns such that all the rows with the same value of the column all get
```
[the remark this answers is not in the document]
lr4d: Added comment about only being true for a given partition,

docs/getting_started.rst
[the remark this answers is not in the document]
kagharpure: Erm.. I have a doubt here. Isn't MECE achieved by default at the row-level? Specifically, two duplicate rows may end up in different partitions, but the same row will never be replicated across partitions, right?

kagharpure:
@fjetter, @lr4d - I agree that the term partitioning has been thrown about a bit loosely in the guide in the previous version that I'd committed. What I am going to do now is to call out the fact that while partitioning happens by default, what users can do is *explicitly* partition by column; I think this will help delineate the difference between the two better.

docs/further_useful_features.rst
lr4d: Wouldn't mentioning that it is possible to use explicit labels go against the point of not talking about features we do not want users to use?

docs/further_useful_features.rst
```diff
@@ -0,0 +1,432 @@
...
+    import pandas as pd
+    from functools import partial
+    from tempfile import TemporaryDirectory
+    from storefact import get_store_from_url
+
+    from kartothek.io.eager import store_dataframes_as_dataset
+
+    dataset_dir = TemporaryDirectory()
+
+    store_factory = partial(get_store_from_url, f"hfs://{dataset_dir.name}")
```
crepererum: can we somewhere here mentioned that we require extended keyspace stores? See #69 for details.

docs/further_useful_features.rst
lr4d: Atm https://github.com/JDASoftwareGroup/kartothek/pull/62/commits/2d450bfaabcd67d3cdcecc72606f486b6cb36662 adds mention of why UUID's are used as labels, does not talk about explicit labels

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+
+    store_factory = partial(get_store_from_url, f"hfs://{dataset_dir.name}")
 
-   dataset_dir = TemporaryDirectory()
-   store = get_store_from_url(f"hfs://{dataset_dir.name}")
 
-Now that we have our data and the storage location, we can persist it to a dataset.
-For that we use in this guide :func:`kartothek.io.eager.store_dataframes_as_dataset`
-to store a ``DataFrame`` we already have in memory in the local task.
+For a more technical description of storage specification in kartothek, see
```
fjetter: What kind of technical description? Does this link exist, yet? What would you want to write down there? Instead of deferring to another page, I'd create a text box which describes that we're using the simplekv interface and the user should ensure that the store he is using is an instance of `ExtendedKeyspaceMixin`, achieved by using the `h-` prefix when using storefact

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
-``table`` and used a generated UUID for the partition name.
+    The import path of this function already gives us a hint about the general
+    structuring of the ``kartothek`` modules. In :mod:`kartothek.io` we have all
+    the building blocks to build data pipelines that read and write from/to storages.
+    The next module level (e.g. ``eager``) describes the scheduling backend.
+
+    The scheduling backends `currently supported` by kartothek are:
+
+    - ``eager`` runs all execution immediately and on the local machine.
+    - ``iter`` executes operations on the dataset on a per-partition basis.
```
fjetter: The per-partition bit is true for all other backends as well. The important part is that it is a python generator/iterator interface.

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+      The standard format to read/store dataframes in ``iter`` is by providing
+      a generator of dataframes.
+    - ``dask`` is suitable for larger datasets. It can be used to work on datasets in
+      parallel or even in a cluster by using ``dask.distributed`` as the backend.
+      There are also ``dask.bag`` and ``dask.dataframe`` which support I/O operations
+      for the respective `dask`_ objects.
```
fjetter: `dask`_ collections.

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
 
 For each table, ``kartothek`` also tracks the schema of the columns.
-When not specified explicitly on write, it is inferred from the passed data.
+Unless specified explicitly on write, it is inferred from the passed data.
 On writing additional data to a dataset, we will also check that the schema
-of the new data matches the schema of the existing data. If it doesn't, we will
-raise an exception.
+of the new data matches the schema of the existing data.
+
+While two tables may have different schemas, every file belonging to a given table needs
```
fjetter: This paragraph is too technical and prob too much information for the getting started guide. This is/should be covered in the writeup of the storage specification but not the getting started guide https://kartothek.readthedocs.io/en/latest/spec/index.html

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+to have the same schema as the table. Furthermore, every partition needs to have a file
+for every table (empty files are allowed).
+
+.. admonition:: Writing multiple dataframes with identical schemas
```
fjetter: this feels a bit artificial. Passing a list/iterable should be the default usage and not a special case. Consider rewriting this. Maybe merge it with the initial write data section

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+    The schemas of the tables are identical across partitions (each dictionary in the
+    ``dfs`` list argument represents a partition).
+
+    .. ipython:: python
+
+        dfs = [
+            {
+                "data": {
+                    "core-table": pd.DataFrame({"col1": ["x"]}),
+                    "aux-table": pd.DataFrame({"f": [1.1]}),
```
fjetter: I recommend to use a common column for the two tables. Storing data this way hardly offers any benefits over storing the data in two separate datasets. OTOH, if you have a common column this is an opportunity for trivial join logic. I wouldn't necessarily mention this in a getting started guide, though.

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+            print(f"Table: {table_name}. Data: \n{table_df}")
+
+
+.. admonition:: Filtering the dataset using predicates
+
+    It is possible to filter data during reads using simple predicates by using
+    the ``predicates`` argument. Technically speaking, ``kartothek`` supports predicates
+    in `disjunctive normal form <https://en.wikipedia.org/wiki/Disjunctive_normal_form>`_.
+
+    When this argument is defined, ``kartothek`` tries to use the Apache Parquet metadata
```
fjetter: We're not using Parquet metadata for partition pruning. We only use our indices for partition pruning. The parquet metadata is used for predicate pushdown (how this works should be a chapter on its own, though)

docs/getting_started.rst
```diff
@@ -1,135 +1,288 @@
...
+    for partition pruning and predicate pushdown to speed up queries when possible. How this
+    works is a complex topic, well beyond the scope of this guide but it is a 'good-to-know'
```
fjetter: I'd scratch the "good-to-know" part and simply refer the user to the appropriate section (Create one if necessary and add a TODO)

docs/getting_started.rst
fjetter: The important thing is that the whole partition pruning/predicate pushdown stuff is done behind the scenes. The use *always* receives only what he requested

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+keep their data stores 'clean' with garbage collection.
+
+The intent of this document is to expose users to some additional features of ``kartothek``
+in a bid to encourage deeper exploration of ``kartothek`` beyond 'getting started' and help
+better enable developers to start thinking about how it can (or cannot) be useful in their
+particular applications.
+
+.. _storage_specification:
+
+Storage specification
```
fjetter: This section is not _further reading_ but a *hard* requirement for using kartothek. Please move this at the top of the getting started guide. At the very least the getting started guide should highlight the most important parts (e.g. h prefix

docs/further_useful_features.rst
```diff
@@ -0,0 +1,438 @@
...
+
+Because partitions are physical in nature, it is not possible to modify the
+partitioning scheme of an existing dataset via an update, instead, the dataset
+would have to be re-created.
+
+.. note:: Under the hood, partitions are structurally identical to each other and each partition
+    is made up of a collection of files containing the subset of data of each table
+    belonging to that partition.
+
+Secondary Indices
```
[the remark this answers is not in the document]
fjetter: Still missing this reference

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+the proper schema and partitioning.
+
+.. note:: Every partition must have data for every table. An empty dataframe in this
+          context is also considered as data.
+
+
+Secondary Indices
+-----------------
+
+The ability to build and maintain 'secondary' indices are an additional ability
```
fjetter: > The ability to [...] are an additional ability [...]

Try rephrasing

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+provided by ``kartothek``. In general, an index is a data structure used to improve
```
fjetter: Use the technical term _inverted index_

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+the speed of read queries. In the context of ``kartothek`` an index is a data structure
```
fjetter: Not sure if we should dig deeper here into how the index actually works. For the user, the most important thing is that it speeds up a query *automatically* if used in combination with `predicates`, i.e. scratch the mapping _value to references_ part in favor of a description about how the indices will be used. In an ideal world, the user doesn't need to bother with the internal design at all

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+that contains a mapping of every unique value of a given column to references to all the
+partitions where this value occurs.
+
+Explicitly partitioning on a column internally creates a :class:`~kartothek.core.index.PartitionIndex`.
```
fjetter: The class itself is probably also not very interesting and may be subject to change. What's important is that there is *always* an index available when partitioning on a field

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+While this index has a one-to-one mapping of column values to partition references,
+secondary indices have the advantage of being able to contain one-to-many mappings of
+column values to partition references.
+
+.. note::
+
+    The examples we've looked at so far have mostly used functions from the ``eager``
```
fjetter: I believe we covered this already in the getting started guide when the different backends are introduced, didn't we?

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+        df_gen,
+        store_factory,
+        "secondarily_indexed",
+        partition_on="date",
+        secondary_indices="X",
+    )
+    dm
+
+    dm = dm.load_all_indices(store_factory())
+    dm.indices["X"].eval_operator("==", 0)  # Show files where `X == 0`
```
fjetter: Fine by me. But it is very important to highlight that the index evaluation is done by kartothek. I've seen users trying to construct their own index eval logic although this is done (and optimized) when using `predicates` (see above comment)

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+
+
+As can be seen from the example above, both ``partition_on`` and ``secondary_indices``
+can be specified together. Multiple ``secondary_indices`` can also be added as a list of
+strings.
+
+In general, secondary indices behave like explicit partitions in terms of when and how they
```
fjetter: This is a very general statement which is probably not even true.
Technically speaking you can build a secondary index for *every* fields but you cannot partition on every field (since there would be no data left after the partitioning). This is only the very first restriction which comes to my mind and there are potentially even more. Thinking about it, I believe you can also build a secondary index on a field which only exists in a single table but cannot partition on it (Don't mention this. this is not a feature, rather a (un)happy coincident atm)

I would refrain from using _In general_ since _general_ implications and similarities are hard to pin down. The actual message you want to pass to the user is: _Mind that the order of partition_on matters_ and this is something I would expect to be part of the actual `partition_on` documentation and not part of the getting started guide, i.e. please check in the `default_docs` implementation if this is the case and add it there.

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+
+It's possible to update datasets by adding new physical partitions to them, ``kartothek``
+provides update functions that generally have the prefix `update_dataset` in their names.
+For example, :func:`~kartothek.io.eager.update_dataset_from_dataframes` is the update
+function for the ``eager`` backend.
+
+To see updating in action, let's set up some data and a storage location first and store
+the data there with ``kartothek``:
+
+.. ipython:: python
```
fjetter: Not necessary but I'm curious: Is it possible to arrange this in a collapsable box? After the getting started guide the reader would probably not read the actual data setup and this would help the reading flow. It's useful to be there if somebody want to look up details but in general this is not the feature we want to promote here

Just wondering. A "dont know", "too complicated" or a "will take a lot of time" is a completely satisfying answer at this point :)

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+        }
+    )
+
+    dm = update_dataset_from_dataframes(
+        [another_df], store=store_factory, dataset_uuid="a_unique_dataset_identifier"
+    )
+    sorted(dm.partitions.keys())
+
+
+Looking at ``dm.partitions``, we can see that another partition has
```
fjetter: While the _add partitions_ piece is what's actually happening, I guess the user is much more interested in the data which is added to the dataset. I instruct people who write tests for datasets usually to *not* assert for given partitions but rather for the actual data to be there, e.g.

```
read_table(..., predicates=[[("A", "==", 5.0)]])
# print new row
...
```

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+    dm = store_dataframes_as_dataset(
+        store_factory,
+        "another_unique_dataset_identifier",
+        dfs=[{"data": {"table1": df, "table2": df2}}],
+    )
+    dm.tables
+    sorted(dm.partitions.keys())
+
+
+.. admonition:: Filenames
```
fjetter: This is rather what we refer to as partition labels. Do not refer to the explicit file names. People should trust and rely on kartothek to manage the files. The mentioning of the uniqueness of the labels is nice, though

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+       ....:        dataset_uuid="another_unique_dataset_identifier"
+       ....:        )
+       ....:
+    ---------------------------------------------------------------------------
+    ValueError: Input partitions for update have different tables than dataset:
+    Input partition tables: {'table2'}
+    Tables of existing dataset: ['table1', 'table2']
+
+
+Adding data to an existing dataset is not the only functionality achievable within an update
```
fjetter: I'd restructure this chapter into subsections. 

# Dataset updates
## Append data
## Delete data
## Replace data

Feel free to choose better titles

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+    )
+    sorted(dm.partitions.keys())
+
+
+As we can see, we specified using a dictionary that data where the column ``E`` has the
+value ``train`` should be removed. Looking at the partitions after the update, we see that
+the partition ``E=train`` has been removed.
+
+.. note:: We defined ``delete_scope`` over a value of ``E`` which is also the column that
+    we partitioned on: ``delete_scope`` only works *as expected* on indexed columns.
```
fjetter: It *only* works on indexed columns and the recommendation is to use it for *partition columns* only

fjetter: Feel free to promote this from `note` to `warning`

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+
+    Furthermore it *should only* be used on partitioned columns due to their one-to-one mapping;
+    without the guarantee of one-to-one mappings, using ``delete_scope`` could have unwanted
+    effects like accidentally removing data with different values.
+
+
+.. admonition:: Using ``delete_scope`` on data not partitioned by columns
```
fjetter: Not entirely sure if this is a feature or a bug. I would consider it a bug if a delete scope is requested for a column for which we don't have index information

docs/further_useful_features.rst
```diff
@@ -0,0 +1,459 @@
...
+
+    Attempting to use ``delete_scope`` *will also* work on datasets not previously partitioned
+    on any column(s); however the effect will simply be to remove **all** previous partitions
+    and replace them with the ones in the update and therfore **should not** be used in such cases.
+
+    If the intention of the user is to delete existing partitions, using :func:`kartothek.io.eager.delete_dataset`
```
fjetter: You mean `If the intention of the user is to delete *all* existing partitions`, don't you?

docs/further_useful_features.rst
fjetter: The replace section would be the usage of both *append* and *delete_scope* simultaneously

docs/further_useful_features.rst
kagharpure: Regardless of the label (bug vs feature) this is possible right now - ie. it doesn't throw an error; thus the admonition, to basically warn the user not use it for the purpose of deletion. Probably this too could benefit from being promoted to a warning.

docs/getting_started.rst
kagharpure: Yes, it's linked to the Storage Specification section in the Further Useful Features guide.

lr4d: I agree with recommending to use the `h-` prefix in storefact, but why bother a new user with technical requirements of the store (`ExtendedKeyspaceMixin` et al)?

docs/further_useful_features.rst
kagharpure: > It _only_ works on indexed columns and the recommendation is to use it for _partition columns_ only

The recommendation is in the next line in the (now) warning.

docs/getting_started.rst
lr4d: A list, generator or list of dask objects are quite different I'd say. How would you suggest rephrasing this?

docs/further_useful_features.rst
lr4d: Highlighted h prefix stores

docs/further_useful_features.rst
lr4d: too complicated

docs/further_useful_features.rst
lr4d: "_how the indices will be used_", could you be more specific?

docs/further_useful_features.rst
lr4d: That's a few lines further. Should I delete the `dm.partitions` part?

docs/further_useful_features.rst
lr4d: Simplified this, saying column order matters for `partition_on`, doesn't for `secondary_indices`. Note that this is part of the partitioning section already, this is not the getting started guide

lr4d:
@fjetter in light of the large number of changes requested, I suggest we sit down when we both have some time and finish the document, otherwise we might continue going back and forth for weeks

PR #64, opened 2019-06-11
delete pipeline garbage collection-WIP

mganesh1308:

PR #67, opened 2019-06-17
Call garbage collection in delete pipelines

lr4d:
Continuation of: https://github.com/JDASoftwareGroup/kartothek/pull/64

Summary: call GC in functions which delete the entire dataset to remove possible unreferenced files.

Added the test `kartothek.io.testing.delete.test_delete_dataset_unreferenced_files`.

Two minor bugs where found in GC functions, when passed a `ds_factory` instead of a `store` and `dataset_uuid`. These were fixed

Original author: @mganesh1308

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,8 @@ Version 3.1.0 (2019-XX-XX)
 - fix issue where it was possible to add an index to an existing dataset by using update functions and partition indices
   (https://github.com/JDASoftwareGroup/kartothek/issues/16).
 
+- fix issue where unreferenced files were not being removed when deleting an entire dataset
+
 **Breaking:**
 
 - categorical normalization was moved from :meth:`~kartothek.core.common_metadata.make_meta` to
--- a/kartothek/io/dask/delayed.py
+++ b/kartothek/io/dask/delayed.py
@@ -117,7 +117,10 @@ def garbage_collect_dataset__delayed(
         chunk_size=chunk_size,
         factory=factory,
     )
-    return [delayed(delete_files)(files, store_factory=store) for files in nested_files]
+    return [
+        delayed(delete_files)(files, store_factory=store or factory.store_factory)
+        for files in nested_files
+    ]
 
 
 def _load_and_merge_mps(mp_list, store, label_merger, metadata_merger, merge_tasks):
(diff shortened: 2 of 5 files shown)


CHANGES.rst
```diff
@@ -27,6 +27,9 @@ Version 3.1.0 (2019-XX-XX)
 - fix issue where it was possible to add an index to an existing dataset by using update functions and partition indices
   (https://github.com/JDASoftwareGroup/kartothek/issues/16).
 
+- fix issue where unreferenced files were not being removed when deleting an entire dataset and create test
```
fjetter: Don't mention the test in the changelog. We're always writing tests and the user is not interested in this

kartothek/io/dask/delayed.py
```diff
@@ -115,7 +117,10 @@ def garbage_collect_dataset__delayed(
         chunk_size=chunk_size,
         factory=factory,
     )
-    return [delayed(delete_files)(files, store_factory=store) for files in nested_files]
+    return [
+        delayed(delete_files)(files, store_factory=store or factory.store_factory)
```
fjetter: Instead of doing this `or` combination, please use `_ensure_factory`

kartothek/io/eager.py
```diff
@@ -743,4 +746,6 @@ def garbage_collect_dataset(dataset_uuid=None, store=None, factory=None):
 
     # Given that `nested_files` is a generator with a single element, just
     # return the output of `delete_files` on that element.
-    return delete_files(next(nested_files), store_factory=store)
+    return delete_files(
+        next(nested_files), store_factory=store or factory.store_factory
```
fjetter: use `_ensure_factory`

kartothek/io/testing/delete.py
```diff
@@ -69,3 +69,21 @@ def test_delete_missing_dataset(store_factory, store_factory2, bound_delete_data
...
+    store_factory, metadata_version, bound_delete_dataset
+):
+    """
+    Ensure that unreferenced files of a dataset are also removed when a dataset is deleted
+    """
+    uuid = "dataset"
+    create_dataset(uuid, store_factory, metadata_version)
+
+    store = store_factory()
+    store.put(f"{uuid}/core/trash.parquet", b"trash")
```
fjetter: Can you add another file w/out the uuid prefix and ensure it isn't deleted?

lr4d: That is already tested in `test_delete_only_dataset` and `test_delete_single_dataset`, wouldn't it be redundant?

fjetter: Yeah, I knew we had this somewhere I just couldn't find it before. Works for me 👍

PR #70, opened 2019-06-20
remove leftovers from old metadata versions

crepererum:


lr4d:
It would make sense to also remove the `@pytest.mark.min_metadata_version(4)` decorator which is present on a lot of tests

crepererum:
Do we also want to remove the test fixture `metadata_version` completely or should we keep this part of the test infrastructure intact because we might need it soonish? CC @fjetter

fjetter:
I kept both the fixture and the mark intentionally since I figured we'll require the same again at some point

I don't care too much about the decorator but I would keep the `metadata_version` as a fixture since this is a very easy way to spot breakage once we introduce a new version.

crepererum:
codecov is a flux, I couldn't find anything that got worse.

PR #72, opened 2019-06-24
Fix #40

### LINKED ISSUE #40: commit_dataset does not work with nested partitions
The following code example

```
from kartothek.io.eager import commit_dataset, write_single_partition, create_empty_dataset_header
from kartothek.core.common_metadata import make_meta
import pandas as pd
import storefact


store = storefact.get_store_from_url("hfs:///tmp/some_directory")
for x in store.keys():
    store.delete(x)

df = pd.DataFrame({"a": [1, 1, 2, 2], "b": [3, 4, 5, 6]})

create_empty_dataset_header(
    store=store,
    dataset_uuid="uuid",
    table_meta={"table": make_meta(df, "table", ["a"])},
    partition_on=["a"],
)

partitions = []
for x in range(2):
    partitions.append(
        write_single_partition(
            store=store,
            dataset_uuid="uuid",
            data={"data": {"table": df}},
            partition_on=["a"],
        )
    )

commit_dataset(
    store=store,
    dataset_uuid="uuid",
    new_partitions=partitions,
    partition_on=["a"],
)
```

Will create the error message `AttributeError: Accessing `label` attribute is not allowed while nested`.

Additionally you can observe the behaviour that after `commit_dataset` the `partitions` list contains one less element. In my understanding `commit_dataset` should not manipulate input objects.

lr4d:
First commit addresses a side-effect caused by `parse_input_to_metapartition`
The third commit adds support for nested metapartitions in `add_metapartition`, which was the root cause of error in #40.

### DIFF
--- a/tests/io/eager/test_commit.py
+++ b/tests/io/eager/test_commit.py
@@ -168,6 +168,39 @@ def test_commit_dataset_from_dict(dataset_function, store):
     assert_frame_equal(df_expected, actual)
 
 
+def test_commit_dataset_from_nested_metapartition(store):
+    """
+    Check it is possible to use `commit_dataset` with nested metapartitions as input.
+    Original issue: https://github.com/JDASoftwareGroup/kartothek/issues/40
+    """
+
+    df = pd.DataFrame({"a": [1, 1, 2, 2], "b": [3, 4, 5, 6]})
+
+    create_empty_dataset_header(
+        store=store,
+        dataset_uuid="uuid",
+        table_meta={"table": make_meta(df, "table", ["a"])},
+        partition_on=["a"],
+    )
+
+    partitions = []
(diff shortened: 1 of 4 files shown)


tests/io/eager/test_commit.py
```diff
@@ -171,6 +171,37 @@ def test_commit_dataset_from_dict(dataset_function, store):
...
+                store=store,
+                dataset_uuid="uuid",
+                data={"data": {"table": df}},
+                partition_on=["a"],
+            )
+        )
+
+    commit_dataset(
+        store=store, dataset_uuid="uuid", new_partitions=partitions, partition_on=["a"]
+    )
```
crepererum: please add a short assertion that the expected partitions are now part of the dataset.

PR #73, opened 2019-06-25
Fix partition_on documentation

fjetter:


kartothek/io_components/docs.py
```diff
@@ -65,8 +65,9 @@
     "partition_on": """
     partition_on: list
         Column names by which the dataset should be partitioned by.
```
crepererum:
```suggestion
         Column names by which the dataset should be partitioned by physically.
```

This makes it clearer that this is not just an index structure.

PR #75, opened 2019-06-27
Dask index build

crepererum:
We could argue if the function should be placed in `delayed` or in `bag`. While it is internally implemented using `Dask.Bag`, the return type just behaves as a usual delayed computation, so I would opt for "implementation detail".

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,8 @@ Version 3.1.0 (2019-XX-XX)
   in :meth:`~kartothek.io_components.metapartition.MetaPartition.add_metapartition`.
   This fixes issue https://github.com/JDASoftwareGroup/kartothek/issues/40 .
 
+- Add :meth:`~karothek.io.dask.bag.build_dataset_indices__bag`
+
 **Breaking:**
 
 - categorical normalization was moved from :meth:`~kartothek.core.common_metadata.make_meta` to
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
-
-
 from functools import partial
 
+import dask.bag as db
+
 from kartothek.core import naming
+from kartothek.core.factory import _ensure_factory
 from kartothek.core.utils import _check_callable
 from kartothek.core.uuid import gen_uuid
+from kartothek.io.eager import read_dataset_as_metapartitions
 from kartothek.io_components.docs import default_docs
+from kartothek.io_components.index import update_indices_from_partitions
 from kartothek.io_components.metapartition import (
     MetaPartition,
     parse_input_to_metapartition,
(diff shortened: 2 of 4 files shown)


kartothek/io/dask/delayed.py
```diff
@@ -582,3 +585,41 @@ def store_delayed_as_dataset(
...
+    A dask.delayed computation object.
+    """
+    ds_factory = _ensure_factory(
+        dataset_uuid=dataset_uuid,
+        store=store,
+        factory=factory,
+        load_dataset_metadata=False,
+    )
+
+    mps = list(read_dataset_as_metapartitions__iterator(factory=ds_factory))
```
lr4d: Use `kartothek.io.eager.read_dataset_as_metapartitions` if you want a list of metapartitions

kartothek/io/dask/delayed.py
```diff
@@ -582,3 +585,41 @@ def store_delayed_as_dataset(
         metadata_merger=metadata_merger,
         metadata_storage_format=metadata_storage_format,
     )
+
+
+@default_docs
+def build_dataset_indices__delayed(
+    store, dataset_uuid, columns, blocksize=None, factory=None
```
lr4d: is `blocksize` documented? If not, can you do so?

kartothek/io_components/index.py
```diff
@@ -18,6 +18,8 @@ def update_indices_from_partitions(partition_list, dataset_metadata_factory):
     in an additional build index step (by the build_dataset_indices__pipeline) which should be used after
     updating partitions of a dataset.
     """
+    # collect potential iterators
+    partition_list = list(partition_list)
```
lr4d: Is it possible to avoid this line with a change in another part of the code?

kartothek/io/dask/delayed.py
```diff
@@ -582,3 +585,45 @@ def store_delayed_as_dataset(
...
+    mps = read_dataset_as_metapartitions(factory=ds_factory)
+
+    return (
+        db.from_sequence(seq=mps, partition_size=partition_size)
+        .map(MetaPartition.build_indices, columns=columns)
+        .map(MetaPartition.remove_dataframes)
+        .reduction(list, list, split_every=False, out_type=db.Bag)
+        .flatten()
+        .map_partitions(list)
+        .map_partitions(
```
fjetter: Isn't the output type of this function a dask.Bag, i.e. for a true delayed interface we should call `to_delayed`? I aggree with you that this is an implementation detail but considering the that we distinguish these modules by what kind of dask collection they return we could offer both functions (where the delayed version simply calls build_dataset_indices_bag().to_delayed)

kartothek/io/dask/delayed.py
```diff
@@ -582,3 +585,45 @@ def store_delayed_as_dataset(
         metadata_merger=metadata_merger,
         metadata_storage_format=metadata_storage_format,
     )
+
+
+@default_docs
+def build_dataset_indices__delayed(
```
fjetter: Not a strong opinion but we could, ofc, add _better_ names, e.g. `build_secondary_indices__delayed`

We (I) shouldn't have put the `dataset` part in every function name :)

tests/io/dask/delayed/test_index.py
```diff
@@ -0,0 +1,20 @@
...
+import dask
+import pytest
+
+from kartothek.io.dask.delayed import build_dataset_indices__delayed
+from kartothek.io.testing.index import *  # noqa: F4
+
+
+def _build_indices(*args, **kwargs):
+    tasks = build_dataset_indices__delayed(*args, **kwargs)
+    s = pickle.dumps(tasks, pickle.HIGHEST_PROTOCOL)
```
fjetter: why pickle stuff? We're using distributed for our test setup and there should be no need to test serializability _manually_

fjetter:
Considering that the return type is a dask it's more than an implementation detail. I agree that there is only a minor difference between the bag and delayed interface but I believe we should stick to it, let alone for the fact that dask/distributed applied different kind of optimizations for the different dask collections.

tests/io/dask/delayed/test_index.py
crepererum: distributed uses "pickle with cloudpickle" fallback. However, cloudpickle (the fallback) is massively more inefficient. I'm OK when an average users writes bags/graph that need to be cloudpickled (e.g. due to nested classes/functions or lambdas) but I am expecting ktk to be efficient. These roundtrip tests are used for other dask tasks as well.

kartothek/io/dask/delayed.py
crepererum: It's actually aligned with the eager version. If we rename it here, we should also rename the other.

kartothek/io/dask/delayed.py
crepererum: Then I would move this to the our bag submodule entirely because bags use a different optimizer which can make a huge difference, especially for the reduction and the map calls.

kartothek/io/dask/delayed.py
fjetter: I know it's aligned and you are right. Let's keep it as is and do a proper global sane renaming once we do a next major release (maybe)

kartothek/io/dask/delayed.py
fjetter: Well if you call `build_dataset_indices_bag().to_delayed(optimize=False)` you fall back to the unoptimized / pure delayed graph. My point is simply that since we introduced these modules we should make use of them.

tests/io/dask/delayed/test_index.py
fjetter: Fair enough. Can you put a comment in there? Maybe we can add a test like this to all pipelines (not this PR)

crepererum: Let's do the generic pipeline stuff later, since this would require some more glue code and some testing that it works. I have some ideas on how to wire this up, but this takes some time.

kartothek/io/dask/delayed.py
crepererum: I've moved it.

PR #78, opened 2019-06-28
remove toolz test dep

crepererum:
toolz was only used in a single place for dict comparison. The same
comparison could just be done using a place `assert` which would be
equally helpful due to pytest's assertion rewrite mechanism (just use
`pytest -vvv`).

### DIFF
--- a/kartothek/io/testing/index.py
+++ b/kartothek/io/testing/index.py
@@ -2,7 +2,6 @@
 
 import pandas as pd
 import pytest
-from toolz.dicttoolz import valmap
 
 from kartothek.core.factory import DatasetFactory
 from kartothek.core.index import ExplicitSecondaryIndex
@@ -10,12 +9,6 @@ from kartothek.io.eager import (
     read_dataset_as_metapartitions,
     store_dataframes_as_dataset,
 )
-
-
-def assert_index_dct_equal(dict1, dict2):
-    dict1 = valmap(sorted, dict1)
-    dict2 = valmap(sorted, dict2)
-    assert dict1 == dict2
 
 
 def test_build_indices(store_factory, metadata_version, bound_build_dataset_indices):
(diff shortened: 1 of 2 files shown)


lr4d:
Note that the dictionary values must be `sorted` in order to compare them properly. I initially did this using a comprehension, @fjetter recommended me to use the more readable `toolz` valmap function at the time

crepererum:
I'm pretty sure pytest treats dicts an unsorted by default and shows you the set difference for keys or the value difference for "same keys with different values". I actually compared loads of dicts using pytest and never had this issue.

fjetter:
Yes, I expect this build to break. In fact, I'd like to use toolz as a prod requirement since I believe we can replace a lot of our dict manipulation logic by (cy)toolz and potentially speed even up a few things. If the build turns out to be green, I'm fine with removing it

kartothek/io/testing/index.py
```diff
@@ -40,7 +33,7 @@ def test_build_indices(store_factory, metadata_version, bound_build_dataset_indi
     # Assert indices are properly created
     dataset_factory = DatasetFactory(dataset_uuid, store_factory, load_all_indices=True)
     expected = {2: ["cluster_1", "cluster_2"], 3: ["cluster_2"], 1: ["cluster_1"]}
-    assert_index_dct_equal(expected, dataset_factory.indices["p"].index_dct)
+    assert expected == dataset_factory.indices["p"].index_dct
```
lr4d: @crepererum What I meant to say is that `dataset_factory.indices["p"].index_dct` and `expected` here may have a different order for keys as well as values (the list of partition labels is not guaranteed to be ordered)

crepererum:
Since when does the dict comparison care about the order? OK, that's new, but at least I've learned something.

> In fact, I'd like to use toolz as a prod requirement
OK, then I see no point in removing it here. I just wanted to remove a "just for this single function" dep which could actually also have been implemented in a single line of slightly more verbose python.

Closing.

kartothek/io/testing/index.py
lr4d: The dict comparison does indeed not care about order of keys, but the values will be interpreted as different because we are comparing e.g. `["cluster_1", "cluster_2"] == ["cluster_2", "cluster_1"]`

crepererum: ah 👍 
But still, if @fjetter wants to extend the usage of toolz, there is no point in having a "remove it from tests" PR

PR #79, opened 2019-06-28
Issue 24 dispatch by subset partition keys

NeroCorleone:


kartothek/io_components/read.py
```diff
@@ -47,14 +48,21 @@ def dispatch_metapartitions_from_factory(
...
+                raise ValueError("Nope.")
+            part_keys = dispatch_by
+        else:
+            part_keys = dataset_factory.partition_keys
+
         # Build up a DataFrame that contains per row a Partition and its
         # primary index columns.
         base_df = None
-        for part_key in dataset_factory.partition_keys:
+        for part_key in part_keys:
```
fjetter: Can you try to replace this loop with https://github.com/JDASoftwareGroup/kartothek/blob/a27ce44986985b9ed7e110472510d28c3cba7f70/kartothek/core/dataset.py#L339
Should go as
```
base_df = dataset_factory. get_indices_as_dataframe(columns=part_keys)
```
And see if this will do the trick (functionally)

fjetter: IF this is the case we need to look into performance since allowed_labels is not respected by get_indices_as_dataframe but let's first see if we can get rid of the duplication

kartothek/io_components/read.py
```diff
@@ -65,8 +73,11 @@ def dispatch_metapartitions_from_factory(
...
 
-        # Group the resulting MetaParitions by partition keys
-        merged_partitions = base_df.groupby(dataset_factory.partition_keys)
+        # Group the resulting MetaParitions by partition keys or a subset of those keys
+        if dispatch_by:
+            merged_partitions = base_df.groupby(dispatch_by)
+        else:
+            merged_partitions = base_df.groupby(dataset_factory.partition_keys)
         merged_partitions = merged_partitions["__partition__"].unique()
         for row, labels in merged_partitions.iteritems():
```
fjetter: This is where you iterate over the groupby. Iterating (potentially w/out the iteritems) yields a tuple (group_name, group)
where group_name is the unique value of the groupby statement (e.g. B=1) and the group is the DF

kartothek/io_components/read.py
```diff
@@ -65,8 +80,11 @@ def dispatch_metapartitions_from_factory(
         assert base_df is not None
         base_df = cast(pd.DataFrame, base_df)
 
-        # Group the resulting MetaParitions by partition keys
-        merged_partitions = base_df.groupby(dataset_factory.partition_keys)
+        # Group the resulting MetaParitions by partition keys or a subset of those keys
+        if dispatch_by:
```
fjetter: At this point you should be able to use `part_keys` as the grouper since you already checked above in which mode we're operating

kartothek/io_components/read.py
```diff
@@ -47,14 +48,28 @@ def dispatch_metapartitions_from_factory(
...
+            "Please only provide the `dispatch_by` argument. "
+        )
+
+    if dispatch_by or concat_partitions_on_primary_index:
         if dataset_factory.explicit_partitions:
             dataset_factory = dataset_factory.load_partition_indices()
 
+        if dispatch_by:
+            if not set(dispatch_by).issubset(dataset_factory.partition_keys):
+                raise ValueError("Nope.")
```
fjetter: With the logical predicates you can relax this test to: all of the columns need to be an index column (not just restricted to partition cols)

kartothek/io_components/read.py
fjetter: You need to extract the logical predicate from row here, I believe

docs/io/examples.rst
```diff
@@ -225,6 +225,9 @@ Read
     import pandas as pd
     from kartothek.io.dask.delayed import read_dataset_as_delayed
 
-    tasks = read_dataset_as_delayed(dataset_uuid="MyFirstDatasetIter", store=store_factory)
+    tasks = read_dataset_as_delayed(
+        dataset_uuid="MyFirstDatasetDask",
+        store=store_factory
+        )
```
fjetter: This will prob cause travis to fail due to improper formatting. consider installing the pre commit hooks

kartothek/io/dask/delayed.py
```diff
@@ -298,9 +299,17 @@ def read_dataset_as_delayed_metapartitions(
         concat_partitions_on_primary_index=concat_partitions_on_primary_index,
         label_filter=label_filter,
         predicates=predicates,
+        dispatch_by=dispatch_by,
     )
 
-    if concat_partitions_on_primary_index:
+    if concat_partitions_on_primary_index and dispatch_by:
```
fjetter: You should move this check to the `dispatch` function since this is applicable for all pipelines

kartothek/io/testing/read.py
```diff
@@ -366,6 +366,40 @@ def test_read_dataset_as_dataframes_concat_primary(
     pdt.assert_frame_equal(expected_df, result_df, check_like=True)
 
 
+def test_read_dataset_as_dataframes_dispatch_by(
+    store_factory, custom_read_parameters, bound_load_dataframes, output_type
+):
+    if output_type != "dataframe":
```
fjetter: I believe we do not need to skip the dataframe backend here

NeroCorleone: That too bad, my tests currently fail for `output_type == "metapartition"` or `output_type == "table"`. Will have to look into it.

kartothek/io_components/read.py
NeroCorleone: True, thanks.

kartothek/io_components/read.py
NeroCorleone: Will do that when the functional part is ready. Hope that your PR did most of the job... 😁

kartothek/io/eager.py
```diff
@@ -243,6 +247,7 @@ def read_table(
     dates_as_object=False,
     predicates=None,
     factory=None,
+    dispatch_by=None,
```
fjetter: TODO: Maybe not necessary for this function

fjetter: Add deprecation warning to remove `concat_partitions_on_primary_index`

fjetter:
Please add a changelog entry and take care of the last comment. Afterwards you can hit ready-for-review and we can let others have a look

kartothek/io_components/metapartition.py
```diff
@@ -622,7 +652,8 @@ def load_dataframes(
             instead of using ``np.datetime64`` to preserve their type. While
             this improves type-safety, this comes at a performance cost. Only
             works for metadata version >= 4.
-
+        predicates: nested list of tuples
+            TODO
```
NeroCorleone: Do we need docs here?

fjetter: Yes, the default docs decorator should work here as well. Not sure about the order of the decorators, though.
my best guess would be:

```
    @_apply_to_list
    @default_docs
    def load_dataframes
```

kartothek/io/eager.py
```diff
@@ -236,7 +240,6 @@ def read_table(
     store=None,
     table=None,
     columns=None,
-    concat_partitions_on_primary_index=False,
```
fjetter: No, don't remove it. Please add a deprecation warning and keep it for now. We'll remove this with the next major release

kartothek/io_components/read.py
```diff
@@ -49,6 +49,13 @@ def dispatch_metapartitions_from_factory(
         )
         dispatch_by = dataset_factory.partition_keys
 
+    if dispatch_by and not set(dispatch_by).issubset(
+        set(dataset_factory.index_columns)
```
lr4d: Why cast to `set` here if you have already done so in the method `index_columns`?

fjetter: 1. Because it doesn't matter and this way works no matter what index_columns returns
2. This code is older than the index_columns code ;)

kartothek/io/testing/read.py
```diff
@@ -366,6 +368,98 @@ def test_read_dataset_as_dataframes_concat_primary(
...
+        df_generator=partitions,
+        store=store_factory,
+        dataset_uuid="partitioned_uuid",
+        metadata_version=4,
+        partition_on=["A", "B"],
+        secondary_indices=["C"],
+    )
+    for dispatch_by in chain.from_iterable(
+        combinations(x, 2) for x in permutations(["A", "B", "C"])
+    ):
```
lr4d:
```
set(permutations(("A", "B", "C"), 2)) == set(chain.from_iterable(
        combinations(x, 2) for x in permutations(["A", "B", "C"])
    ))
Out[170]: True
```

tests/io_components/test_metapartition.py
```diff
@@ -87,6 +87,53 @@ def test_store_single_dataframe_as_partition_no_metadata(store, metadata_version
...
+    predicates = None
+    loaded_mp = meta_partition.load_dataframes(store=store, predicates=predicates)
+    data = {
+        "core": pd.DataFrame(
+            {"P": [5, 6, 7, 8, 9], "L": [5, 6, 7, 8, 9], "TARGET": [15, 16, 17, 18, 19]}
+        ).set_index(np.arange(5, 10))
+    }
+    pdt.assert_frame_equal(loaded_mp.data["core"], data["core"])
+
+    predicates = [[("L", ">", 2), ("TARGET", "<", 17)]]
```
lr4d: How about testing this for "L < 6" to check if both predicates are actually being used?

kartothek/io_components/read.py
```diff
@@ -65,12 +84,18 @@ def dispatch_metapartitions_from_factory(
...
+        merged_partitions = base_df.groupby(
+            by=list(dispatch_by), sort=False, as_index=False
+        )
+        for group_name, group in merged_partitions:
+            if not isinstance(group_name, tuple):
+                group_name = (group_name,)
             mps = []
-            for label in labels:
+            logical_conjunction = list(
+                zip(dispatch_by, ["=="] * len(dispatch_by), group_name)
```
lr4d: This looks fishy to me, perhaps I'm missing something, is `len(group_name) == len(dispatch_by)`?

NeroCorleone: yes, because `merged_partitions` is a grouped by `dispatch_by`:
`merged_partitions = base_df.groupby(by=list(dispatch_by), sort=False, as_index=False)`
and `group_name contains` the names of the columns we have grouped by. 
You can check out `test_read_dataset_as_dataframes_dispatch_by_multi_col` where the case of multiple columns is tested and `len(dispatch_by) > 1`

kartothek/io/testing/read.py
```diff
@@ -366,6 +367,76 @@ def test_read_dataset_as_dataframes_concat_primary(
...
+        {"A": [2, 2], "B": [10, 10], "C": [1, 2], "Content": ["cluster4", "cluster4"]}
+    )
+    clusters = [cluster1, cluster2, cluster3, cluster4]
+    partitions = [{"data": [("data", c)]} for c in clusters]
+
+    store_dataframes_as_dataset__iter(
+        df_generator=partitions,
+        store=store_factory,
+        dataset_uuid="partitioned_uuid",
+        metadata_version=4,
```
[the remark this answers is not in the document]
NeroCorleone: Indeed, why not... This is what happens when you copy-paste tests :)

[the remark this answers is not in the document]
fjetter: Even better. Use the `metadata_version` fixture

kartothek/io/testing/read.py
```diff
@@ -366,6 +368,96 @@ def test_read_dataset_as_dataframes_concat_primary(
...
```
fjetter: metadata version fixture

kartothek/io_components/metapartition.py
```diff
@@ -1383,10 +1409,19 @@ def merge_metapartitions(metapartitions, label_merger=None, metadata_merger=None
         LOGGER.debug("Merging metapartitions")
         data = defaultdict(list)
         new_metadata_version = -1
+        logical_conjunction = None
+
         for mp in metapartitions:
             new_metadata_version = max(new_metadata_version, mp.metadata_version)
             for label, df in mp.data.items():
                 data[label].append(df)
+            if mp.logical_conjunction or logical_conjunction:
```
lr4d: This `or` appears to be unused as you have just set `logical_conjunction = None` a few lines back

lr4d: You can probably simplify the if-else statement below also

NeroCorleone: To be honest, I do not understand what this should do functionally. Could you maybe clarify @fjetter ? We iterate over the `metapartitions` and determine the value of `logical_conjunction` within this iteration -- this would be the last not-None value of `mp.logical_conjunction`. Or, if `mp.logical_conjunction` is `None` for all `mp`, then `logical_conjunction` would be `None`?

fjetter: It is not unused since the instance attribute and the variable are not the same. The variable is used to track the unique values and yes, there are many ways to do so. Since the logical_conjunction is not a trivial data type (it's a list) the straight forward approaches, e.g. using a set, doesn't work w/out manipulation of the values.
Since this is a matter of style, I don't see a reason why this should block a merge

kartothek/io_components/metapartition.py
```diff
@@ -562,7 +594,9 @@ def _apply_partition_key_predicates(self, table, indices, split_predicates):
                     return None
         return filtered_predicates
 
+    @default_docs
```
fjetter: This should be fixed. There should only be one decorator

kartothek/io_components/metapartition.py
lr4d: You're right. Should've looked into it more before requesting change

kartothek/io_components/metapartition.py
fjetter: I took care of it. Will merge as soon as build finishes

Issue #81, opened 2019-07-04
Kartothek breaks with pyarrow 0.14.0


fjetter:
Not surprised but a bit disappointed, we have breakage with the new arrow release, e.g. https://travis-ci.org/fjetter/kartothek/builds/554210679

fjetter:
There are even segfaults 🎉

fjetter:
https://issues.apache.org/jira/browse/ARROW-5873

fjetter:
https://issues.apache.org/jira/browse/ARROW-5878

fjetter:
Ihttps://issues.apache.org/jira/browse/ARROW-5889

fjetter:
I suggest to postpone arrow 0.14.x support until 0.14.1 / https://issues.apache.org/jira/browse/ARROW-5878 is resolved. This seemed to have triggered a discussion regarding the parquet spec and I think we should sit this one out

crepererum:
Fair, let's blacklist 0.14.0 then in our requirements.

byjott:
So 0.14.1 has been released. What's the status with that version?

fjetter:
There is a WIP https://github.com/JDASoftwareGroup/kartothek/pull/108

xhochy:
We support 0.14.1 and explicitly omit 0.14, closing this as won't fix.

PR #83, opened 2019-07-05
WIP pyarrow 0.14.0

fjetter:
The segfault segfault seems unrelated

This is probably the first pyarrow release which let's us delete code. Our timestamp unit calculation for predicate pushdown seems to be obsolete now.

So far, I uncovered one bug in pyarrow (timestamps without timezone info are assumed to be UTC). Still need to file a bug

reference: #81


kartothek/serialization/_arrow_compat.py
```diff
@@ -11,6 +14,30 @@ def _fix_pyarrow_0130_table(table):
     return table
 
 
+def _fix_pyarrow_0140_table(table):
```
fjetter: question here would be if we need to fix this in the write path. We should notice this once I enable all arrow versions and we'll see if we're forward compatible

crepererum:
We should probably use [pytest-faulthandler](https://github.com/pytest-dev/pytest-faulthandler) for all ktk tests so segfaults result in more useful reports.

crepererum:
BTW: I've just seen that since pytest 5.0, we don't need pytest-faulthandler since it's now shipped by pytest itself and is even enabled by default.

fjetter:
pytest-faulthandler is on by default in pytest==5.0.0

The segfault is caused by another race condition while using the same dataframe multiple times. I.e. there is a nother reference counting issue in arrow

fjetter:
The actually bigger issue is not the segafault but the timezone handling which was introduced in the latest version. both parquet and arrow now track the timezone of a timestamp but the behavior is a bit unstable atm. I'll create upstream issues and afterwards we need to decide whether or not we whitelist arrow 14 or wait for a patch release

crepererum:
@fjetter I think we can update our minimum pytest requirements.

tests/core/test_common_metadata.py
```diff
@@ -56,6 +58,7 @@ def test_store_schema_metadata(store, df_all_types):
         pa.field("byte", pa.binary()),
         pa.field("date", pa.date32()),
         pa.field("datetime64", pa.timestamp("us")),
```
fjetter: Any idea why this is `us`? Shouldn't the default accuracy be `ns`?? cc @crepererum @xhochy

fjetter:
The segfault seems to prevail even after copying the test data. Good thing: It crashes always at the exact same line of code https://github.com/JDASoftwareGroup/kartothek/blob/c47e52116e2dc726a74d7d6b97922a0252722ed0/tests/io_components/test_metapartition.py#L259

The only place where we interact with native code in that test is the `make_meta` call in `MetaPartition.__init__` and I suspect that something is wrong with the schema inference we're doing there.

However, cannot reproduce locally, yet.

It's worth mentioning that I encountered a segmentation fault in this test before on python 2.7 and a lower pyarrow version. I didn't investigate since we dropped py2 support and I hoped we would be spared

tests/core/test_common_metadata.py
crepererum: I think you're right:

```python
>>> import pandas as pd
>>> import pyarrow as pa
>>> df = pd.DataFrame({"x": [pd.Timestamp("now")]})
>>> pa.Schema.from_pandas(df)
x: timestamp[ns]
metadata
--------
OrderedDict([(b'pandas',
              b'{"index_columns": [{"kind": "range", "name": null, "start": '
              b'0, "stop": 1, "step": 1}], "column_indexes": [{"name": null,'
              b' "field_name": null, "pandas_type": "unicode", "numpy_type":'
              b' "object", "metadata": {"encoding": "UTF-8"}}], "columns": ['
              b'{"name": "x", "field_name": "x", "pandas_type": "datetime", '
              b'"numpy_type": "datetime64[ns]", "metadata": null}], "creator'
              b'": {"library": "pyarrow", "version": "0.13.0"}, "pandas_vers'
              b'ion": null}')])
```

crepererum:
@fjetter https://docs.travis-ci.com/user/running-build-in-debug-mode/ could help us to attach a debugger. I'll have a look at this whole thing later today and try to get some useful stuff out of it.

fjetter:
A question which is also popping up right now: Do we want to be forward compatible? Is a pyarrow==0.13.0 reader supposed to be able to read a file which was produced by pyarrow==0.14.0?

This issue currently pops up since timestamps without timezones no longer store a logical parquet type which can be understood by older versions

pyarrow 0.13.0 parquet schema
```
datetime64: INT64
datetime64_ts: INT64 TIMESTAMP_MICROS
```

pyarrow 0.14.0 parquet schema
```
datetime64: INT64 Timestamp(isAdjustedToUTC=false, timeUnit=microseconds)
datetime64_ts: INT64 Timestamp(isAdjustedToUTC=true, timeUnit=microseconds)
```

i.e. pyarrow 0.13.0 reads the column of the 0.14.0 release as an int. While we may report this upstream as well the questions is still something we should think about

xhochy:
> A question which is also popping up right now: Do we want to be forward compatible? Is a pyarrow==0.13.0 reader supposed to be able to read a file which was produced by pyarrow==0.14.0?

Yes, a `pyarrow=0.13.0` should be able to read Parquet files produced from `pyarrow=0.14.0`, if not this is a flaw that should be fixed upstream or at least upstream should provide a flag to produce the correct files.

tests/core/test_common_metadata.py
fjetter: I believe it is `us` because we use `coerce_timestamps="us"` in the write path but honestly, I can't remember why.

fjetter:
I suggest to close this one for now. Discussion should continue on #81

PR #87, opened 2019-07-11
fix handling of date predicates

crepererum:


crepererum:
@fjetter w/o the patch, we actually crash (because it generates invalid, empty predicate lists), because the index evaluation during MP dispatch actually matches certain partitions while the MP load_dataframes part does not. No matter what we think about strictness, this is a bug. 

The test case shows that matching on the same type (note that DF type and predicate type are in fact identical) works now w/o any issues. So I would consider this bug as fixed.

If we want to be strict, we should probably fix the entire index DF reconstruction. Namely, the index type handling currently does not care about "dates as objects" at all, neither in the the MP dispatch nor in the DF loading. So I would agree to create a follow-up issue to streamline this, but this is a larger work.

Also note that this is a 3.0=>3.1 regression.

crepererum:
I've found a more streamlined solution but I am pretty sure we should add proper test coverage for the mixed case (aka strictness), because I think there might be more bugs in the MP dispatch.

PR #88, opened 2019-07-12
Enable filter_array_like properly for categoricals

fjetter:
Categoricals have an issue with the numpy issubdtype, hence the mapping to the kind attribute of the arrays/type

To properly support the predicate eval range, I added a section about the handling of the categoricals. IMHO we should support it but highlight that it may come with undesired negative side effects.

### DIFF
--- a/tests/serialization/test_filter.py
+++ b/tests/serialization/test_filter.py
@@ -1,0 +1,68 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+from kartothek.serialization import filter_array_like
+
+
+@pytest.fixture(
+    params=[
+        np.array(range(10)),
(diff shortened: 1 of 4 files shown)


tests/serialization/test_filter.py
```diff
@@ -0,0 +1,68 @@
...
+    ix = 3
+    value = array_like[ix]
+    res = filter_array_like(array_like, "==", value)
+
+    assert all(array_like[res] == array_like[ix])
+
+
+def test_filter_array_like_larger_eq(array_like):
+    ix = 3
+    value = array_like[3]
```
lr4d: nitpick: use `ix` instead of `3`

PR #89, opened 2019-07-12
Type stability pred pushdown

### LINKED ISSUE #66: Index doesn't raise if queried with wrong data type
Our indices are type stable and we require the input type for a given query to be of the exact type the index is using. This is important to ensure valid predicate evaluation. The current behavior is inconsistent and doesn't raise as expected if the queries type is invalid

```python
import pyarrow as pa
from kartothek.core.index import PartitionIndex, ExplicitSecondaryIndex, IndexBase

ind = IndexBase(
    column='col',
    index_dct={
        1234: ["part"]
    },
    dtype=pa.int64()
)

# expected to raise, returns empty set
ind.eval_operator("==", "1234")
# expected to raise but casts the value and returns {"part"}
ind.eval_operator("==", 1234.0)

# only valid vall
ind.eval_operator("==", 1234)
```

fjetter:
Closes #66

### DIFF
--- a/kartothek/serialization/_generic.py
+++ b/kartothek/serialization/_generic.py
@@ -239,12 +239,16 @@ def _ensure_type_stability(array_like, value, strict_date_types, require_ordered
     value_dtype = pd.Series(value).dtype
 
     if require_ordered and pd.api.types.is_categorical(array_like):
-        array_value_type = array_like.cat.categories.dtype
-        if array_like.cat.categories.is_monotonic:
-            array_like = array_like.cat.as_ordered()
+        if isinstance(array_like, pd.Categorical):
+            categorical = array_like
         else:
-            array_like = array_like.cat.reorder_categories(
-                array_like.cat.categories.sort_values(), ordered=True
+            categorical = array_like.cat
+        array_value_type = categorical.categories.dtype
+        if categorical.categories.is_monotonic:
+            array_like = categorical.as_ordered()
+        else:
+            array_like = categorical.reorder_categories(
+                categorical.categories.sort_values(), ordered=True
             )
     else:
         array_value_type = array_like.dtype
--- a/kartothek/serialization/_generic.py
+++ b/kartothek/serialization/_generic.py
@@ -4,8 +4,6 @@
 This module contains functionality for persisting/serialising DataFrames.
 """
 
-
-import datetime
 from typing import Dict
 
 import numpy as np
@@ -203,6 +201,42 @@ def filter_df_from_predicates(df, predicates, strict_date_types=False):
     return df[indexer]
 
 
+def _handle_categorical_data(array_like, require_ordered):
+    if require_ordered and pd.api.types.is_categorical(array_like):
+        if isinstance(array_like, pd.Categorical):
+            categorical = array_like
+        else:
+            categorical = array_like.cat
+        array_value_type = categorical.categories.dtype
+        if categorical.categories.is_monotonic:
+            array_like = categorical.as_ordered()
+        else:
+            array_like = categorical.reorder_categories(
+                categorical.categories.sort_values(), ordered=True
+            )
+    else:
+        array_value_type = array_like.dtype
+    return array_like, array_value_type
+
+
+def _handle_null_arrays(array_like, value_dtype):
+    # NULL types might not be preserved well, so try to cast floats (pandas default type) to the value type
+    # Determine the type using the `kind` interface since this is common for a numpy array, pandas series and pandas extension arrays
+    if array_like.dtype.kind == "f" and np.isnan(array_like).all():
+        if array_like.dtype.kind != value_dtype.kind:
+            array_like = array_like.astype(value_dtype)
+    return array_like, array_like.dtype
+
+
+def _handle_timelike_values(array_value_type, value, value_dtype, strict_date_types):
+    if is_list_like(value):
+        value = [pd.Timestamp(val).to_datetime64() for val in value]
+    else:
+        value = pd.Timestamp(value).to_datetime64()
+    value_dtype = pd.Series(value).dtype
+    return value, value_dtype
+
+
 def _ensure_type_stability(array_like, value, strict_date_types, require_ordered):
     """
     Ensure that the provided value and the provided array will have compatible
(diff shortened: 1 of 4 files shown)


kartothek/serialization/_generic.py
```diff
@@ -203,6 +203,49 @@ def filter_df_from_predicates(df, predicates, strict_date_types=False):
...
+    if array_like.dtype.kind == "f" and np.isnan(array_like).all():
+        if array_like.dtype.kind != value_dtype.kind:
+            array_like = array_like.astype(value_dtype)
+            array_value_type = array_like.dtype
+
+    type_comp = {value_dtype.kind, array_value_type.kind}
+    if (
+        len(type_comp) > 1
+        # UINT and INT are accepted
+        and not type_comp == {"u", "i"}
```
fjetter: This section is a bit weird. First of all, previously we allowed comparison between int and uint. with strict type checking this wouldn't be allowed which is why I included this section. 
The following two cases are dealing with the different kinds of string types there are in numpy (O | object, S | (byte-)string, U | Unicode).

The way it is implemented atm is essentially how we handled it before and I believe between these types we do have valid comparison operators which is why I would understand to keep these exceptions. I don't have a strong opinion about how strict we should be here and am open for discussion :)

tests/serialization/test_filter.py
```diff
@@ -66,3 +68,23 @@ def test_filter_array_like_categoricals(op, expected, cat_type):
...
+        (datetime.date(2019, 1, 1), 1),
+        (datetime.datetime(2019, 1, 1), 1),
+        (datetime.datetime(2019, 1, 1), datetime.date(2019, 1, 1)),
+        (datetime.date(2019, 1, 1), datetime.datetime(2019, 1, 1)),
+    ],
+)
+@pytest.mark.parametrize("op", ["==", "!=", "<", "<=", ">", ">="])
+def test_raise_on_type(value, filter_value, op):
+    array_like = pd.Series([value])
+    with pytest.raises(TypeError):
```
crepererum: please assert the exception message here, otherwise we might catch some other ktk bug instead

tests/core/test_index.py
```diff
@@ -558,6 +558,17 @@ def test_eval_operators(index_data, op, value, expected):
     assert result == expected
 
 
+def test_eval_operators_type_safety():
+    # gh66
+    ind = IndexBase(column="col", index_dct={1234: ["part"]}, dtype=pa.int64())
+    with pytest.raises(TypeError):
```
crepererum: please assert the exception message here, otherwise we might catch some other ktk bug instead

kartothek/serialization/_generic.py
```diff
@@ -203,6 +203,49 @@ def filter_df_from_predicates(df, predicates, strict_date_types=False):
     return df[indexer]
 
 
+def _ensure_type_stability(array_like, value, strict_date_types):
```
crepererum: please add a docstring describing this function. Even though it is an internal function, it is quite complex and devs should be able to understand it easiliy

kartothek/serialization/_generic.py
lr4d: I don't have too much authority about this, but opening the door to comparing arrays with different types of only a _selection_ of type combinations (and not others) seems ambiguous.
For example, if I would be a user, and compare arrays of type object with string, successfully; I might also expect to be able to compare, for example, integers with booleans, or integers with floats.
My suggestion would thus be to either:
 - A) Have a clear definition of what types should be able to be compared to each other (perhaps there is some appropriate numpy/pandas function which does something similar and we could use?)
- B) Strictly disallow comparisons of different types.

crepererum: @fjetter what's actually the reason we're allowing non-strict comparisons?

fjetter: There is none. I essentially fixed the test cases and figured we should have a discussion about this.
At least for these types there is a non ambiguous definition of equality. 

FYI I believe the string cases are quite artificial. I only. Caught it with hypothesis and I'm not sure if we can construct these with parquet tbh

crepererum: Looking at the algorithm again, I think we should keep it as it is proposed here. The main issue (as far as I understood) is that Python doesn't have proper scalar types (e.g. unsigned ints) and therefore we must be a little bit forgiving when comparing a Python scalar with a NumPy array. So I would say: let's use this PR and add a tracking issue to remove the `strict_date_types` param and hard-default to `strict_date_types=True`, because I think there is no good reason for `strict_date_types=False` (aka being more forgiving than really required).

fjetter: Yes it boils down to python not having a proper scalar typing system and I'd like to not have to do something like `("uint_col", "==", np.uint(1))`, especially considering that _our type model_ is arrow and not numpy

> remove the strict_date_types param 

I believe we'll need the parameter since we support `date_as_object` which affects this setting if we want to keep the definition of the predicates independent from this toggle. 
I agree that we should switch the default behavior

crepererum: > I believe we'll need the parameter since we support `date_as_object` which affects this setting
A right, I forgot. Out of interest: did I understand it correctly that this setting isn't required anymore with pyarrow 0.14?

tests/serialization/test_filter.py
```diff
@@ -66,3 +69,72 @@ def test_filter_array_like_categoricals(op, expected, cat_type):
...
+
+    op = "<"
+    for col in df.columns:
+        value = df[col][ix]
+        predicates = [[(col, op, value)]]
+        actual = filter_df_from_predicates(df, predicates)
+        if pd.api.types.is_categorical(df[col]):
+            df[col] = df[col].astype(df[col].cat.as_ordered().dtype)
+        expected = df[df[col] < value]
+        pdt.assert_frame_equal(actual, expected, check_categorical=False)
```
lr4d: Nitpick: can these for loops be compacted into a single nested loop? Don't know about the "correctness" of using `eval` but it could be suitable for this case ( as in `expected = eval(f"df[df[col] {op} value])"`)

lr4d: Also, just wondering, do `>=` and `<=` not apply here?

fjetter: I usually refrain from using eval/exec but in this case this is probably worth an exception.

> Also, just wondering, do >= and <= not apply here?

Yes but I didn't want to copy more code. The important aspect here was that ordered operators where tested. With the eval code I added them to them loop

tests/serialization/test_filter.py
```diff
@@ -66,3 +69,47 @@ def test_filter_array_like_categoricals(op, expected, cat_type):
...
+            "B": ["A", "B"] * 5,
+            "C": pd.Series(["X", "Y"] * 5).astype("category"),
+            "D": pd.Series([datetime.date(2019, 1, 1), datetime.date(2019, 1, 2)] * 5),
+            "E": [datetime.datetime(2019, 1, 1), datetime.datetime(2019, 1, 2)] * 5,
+        }
+    )
+
+    ops = ["==", "!=", ">=", "<=", ">", "<"]
+    ix = 4
+    for op in ops:
```
crepererum: please use `parametrize` for these loops

tests/serialization/test_filter.py
```diff
@@ -66,3 +69,46 @@ def test_filter_array_like_categoricals(op, expected, cat_type):
...
+            "A": range(10),
+            "B": ["A", "B"] * 5,
+            "C": pd.Series(["X", "Y"] * 5).astype("category"),
+            "D": pd.Series([datetime.date(2019, 1, 1), datetime.date(2019, 1, 2)] * 5),
+            "E": [datetime.datetime(2019, 1, 1), datetime.datetime(2019, 1, 2)] * 5,
+        }
+    )
+
+    ix = 4
+    for col in df.columns:
```
crepererum: same, I would use a `parametrize` here as well, because it simplifies the debugging in case of a test failure.

PR #93, opened 2019-07-17
Add functionality to read dataset using dask bag

NeroCorleone:
The implementation is nearly the same as in dask.delayed. Shared
functions have been moved to kartothek.io.dask._utils


fjetter:
Can you please add a changelog entry before we merge?

PR #96, opened 2019-07-24
Add keyword for controlling partition size for dask bag

NeroCorleone:
Adding a keyword `npartition` to the read functionalities for dask's bag
backend makes it possible to fix the number of partitions when creating
the bag `db.from_sequence(some_sequence, npartitions)`.

### DIFF
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -55,6 +55,7 @@ def read_dataset_as_metapartitions_bag(
     predicates=None,
     factory=None,
     dispatch_by=None,
+    npartitions=None,
 ):
     """
     Retrieve dataset as `dask.bag` of `MetaPartition` objects.
@@ -73,7 +74,7 @@ def read_dataset_as_metapartitions_bag(
         predicates=predicates,
         dispatch_by=dispatch_by,
     )
-    mps = db.from_sequence(mps)
+    mps = db.from_sequence(mps, npartitions=npartitions)
 
     if concat_partitions_on_primary_index or dispatch_by:
         mps = mps.map(
@@ -128,6 +129,7 @@ def read_dataset_as_dataframe_bag(
     predicates=None,
     factory=None,
     dispatch_by=None,
+    npartitions=None,
 ):
     """
     Retrieve data as dataframe from a `dask.bag` of `MetaPartition` objects
@@ -146,6 +148,7 @@ def read_dataset_as_dataframe_bag(
         load_dataset_metadata=False,
         predicates=predicates,
         dispatch_by=dispatch_by,
+        npartitions=npartitions,
     )
     return mps.map(_get_data)
 
--- a/tests/io/dask/bag/test_read.py
+++ b/tests/io/dask/bag/test_read.py
@@ -1,13 +1,44 @@
 import pickle
 from functools import partial
 
+import pandas as pd
 import pytest
 
 from kartothek.io.dask.bag import (
     read_dataset_as_dataframe_bag,
     read_dataset_as_metapartitions_bag,
 )
+from kartothek.io.iter import store_dataframes_as_dataset__iter
 from kartothek.io.testing.read import *  # noqa
+
+
+def test_read_dataset_as_dataframes_npartitions(store_factory, metadata_version):
+    cluster1 = pd.DataFrame(
+        {"A": [1, 1], "B": [10, 10], "C": [1, 2], "Content": ["cluster1", "cluster1"]}
+    )
+    cluster2 = pd.DataFrame(
+        {"A": [1, 1], "B": [10, 10], "C": [2, 3], "Content": ["cluster2", "cluster2"]}
+    )
+    cluster3 = pd.DataFrame({"A": [1], "B": [20], "C": [1], "Content": ["cluster3"]})
+    cluster4 = pd.DataFrame(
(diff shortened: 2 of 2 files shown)


NeroCorleone:
By the way, I just saw that we use the `partition_size` parameter here:
https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/io/dask/bag.py#L224
Would it make more sense to make `partition_size` controllable?

NeroCorleone:
@fjetter what exactly does "calling repartition" mean?

fjetter:
https://docs.dask.org/en/latest/bag-api.html#dask.bag.Bag.repartition

crepererum:
@fjetter `repartition` likely will produce a different graph because the repartition operation is not during planning time (on the client) but during execution time, therefore wasting time and resources. So why not creating the bag in the correct way in the first place?

kartothek/io/dask/bag.py
```diff
@@ -146,6 +148,7 @@ def read_dataset_as_dataframe_bag(
         load_dataset_metadata=False,
         predicates=predicates,
         dispatch_by=dispatch_by,
+        npartitions=npartitions,
```
crepererum: I would rather go for `partition_size` instead of `npartitions` for the following reason:
- the dask workers are likely limited by memory, so you want to cut down the load per worker
- you're tasks are likely being to small / to big and you want to control the balance between overhead and worker load
- the number of metapartitions is data-dependend, the worker / overhead per worker should not

fjetter:
> repartition likely will produce a different graph because the repartition operation is not during planning time (on the client) but during execution time

`repartition` is running during planing and not during execution.

> So why not creating the bag in the correct way in the first place?

I'd argue that it is exactly the same thing

My issue here is rather the fact that we expose dask keyword one at a time and I would like to keep the surface of our API minimal. I don't see the benefit of exposing these arguments for the read path since the read doesn't perform any logic after object construction other than loading the files. 
The example mentioned by @NeroCorleone regarding the index build pipeline is slightly different since the pipeline actually performs computations where these parameters have a strong impact. The same is true for any write pipeline since we are performing partition_on, build_indices, etc. For read, I don't see the benefit.

> I would rather go for partition_size instead of npartitions

What does partition size do? would partition_size=2 imply that two metpartitions would make up one dask partition?


Overall I'm not against introducing these parameters. I'm rather asking myself: 1. Do we really need it or is this something the user can do? 2. If we need it, why not pass a dictionary and expose all construction arguments? Just trying to understand the usecase...

crepererum:
> repartition is running during planing and not during execution.

No, it is not. We have some `map` operations in-between. See the following example

## Map-then-repartition
```python
db.from_sequence(range(10)).map(lambda x: x + 1).repartition(2).sum().visualize(filename="graph.png")
```

graph

## Repartition-than-map

```python
db.from_sequence(range(10)).repartition(2).map(lambda x: x + 1).sum().visualize(filename="graph.png")
```

graph

## Direct from_sequence control
```python
db.from_sequence(range(10), partition_size=5).map(lambda x: x + 1).sum().visualize(filename="graph.png")
```

graph

Furthermore, `repartition` does not allow you to specify the `partition_size` directly.


> What does partition size do? would partition_size=2 imply that two metpartitions would make up one dask partition?

yes.

fjetter:
Ok, I didn't notice the `map` in between, thanks for the visualizations.

From what I understand is that the `partition_size` would indeed be a more suitable parameter. Since the bag constructor only allows to keywords (partition_size and npartitions) we probably don't need to expose anything more sophisticated than the plain keyword (e.g. a dict to pass generic keywords would be overkill)

crepererum:
> (e.g. a dict to pass generic keywords would be overkill)

I am very much in favor of not exposing dicts. It makes everything (docs, error messages, type annotations, IDE support) just way more difficult 👍

fjetter:
I assume graph 2 and 3 would be the same after optimization which is run when computing, i.e. in the end there wouldn't be a difference. At least I would hope so... 
Either way having the optimized graph from the very start is a good thing 👍

fjetter:
> I am very much in favor of not exposing dicts

Yes, indeed. I also consider this a _last resort_ but sometimes a tool to handle increasingly complex signatures. For now this is not an option so no I'll stop arguing :)

crepererum:
> I assume graph 2 and 3 would be the same after optimization which is run when computing

Good catch, I did not optimize the graph. Sadly, this assumption does not hold neither (but I thought the same as you).

## Map-then-repartition
```python
db.from_sequence(range(10)).map(lambda x: x + 1).repartition(2).sum().visualize(filename="graph.png", optimize_graph=True)
```

graph

## Repartition-than-map
```python
db.from_sequence(range(10)).repartition(2).map(lambda x: x + 1).sum().visualize(filename="graph.png", optimize_graph=True)
```

graph

## Direct from_sequence control
```python
db.from_sequence(range(10), partition_size=5).map(lambda x: x + 1).sum().visualize(filename="graph.png", optimize_graph=True
```

graph

NeroCorleone:
Nice, thanks for clarifying. I will update the PR and replace `npartitions` by `partition_size`.
Anything else I should adjust? Is the test construction okay-ish?

crepererum:
> Anything else I should adjust? Is the test construction okay-ish?

Looks good to me 👍

NeroCorleone:
Will close this PR in favor of https://github.com/JDASoftwareGroup/kartothek/pull/98

PR #97, opened 2019-07-24
Use a pip version less than 19.2

kagharpure:
pip 19.2.1 was released yesterday but the issue with `find_links` while using `pip-compile` shows up there too so specifying that `pip` be upgraded to the latest `19.1.*` version.


fjetter:
these things are very annoying :(

kagharpure:
@fjetter, @crepererum - I don't have merge permissions :)

PR #98, opened 2019-07-24
Add keyword for controlling partitions size for dask bag

NeroCorleone:
Adding a keyword `partition_size` to the read functionalities for dask's bag
backend makes it possible to control the size of dask partitions when creating
the bag `db.from_sequence(some_sequence, partition_size)`. For example,`mps
= db.from_sequence(metapartitions, partition_size=100)` would create
dask partitions containing 100 elements of `metapartitions`.

### DIFF
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -59,6 +59,13 @@ def read_dataset_as_metapartitions_bag(
 ):
     """
     Retrieve dataset as `dask.bag` of `MetaPartition` objects.
+
+    Parameters
+    ----------
+
+    Returns
+    -------
+    A dask.bag object containing the metapartions.
     """
     ds_factory = _ensure_factory(
         dataset_uuid=dataset_uuid,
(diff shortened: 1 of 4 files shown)


kartothek/io/dask/bag.py
```diff
@@ -55,6 +55,7 @@ def read_dataset_as_metapartitions_bag(
     predicates=None,
     factory=None,
     dispatch_by=None,
+    partition_size=None,
```
fjetter: Docstring

kartothek/io/dask/bag.py
```diff
@@ -128,6 +129,7 @@ def read_dataset_as_dataframe_bag(
...
```
fjetter: docstring

kartothek/io/dask/bag.py
```diff
@@ -55,6 +55,7 @@ def read_dataset_as_metapartitions_bag(
...
 ):
     """
     Retrieve dataset as `dask.bag` of `MetaPartition` objects.
```
crepererum: that's more a nitpick from the last PR but should be fixed here anyway:
both function actually need the following section in the docstrings to get the parameter list rendered correctly:

```
Parameters
----------
```

(yes, the empty section is enough)

Also, the return type description is missing entirely.

NeroCorleone: Thanks for pointing that out, I will include a "Parameters" and "Return Type" section. 
Anyway, did I get it right that the _actual_ documentation of the parameters should be done with the `default_docs` decorator?

NeroCorleone: (Meaning, I do not have to copy and paste the whole description of parameters here?)

lr4d: Yes, `@default_docs` takes care of the parameter descriptions . You can verify the docstring is how you want it to be by using the ipython `?` magic e.g.`read_dataset_as_metapartitions_bag?`, this way you don't need to build the docs

Issue #54, opened 2019-05-31
Segmentation fault when running pytest on OSX


mganesh1308:
Pytest breaks when we run kartothek on OSX Mojave

I have followed the steps mentioned in the CONTRIBUTING.md.

It breaks with `Segmentation fault: 11` when running `test_eq` module in the `test_metapartition.py`.

Here the pytest logs:

```
(kartothek-dev) ganesh:kartothek 10251XX$ pytest -v -k test_metapartition.py
=================================================================================================== test session starts ====================================================================================================
platform darwin -- Python 3.6.6, pytest-4.5.0, py-1.8.0, pluggy-0.12.0 -- /Users/10251XX/open_kartothek/kartothek-dev/bin/python3.6
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/10251XX/open_kartothek/kartothek/.hypothesis/examples')
rootdir: /Users/10251XX/open_kartothek/kartothek, inifile: setup.cfg
plugins: hypothesis-4.23.9, cov-2.7.1, mock-1.10.4, flake8-1.0.4
collected 1524 items / 1455 deselected / 69 selected                                                                                                                                                                       

tests/io_components/test_metapartition.py::test_store_single_dataframe_as_partition_no_metadata[4] PASSED                                                                                                            [  1%]
tests/io_components/test_metapartition.py::test_store_multiple_dataframes_as_partition[json-4] PASSED                                                                                                                [  2%]
tests/io_components/test_metapartition.py::test_store_single_dataframe_as_partition[json-4-dataset_uuid/core/test_label.parquet] PASSED                                                                              [  4%]
tests/io_components/test_metapartition.py::test_store_multiple_dataframes_as_partition[msgpack-4] PASSED                                                                                                             [  5%]
tests/io_components/test_metapartition.py::test_store_single_dataframe_as_partition[msgpack-4-dataset_uuid/core/test_label.parquet] PASSED                                                                           [  7%]
tests/io_components/test_metapartition.py::test_load_dataframes[4-True] PASSED                                                                                                                                       [  8%]
tests/io_components/test_metapartition.py::test_load_dataframes[4-False] PASSED                                                                                                                                      [ 10%]
tests/io_components/test_metapartition.py::test_remove_dataframes[4] PASSED                                                                                                                                          [ 11%]
[...]

PR #99, opened 2019-07-24
Add osx to testing environments

kagharpure:
Using the osx image's default python3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,7 @@ env:
   - KARTOTHEK_ARROW_VERSION=0.13.0
 
 before_install:
-  # https://github.com/JDASoftwareGroup/kartothek/issues/94
-  - pip install --upgrade pip==19.1.*
-  - pip install pip-tools
-  - ./compile_requirements.sh
-  - python setup.py bdist_wheel
+  - ./ci/before_install.sh
 
 install:
   - pip install -r test-requirements-pinned.txt
@@ -34,6 +30,21 @@ script:
 after_success:
   - pip install codecov
   - codecov
+
+matrix:
+  allow_failures:
+    - os: osx
+  include:
+    - name: "osx + builtin python3"
+      os: osx
+      language: sh
+      env:
+        - KARTOTHEK_ARROW_VERSION=0.13.0
+      before_install:
+        - python3 -m pip install --upgrade virtualenv
+        - virtualenv -p python3 "$HOME/ktk_venv"
+        - source "$HOME/ktk_venv/bin/activate"
+        - ./ci/before_install.sh
 
 deploy:
   provider: pypi
--- a/ci/before_install.sh
+++ b/ci/before_install.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+# https://github.com/JDASoftwareGroup/kartothek/issues/94
+pip install --upgrade pip==19.1.*
+pip install pip-tools
+./ci/compile_requirements.sh
+python setup.py bdist_wheel
--- a/ci/compile_requirements.sh
+++ b/ci/compile_requirements.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -xeo pipefail
 
-productionIndex=platform
-developmentIndex=platform_dev
-
 if [ ! -z ${KARTOTHEK_ARROW_VERSION} ];
 then
     echo pyarrow==$KARTOTHEK_ARROW_VERSION > kartothek_env_reqs.txt


.travis.yml
```diff
@@ -8,8 +8,11 @@ python:
   - 3.7
 
 env:
-  - KARTOTHEK_ARROW_VERSION=0.12.1
```
fjetter: Instead of relying on the ordering here you should define the environemnt for the OSX job explicitly. I believe you can provide an `env` key which will overwrite this global setting

.travis.yml
```diff
@@ -8,8 +8,11 @@ python:
...
   - KARTOTHEK_ARROW_VERSION=0.13.0
+  - KARTOTHEK_ARROW_VERSION=0.12.1
+
+before_cache: rm -f "$HOME/.cache/pip/log/debug.log"
```
fjetter: Is this caching actually required or a nice to have?

.travis.yml
```diff
@@ -35,6 +38,29 @@ after_success:
...
+  include:
+    - name: "osx + builtin python3"
+      os: osx
+      language: sh
+      before_cache:
+              - rm -f "$HOME/Library/Caches/pip/log/debug.log"
+      cache:
+        directories:
+          - "$HOME/Library/Caches/pip"
+      before_install:
```
fjetter: What's different to the other jobs? Why do we need to define this step again. The individual steps look pretty familiar

.travis.yml
```diff
@@ -35,6 +38,29 @@ after_success:
   - pip install codecov
   - codecov
 
+matrix:
+  allow_failures:
+    - os: osx
+  include:
+    - name: "osx + builtin python3"
+      os: osx
+      language: sh
```
fjetter: Language is python. why specify as shell?

.travis.yml
fjetter: The build times look worse, this is why I'm asking. This might, of course, be unrelated and simply due to high load on travis but if it's related to this caching, I'd rather not use it

kagharpure: a) Good to have
b) I tried a bunch of runs last night - I think there was a difference of ~1 minute; therefore a) good to have. Can be removed.

.travis.yml
kagharpure: The cache location is different on osx and simply setting `cache: pip` didn't seem to work as expected for me (I did some searching, I think this is due to Travis not officially supporting python on osx).

fjetter: I was referring to everything else than the cache

.travis.yml
kagharpure: Fair point, changing.

.travis.yml
kagharpure: :D 

tldr: doesn't work with osx. 

Long version: when you specify `language: python` along with `os: osx`, it tries to download the specified python version from a location that doesn't have it and the build errors out. I was trying some complicated mess of an approach earlier on with homebrew (where I was facing some issues with cache-ing) but I realized early yesterday that the osx images on azure pipelines already comes with a version of `python3` (3.6.5), so I am using that directly now.

.travis.yml
fjetter: From what I can see the build takes longer than before.

kagharpure:
> You mentioned issues with homebrew earlier. I don't see you interacting with homebrew

tldr: homebrew issues aren't relevant anymore.

Long version: when you specify `language: python` along with `os: osx`, it tries to download the specified python version from a location that doesn't have it and the build errors out. I was trying some complicated mess of an approach earlier on with homebrew (where I was facing some issues with cache-ing) but I realized early yesterday that the osx images on azure pipelines already comes with a version of python3 (3.6.5), so I am using that directly now.

.travis.yml
kagharpure: Right. So: 

a) `before_install` doesn't stack
b) we need to explicitly setup a virtual environment in case of osx (which happens by default for linux), so the common steps need essentially to be redefined.

.travis.yml
kagharpure: Might be because of load; unsure. The time saving I observed at night was ~1 minute so not that big of a difference. Removing.

.travis.yml
kagharpure: And splitting it out into a common shell script doesn't work as expected. For example: https://travis-ci.com/kagharpure/kartothek/jobs/219631966

kagharpure: Actually, hang on; I think sourcing the script ought to fix this because it's then not running it in a separate shell - just checking now.

fjetter: The linked build fails because the requirements pinning wasn't successful. not because the build script was wrong.

I'm wondering if the setup of the virtualenv can't be used for all builds. I don't see a reason why mac is special here. The same script should run on linux as well

kagharpure: > The linked build fails because the requirements pinning wasn't successful. not because the build script was wrong.

That's because it's using the system `pip`, not the pip updated in the venv created above.

 > 'm wondering if the setup of the virtualenv can't be used for all builds. I don't see a reason why mac is special here. The same script should run on linux as well

If the os is linux, travis creates a venv by default, the user needn't explicitly specify that it be created.

.travis.yml
crepererum: > Long version: when you specify language: python along with os: osx, it tries to download the specified python version from a location that doesn't have it and the build errors out.

That sounds like a travis bug to me. Is there an upstream issue for that? If not, can you please create one? In either case, we need a link to that issue as a comment in the yml file.

kagharpure: It's called out fairly clearly here: https://docs.travis-ci.com/user/languages/python/

kagharpure:
@fjetter, @crepererum - anything left to do before we can consider this merge-ready?

kagharpure:
@fjetter / @crepererum - can either of you merge please? I am not authorized to do so.

lr4d:
@kagharpure Do you know why we are getting the segfault during the OS X tests?

fjetter:
@lr4d see https://github.com/JDASoftwareGroup/kartothek/issues/54

The fix should be trivial

lr4d:
Thanks, @fjetter . I see it has been fixed upstream so we might as well wait for the next pyarrow release

PR #100, opened 2019-07-26
WIP: Nightly numpy pandas pyarrow

mganesh1308:
- Nightly build for Numpy and Pandas
- Nightly build for Pyarrow
- Integration test for Pyarrow in tests/integration
- added ci/ folder for nightly build files for travis
- added before_script.py for travis.yml file


fjetter:
The ci setup looks good on first quick pass.

Can you explain where all of these tests come from? This looks like a strong duplication of existing tests and I'm not entirely sure why we would need this

tests/integration/test_common_metadata_pyarrow.py
```diff
@@ -0,0 +1,223 @@
+# -*- coding: utf-8 -*-
```
crepererum: not sure where these files come from. Are you sure they should be here?

ci/set_numpy_pandas_integration.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+echo "Upgrading to  Nightly build of numpy pandas"
+
+PRE_WHEELS="https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
+
+pip install --pre --no-deps --upgrade --timeout=180 --no-cache-dir -f $PRE_WHEELS numpy pandas
```
crepererum: So this is one way we could do this, but this, but I think this is very unstable for mulitple reasons:

- we don't install dependencies of numpy and pandas (but instead the packages from the pinning)
- we might break version constraints of other packages (like dask) that sometimes constraint the upper bounds for numpy/pandas due to known problems

I think we should rather incooperate the pre-releases into the pinning process (pip-tools can pin pre-releases, but I'm not sure if it can select multiple package sources). Can you please try and figure out if that would work?

.travis.yml
```diff
@@ -11,8 +11,33 @@ env:
...
+matrix:
+  allow_failures:
+    - name: "Integration test for Numpy Pandas"
+    - name: "Integration test for PyArrow"
+  include:
+    - name: "Integration test for Numpy Pandas"
+      if: branch=master
+      python: "3.7"
+      env:
+        - NUMPY_PANDAS_INTEGRATION=1
```
kagharpure: Where is this environment variable being used?

.travis.yml
```diff
@@ -11,8 +11,33 @@ env:
...
+      before_script:
+        - ./ci/before_script.sh
+        - ./ci/set_numpy_pandas_integration.sh
+    - name: "Integration test for PyArrow"
+      if: branch=master
+      python: "3.7"
+      env:
+        - PYARROW_INTEGRATION=1
```
kagharpure: Where is this environment variable being used?

mganesh1308:
> The ci setup looks good on first quick pass.
> 
> Can you explain where all of these tests come from? This looks like a strong duplication of existing tests and I'm not entirely sure why we would need this

@fjetter Yes, These are the tests we have pulled from units tests of kartothek which was failed with the errors for pyarrow latest builds and we are adding here for the integration tests.

.travis.yml
mganesh1308: @kagharpure, Initially we have used that variable for enabling some conditions but now we can remove that.

.travis.yml
mganesh1308: @kagharpure, Initially we have used that variable for enabling some conditions but now we can remove that.

ci/set_numpy_pandas_integration.sh
mganesh1308: ya, will check on that @crepererum

tests/integration/test_common_metadata_pyarrow.py
mganesh1308: @crepererum, These are the tests we have pulled from units tests of kartothek which was failed with the errors for pyarrow latest builds and we are adding here for the integration tests, we need to add/remove anything here?

fjetter:
@mganesh1308 Why do you distinguish between unit and integration tests? why not let everything run?

fjetter:
Closed in favour of #103

PR #101, opened 2019-07-29
Threading with back pressure handling

fjetter:
I wanted to introduce this a long time ago and dug out the code again. Back then there was the concern that we do not have any guarantee about the thread safety of the stores which is why I introduced this `StorePool` class which would allow us to use distinct store object but don't need to recreate the connections and objects for every thread.
(please, don't be picky about the implementation details, I haven't spend much time on this)

Nice thing about this approach is that we benefit in all eager and iter read implementations.

Refinement options:
* Global theadpool
* Global store pool (possibly even throughout ktk but private / non public API)

Please give me your thoughts. If this is something we feel comfortable with, I'll clean up the code

### DIFF
--- a/kartothek/io/iter.py
+++ b/kartothek/io/iter.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 from functools import partial
-from typing import cast
 
 from kartothek.core.factory import _ensure_factory
 from kartothek.core.naming import (
@@ -26,6 +25,8 @@ from kartothek.io_components.write import (
     store_dataset_from_partitions,
 )
 
+from ._storepool import MapStorepool, MapStorepoolThreaded
+
 
 @default_docs
 def read_dataset_as_metapartitions__iterator(
@@ -42,6 +43,7 @@ def read_dataset_as_metapartitions__iterator(
     predicates=None,
     factory=None,
     dispatch_by=None,
+    nthreads=0,
 ):
     """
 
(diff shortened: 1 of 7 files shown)


kartothek/io/iter.py
```diff
@@ -27,6 +30,36 @@
...
 
+class StorePool:
+    def __init__(self, store_factory):
+        """Generally we cannot assume that store objects are thread safe. Therefore, pool the objects to limit the number of required connections"""
+        # deques are faster than ordinary queues since they don't require locking
+        self._pool = deque()
+        self.store_factory = store_factory
+
+    def store(self):
+        try:
```
xhochy: Use `len(self._pool)` instead of  exceptions

fjetter: Wouldn't guarantee that the pop will be successful, would it?

fjetter:
> but please get rid of all the print statements.

Obviously, I'll get rid of it. That was just for debugging.

kartothek/io/iter.py
xhochy: You would need a mutex around this section obviously.

crepererum:
Just to understand the scope: this is only for eager or do you plan to do that for other backends as well? (like nested parallelism for dask)

fjetter:
@marco.neumann this is only for eager/iter but we can investigate this for other schedulers as well.
I see some potential for parallelization with nested metapartitions but I'd consider this oos for now

kartothek/io/iter.py
fjetter: I'll look into the solution with mutex but my first try included a queue and the sync/lock overhead was quite significant

kartothek/io/iter.py
```diff
@@ -68,33 +103,62 @@ def read_dataset_as_metapartitions__iterator(
...
+        with ThreadPoolExecutor(nthreads) as pool:
+            pool.map(fetch_partition, mps)
+            # It is very important that this loop is within the threadpool ctx mngr. Otherwise all computations are calculated before anything is yielded to be able to shut down the executor
+            print("enter loop", flush=True)
+            # must not iterate over the futures object since this may block
+            for _ in range(len(mps)):
+                yield result_queue.get()
+                print("yielded", flush=True)
+            print("done", flush=True)
+    else:
```
amerkel2: What would be the overhead/dowside of using a single code path, i.e., spawning one thread in case no parallelism is wanted?

fjetter: this implementation implies that there are up to (1 + nthreads) partitions in memory. If the iter interface is used in situations where the memory footprint is supposed to be low this might be undesired.
I'll add this to the docs, of course

kartothek/io/iter.py
```diff
@@ -68,33 +70,37 @@ def read_dataset_as_metapartitions__iterator(
...
-        yield mp
+
+            def func_concat(func, mps, store, *args, **kwargs):
+                # must not use theaded map from outside since all threads share
+                # one result queue and share one concurrency limit. This might
+                # lead to either a deadlock, inconsistent results or both
+                res = map(partial(func, store=store), mps)
+                return MetaPartition.concat_metapartitions(list(res))
+
+            func = partial(func_concat, func)
```
lr4d: Would appreciate a bit more clarity of function names

lr4d:
@fjetter status?

fjetter:
Closing for now until a proper use case arises

PR #102, opened 2019-07-29
Downpin simplekv so docs don't break

kagharpure:
New version of `simplekv` dropped on the 26th which doesn't seem to have `pickle_azure_store` anymore, so downpinning it to `<0.13`


crepererum:
Let's see if we can get blue-yonder/storefact#14 merged+released quickly and put this PR here on hold.

kagharpure:
blue-yonder/storefact#14 is merged so this isn't needed anymore, closing.

PR #104, opened 2019-07-30
fix rejection of bool predicates when bool cols has None

crepererum:


tests/serialization/test_filter.py
```diff
@@ -92,20 +96,76 @@ def test_raise_on_type(value, filter_value, op):
...
+    df = pd.DataFrame({"A": data})
+    df["B"] = range(len(df))
+
+    predicates = [[("A", op, value)]]
+    actual = filter_df_from_predicates(df, predicates)
+    if pd.api.types.is_categorical(df["A"]):
+        df["A"] = df["A"].astype(df["A"].cat.as_ordered().dtype)
+    if isinstance(value, datetime.date) and (df["A"].dtype == "datetime64[ns]"):
+        # silence pandas warning
+        value = pd.Timestamp(value)
```
fjetter: If it's about silencing a warning, use filter warnings, e.g. https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings

This casting otherwise defeats the purpose of the test, doesn't it?

crepererum:
> Can you please be more specific about the intended behaviour? afaik, in pandas there is no such thing as a nullable bool, is there?

In ktk there is. We support bool as a payload type, and pandas can only represent this either with `bool` or (in case of any missing values) with `object`. Pyarrow supports this as well. The problem  now is that we're too strict in the "contains missing values" case and reject all predicates.

tests/serialization/test_filter.py
crepererum: pandas will reject data<->datetime comparisons entirely in the future. We can either silence the warning and break soon or can decide to always be strict and stop all this magic altogether or support it and therefore need a slightly more complicates test.

kartothek/serialization/_generic.py
```diff
@@ -268,23 +268,26 @@ def _ensure_type_stability(array_like, value, strict_date_types, require_ordered
...
-        {"u", "i"},
+        ("u", "i"),
+        ("i", "u"),
         # various string kinds
-        {"S", "O"},
-        {"U", "O"},
+        ("S", "O"),
+        ("U", "O"),
+        # bool w/ Nones
+        ("b", "O"),
```
lr4d: I suppose you are using tuples here instead of sets because the order matters ? i.e. `bool, obj` is valid but `obj, bool` isn't?
Is this also the case for `S, O` and `U, O` ?

fjetter:
Ok, I misunderstood the intention here. Now I also understand the test cases. 👍

kartothek/serialization/_generic.py
crepererum: It sadly does. For example, if you would use a tuple, you could query a string (object) column with a bool. So you want to only be fuzzy in one direction.

crepererum: > Is this also the case for S, O and U, O ?
I don't think so. @fjetter ?

crepererum: Ah, the tests found some. I'll have a look.

fjetter: Frankly, I don't care much about `S` types since they are already deprecated by NumPy. They are, however, tested by hypothesis. Whatever makes the tests green :)

I was wondering about the order as well. I was hoping it was agnostic to the order but the case of bool/none alone make a case for a relevant ordering, doesn't it?

valid (O, b)
```
array: [True, False, None] 
value: False
```

invalid (b, O)
```
[True, True, False] 
"ARandomString"
```

lr4d:
```

Comment (crepererum) [reply kartothek/serialization/_generic.py]: > I was wondering about the order as well. I was hoping it was agnostic to the order but the case of bool/none alone make a case for a relevant ordering, doesn't it?

Correct, that's exactly the reason for making this part ordered.

Comment (lr4d) [reply kartothek/serialization/_generic.py]: How about these cases?

O, b
```
array: ["ARandomString"]
value: True
```

b, O
```
array: [True, False]
value: None
```

tests/serialization/test_filter.py
fjetter: Yes, it is supposed to raise. I would've expected that *we* take care of the appropriate casting already or is it about the expected value?

fjetter: ok, my bad... I should start reading the code more carefully... sorry

kartothek/serialization/_generic.py
crepererum: > O, b

covered by the tests and forbidden (see `test_raise_on_type` with `(True, "True")`).

crepererum: I'll fix the other one.

PR #107, opened 2019-07-31
Fix string<-> bool regression from #104

crepererum:
This was found by @lr4d in the review of #104. I would like to fix the edge case ("bool predicates cannot be used to query string columns!"). Notice that this somewhat rolls back #104 and uses an alternative solution.

I think this doesn't require a changelog entry since #104 was never released and I am just fixing the regression introduced by it.

### DIFF
--- a/kartothek/serialization/_generic.py
+++ b/kartothek/serialization/_generic.py
@@ -219,13 +219,27 @@ def _handle_categorical_data(array_like, require_ordered):
     return array_like, array_value_type
 
 
-def _handle_null_arrays(array_like, value_dtype):
+def _handle_null_arrays(array_like, array_value_type, value_dtype):
     # NULL types might not be preserved well, so try to cast floats (pandas default type) to the value type
     # Determine the type using the `kind` interface since this is common for a numpy array, pandas series and pandas extension arrays
-    if array_like.dtype.kind == "f" and np.isnan(array_like).all():
-        if array_like.dtype.kind != value_dtype.kind:
+    if array_value_type.kind == "f" and np.isnan(array_like).all():
+        if array_value_type.kind != value_dtype.kind:
             array_like = array_like.astype(value_dtype)
-    return array_like, array_like.dtype
+            array_value_type = value_dtype
+    return array_like, array_value_type
+
+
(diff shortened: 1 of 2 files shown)


lr4d:
So the main difference here is that you are checking that `value=bool, array_type=str` does not happen, right?
I just commented the first example that came to mind in #104, can we increase the testing someway to ensure no other edge cases slip through?

crepererum:
> So the main difference here is that you are checking that value=bool, array_type=str does not happen, right?

yes.

> I just commented the first example that came to mind in #104, can we increase the testing someway to ensure no other edge cases slip through?

I've already tried. If you can come up with a more aggressive testing scenario, please tell me.

kartothek/serialization/_generic.py
```diff
@@ -266,28 +280,28 @@ def _ensure_type_stability(array_like, value, strict_date_types, require_ordered
...
-        ("u", "i"),
-        ("i", "u"),
+        {"u", "i"},
         # various string kinds
-        ("O", "S"),
-        ("O", "U"),
-        # bool w/ Nones
-        ("b", "O"),
+        {"S", "O"},
+        {"U", "O"},
```
lr4d: Would this allow the comparison of an array of non-string objects with a string?

crepererum:
So I have tried out multiple things and having type safety even in NumPy doesn't work, if you don't want to scan types in objects columns manually. The issue is that for NumPy, there is no difference between converting between things like

- `O` containing `str` to `U`
- `O` containing `date` to `U`

So I would say if we don't completely re-write that thing to use arrow, the small loopholes in the comparison are not really fixable without affecting performance too much.

fjetter:
1. I'd be fine with rewriting to use arrow type system. I used the NumPy type system when I wrote these checks since the comparison is done by NumPy. Going to arrow, though, must not transform the entire array to an arrow array since this would kill performance.
2. If we don't support 100% proper type recognition for object type columns, that's a hit I'm willing to make. If it causes actually serious issues we might revisit this.

lr4d:
@crepererum What's the status on this? If we are choosing to leave the type recognition as is for now, I guess we can close this

lr4d:
I've opened #193 to track this issue. Closing this for now

PR #103, opened 2019-07-30
WIP: Nightly builds pyarrow pandas

mganesh1308:

PR #114, opened 2019-08-08
Nightly pandas, numpy and arrow integration

### LINKED ISSUE #19: Test against Pandas and NumPy nightlies
See https://github.com/xhochy/fletcher/blob/0e51c14fb36ab2e6a6d3086e0e07ff2e5cb2aead/ci/circle_build_linux.sh#L39-L44 on how to install them. This should give us advance warning when they break.

This should be only done on master, especially together with https://github.com/JDASoftwareGroup/kartothek/issues/18

crepererum:
Closes #19
Superseds #103

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
 env:
   - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
   - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
+  - PIP_COMPILE_ARGS="-P pyarrow==0.13.0 --pre -f https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
 
 before_install:
   - ./ci/before_install.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,14 @@ python:
   - 3.7
 
 env:
-  - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
-  - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
-  - PIP_COMPILE_ARGS="-P pyarrow==0.13.0 --pre -f https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
+  global:
+    - GITHUB_CLIENT_ID=32eb43d9e96045689676
+    - secure: "MvdUzU2BrcgAKWjeeF0X2ajZ7fNpn5373edG2/bUXAsBImCUhSHxS33Pzp9pwSjxAxkhFW+Ic5FUg9JpxS6C2m+BxgQry43YHizNZY6yBaM1i7Tm+LXZJv68ppUdkSVRhnm0aAxz+TLa+lYCKyc2H5DmmF/R8d1FiPXWtA/9xQItV4ME9G+6s6VxuZtpUpndUYdZWhBzFQUIQh9vn9ymoDww7NRKYEnYQbBtyZoIS9aZC1zAcYfuUjxri0LL/UpDm+MszAH7VTJwGdjOgFHmAsJzHw7irYLBOxtEc3PvRB2EvLh0QxJpcWXw/KKie9xwtJBqWc+l+aXaWUYKQ5dyrZ6c/Q9RkBV0aK3680ri2WY6RGTxQVPIQTq5XvacsJ5b/S9UwdiIrZFjSI/kWtgXLFUoqiQiBfCvWBw5/lMY9Qv7SeNU2MzS6Sbz8Igki6gNQl9AaGuCcIBl/UGus3wmZDm8WppKgh0EMEA1sP1CsuBUNm9+VK1UUkgNhcDP7Ac6F6hIPHcH8mCsjU8CAyaGVyQiJGoghSCjgDAjIhT3q5YKzFdBtgMZGvP6k0Wl1iE7Pr54qyrEiOYY9bDz7WN9Kl41lUJU5A5F4Y3BkFrWDXCEy7se/HgqpWft+taMVZxrXaZ1N8pv0ahEfbBVZL04rnw4+hWO9HeFBqFKlu23czE="
+  matrix:
+    - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
+    - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
+    - PIP_COMPILE_ARGS="-P pyarrow==0.13.0 --pre -f https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
+    - ARROW_NIGHTLY=1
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -8,8 +8,14 @@ python:
...
 env:
-  - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
-  - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
+  global:
+    - GITHUB_CLIENT_ID=32eb43d9e96045689676
+    - secure: "MvdUzU2BrcgAKWjeeF0X2ajZ7fNpn5373edG2/bUXAsBImCUhSHxS33Pzp9pwSjxAxkhFW+Ic5FUg9JpxS6C2m+BxgQry43YHizNZY6yBaM1i7Tm+LXZJv68ppUdkSVRhnm0aAxz+TLa+lYCKyc2H5DmmF/R8d1FiPXWtA/9xQItV4ME9G+6s6VxuZtpUpndUYdZWhBzFQUIQh9vn9ymoDww7NRKYEnYQbBtyZoIS9aZC1zAcYfuUjxri0LL/UpDm+MszAH7VTJwGdjOgFHmAsJzHw7irYLBOxtEc3PvRB2EvLh0QxJpcWXw/KKie9xwtJBqWc+l+aXaWUYKQ5dyrZ6c/Q9RkBV0aK3680ri2WY6RGTxQVPIQTq5XvacsJ5b/S9UwdiIrZFjSI/kWtgXLFUoqiQiBfCvWBw5/lMY9Qv7SeNU2MzS6Sbz8Igki6gNQl9AaGuCcIBl/UGus3wmZDm8WppKgh0EMEA1sP1CsuBUNm9+VK1UUkgNhcDP7Ac6F6hIPHcH8mCsjU8CAyaGVyQiJGoghSCjgDAjIhT3q5YKzFdBtgMZGvP6k0Wl1iE7Pr54qyrEiOYY9bDz7WN9Kl41lUJU5A5F4Y3BkFrWDXCEy7se/HgqpWft+taMVZxrXaZ1N8pv0ahEfbBVZL04rnw4+hWO9HeFBqFKlu23czE="
+  matrix:
+    - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
+    - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
+    - PIP_COMPILE_ARGS="-P pyarrow==0.13.0 --pre -f https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
```
lr4d: I get that this means pandas and numpy nightlies but I don't know if it would be that obvious to a new committer. Is it worth it to clarify?

ci/compile_requirements.sh
```diff
@@ -1,13 +1,32 @@
...
+set -xueo pipefail
+
+if [ -z ${PIP_COMPILE_ARGS+x} ]; then
+    PIP_COMPILE_ARGS=""
+fi
+
+if [[ "${ARROW_NIGHTLY+1}" == 1 ]]; then
+    ARROW_URL="$(./ci/get_pyarrow_nightly.py)"
+    ARROW_FILE="$(basename $ARROW_URL)"
+    ARROW_VERSION="$(echo "$ARROW_FILE" | awk '{split($1,a,"-"); print a[2]}')"
```
lr4d: you are nesting double quotes here

crepererum: I am nesting `"..."`, `$(...)`, `"..."`. That's valid in bash.

.travis.yml
fjetter: Maybe a comment about where this URL is from?

ci/compile_requirements.sh
fjetter: We were talking about an arrow issue connected to packaging of nighties, weren't we? Can you attach a link to that ticket. I don't mind the messiness for now but I'd like to keep track no this. This doesn't feel like a stable solution, especially not in comparison to pandas/numpy

.travis.yml
```diff
@@ -8,8 +8,14 @@ python:
   - 3.7
 
 env:
-  - PIP_COMPILE_ARGS="-P pyarrow==0.12.1 -P dask<2.2.0"
-  - PIP_COMPILE_ARGS="-P pyarrow==0.13.0"
+  global:
+    - GITHUB_CLIENT_ID=32eb43d9e96045689676
+    - secure: "MvdUzU2BrcgAKWjeeF0X2ajZ7fNpn5373edG2/bUXAsBImCUhSHxS33Pzp9pwSjxAxkhFW+Ic5FUg9JpxS6C2m+BxgQry43YHizNZY6yBaM1i7Tm+LXZJv68ppUdkSVRhnm0aAxz+TLa+lYCKyc2H5DmmF/R8d1FiPXWtA/9xQItV4ME9G+6s6VxuZtpUpndUYdZWhBzFQUIQh9vn9ymoDww7NRKYEnYQbBtyZoIS9aZC1zAcYfuUjxri0LL/UpDm+MszAH7VTJwGdjOgFHmAsJzHw7irYLBOxtEc3PvRB2EvLh0QxJpcWXw/KKie9xwtJBqWc+l+aXaWUYKQ5dyrZ6c/Q9RkBV0aK3680ri2WY6RGTxQVPIQTq5XvacsJ5b/S9UwdiIrZFjSI/kWtgXLFUoqiQiBfCvWBw5/lMY9Qv7SeNU2MzS6Sbz8Igki6gNQl9AaGuCcIBl/UGus3wmZDm8WppKgh0EMEA1sP1CsuBUNm9+VK1UUkgNhcDP7Ac6F6hIPHcH8mCsjU8CAyaGVyQiJGoghSCjgDAjIhT3q5YKzFdBtgMZGvP6k0Wl1iE7Pr54qyrEiOYY9bDz7WN9Kl41lUJU5A5F4Y3BkFrWDXCEy7se/HgqpWft+taMVZxrXaZ1N8pv0ahEfbBVZL04rnw4+hWO9HeFBqFKlu23czE="
```
fjetter: This is your personal token?

ci/compile_requirements.sh
lr4d: my bad

.travis.yml
crepererum: yes

crepererum:
> Do we want to have arrow + pandas + Numpy nightly or is this overkill?

I would like to avoid that to not make the nightly overly unstable. It's just too many nightly / unreleased artifacts together.

> I also think that it suffices to build the nightlies on master only. what do you think?

I would rather flag these with "can fail" and run them for PRs as well, because otherwise it gets quite hard to fix the failing master using a PR (sure, you can re-enable the "for PRs as well" and the disable this, but that's really but DX).

fjetter:
> I would rather flag these with "can fail" and run them for PRs as well, because otherwise it gets quite hard to fix the failing master using a PR (sure, you can re-enable the "for PRs as well" and the disable this, but that's really but DX).

Makes sense. Let's see how it behaves. We can change the behaviour again if it makes issue (like consuming too much time until we receive feedback)

lr4d:
> I would rather flag these with "can fail" and run them for PRs as well, because otherwise it gets quite hard to fix the failing master using a PR

Good point

PR #130, opened 2019-09-13
Improve predicate performance for `in` operator

lr4d:
Benchmark code (on local filesystem):
```
from functools import partial

import numpy as np
import pandas as pd
import pyarrow.parquet as pq
import storefact

from kartothek.io.eager import store_dataframes_as_dataset
from kartothek.serialization import ParquetSerializer

tmpdir = TemporaryDirectory().name
store_factory = partial(storefact.get_store_from_url, f"hfs://{tmpdir}")

df = pd.DataFrame({"x": np.tile([17, 18], 10 ** 6)})

uuid = "test"
dm = store_dataframes_as_dataset(store=store_factory, dataset_uuid=uuid, dfs=df)

storage_key = dm.partitions[list(dm.partitions.keys())[0]].files["table"]
filepath = f"{tmpdir}/{storage_key}"

parquet_file = pq.ParquetFile(filepath)

ps = ParquetSerializer()

%timeit ps.restore_dataframe(store_factory(), storage_key, predicates=[[("x", "in", (13, 15, 20))]])
```

Original: `44.9 ms ± 1.61 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`
New: `4.18 ms ± 303 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`
- [x] Tests added / passed
- [x] Changelog entry

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -263,13 +263,9 @@ def _normalize_predicates(parquet_file, predicates, for_pushdown):
             if op == "in":
                 values = [_normalize_value(l, pa_type) for l in literal[2]]
                 if for_pushdown and values:
-                    # In the case of predicate pushdown, we only compare ranges
-                    # and their overlap. Therefore we only need the min and max
-                    # values of the values.
-                    normalized_value = (
-                        _timelike_to_arrow_encoding(min(values), pa_type),
-                        _timelike_to_arrow_encoding(max(values), pa_type),
-                    )
+                    normalized_value = [
+                        _timelike_to_arrow_encoding(value, pa_type) for value in values
+                    ]
                 else:
                     normalized_value = values
             else:
@@ -405,7 +401,14 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
     elif op == ">":
         return max_value > val
     elif op == "in":
-        return bool(val) and not ((max_value < val[0]) or (min_value > val[1]))
+        # This implementation is chosen for performance reasons. See
+        # https://github.com/JDASoftwareGroup/kartothek/pull/130 for more information/benchmarks.
+        # We accept the predicate if there is any value in the provided array which is equal to or between
+        # the parquet min and max statistics. Otherwise, it is rejected.
+        for x in val:
+            if min_value <= x <= max_value:
+                return True
+        return False
     else:
         raise NotImplementedError("op not supported")
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Version 3.X.X (2019-09-XX)
 ==========================
 - ``MetaPartition.load_dataframes`` now raises if table in ``columns`` argument doesn't exist
 - require ``urlquote>=1.1.0`` (where ``urlquote.quoting`` was introduced)
+- Improve performance for some cases where predicates are used with the `in` operator.
 
 
 Version 3.4.0 (2019-09-17)
--- a/tests/serialization/test_filter.py
+++ b/tests/serialization/test_filter.py
@@ -90,14 +90,14 @@ def test_filter_array_like_categoricals(op, expected, cat_type):
         (True, None),
     ],
 )
-@pytest.mark.parametrize("op", ["==", "!=", "<", "<=", ">", ">="])
+@pytest.mark.parametrize("op", ["==", "!=", "<", "<=", ">", ">=", "in"])
 def test_raise_on_type(value, filter_value, op):
     array_like = pd.Series([value])
     with pytest.raises(TypeError, match="Unexpected type encountered."):
         filter_array_like(array_like, op, filter_value, strict_date_types=True)
 
 
-@pytest.mark.parametrize("op", ["==", "!=", ">=", "<=", ">", "<"])
+@pytest.mark.parametrize("op", ["==", "!=", ">=", "<=", ">", "<", "in"])
 @pytest.mark.parametrize(
     "data,value",
     [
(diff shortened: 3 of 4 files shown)


lr4d:
This improves the performance when the Parquet files' statistics
are between the values provided to "in" but none of the values provided
to "in" are between the Parquet statistics.

Current implementation is as follows:
e.g.
`predicates = [[("x", "in", (13, 15, 20))]]`
Given a Parquet file `A.parquet` with statistics: x.min = 17, x.max=18

The predicate is parsed as x >= 13, x <= 20 in predicate logic. This would
yield true since according to Parquet the range of x goes from 17 to 18.

`A.parquet` would be read entirely, but that is unnecessary given the predicates.

This change parses the provided predicate as:
`any([x.min <= v <= x.max for v in (13, 15, 20)])`

Thus, `A.parquet` would not be read entirely, as there is no chance for values of `A` to be in the range specified. This results in a significant performance improvement for this scenario

crepererum:
I think this is a tradeoff. If the number of variables in the predicate is rather low and especially wide-spread, so the set of matched row-groups is low, then your change would be especially effective. If the array in the predicate is very long, then your change might even be counter-productive due to all the python-driven iteration over all the predicate values. So: can we have another test case where the predicate is large?

We could probably also implement both variants (per-value filter and min-max-based filter) and select the algorithm based on the size of the predicate (e.g. if the "in"-clause contains more than 100 elements, just use min and max to filter row-groups).

lr4d:
>If the array in the predicate is very long, then your change might even be counter-productive due to all the python-driven iteration over all the predicate values.

Yes, this occurred to me. I will do some more benchmarks when I have the time to check performance of larger predicates.

kartothek/serialization/_parquet.py
```diff
@@ -405,7 +401,12 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
     elif op == ">":
         return max_value > val
     elif op == "in":
-        return bool(val) and not ((max_value < val[0]) or (min_value > val[1]))
+        # We accept the predicate if there is any value in the provided array which is equal to or between
+        # the parquet min and max statistics. Otherwise, it is rejected.
+        for x in val:
+            if min_value <= x <= max_value:
```
fjetter: Maybe try to convert x to a `np.array` and calculate `any( (x_arr >= min_value) & (x_arr <= max_value) )` instead of doing the for loop. This might soften the blow for large `in` predicates.

lr4d:
I ran some benchmarks again with predicates in ranges of 2M, 20K and 3. I compared master, the `for` loop implementation and the implementation suggested by @fjetter using numpy

Predicates used:
```python
matching_long_predicate_value = tuple(range(-10**6, 10**6))
long_predicate_value = (*range(-10**6, 15), *range(20, 10**6))
mid_predicate_value = (*range(-10**4, 15), *range(20, 10**4))
short_predicate_value = (13, 15, 20)
matching_short_predicate_value = (13, 18, 20)
```

Results:
```
# Using for loop

# Long predicate which matches: (length=2M)
# 7.01 s ± 1.92 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Long predicate (lenght=1999995):
# 6.62 s ± 672 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Mid-sized predicate (length=19995)
# 61 ms ± 1.16 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Short predicate:
# 3.96 ms ± 375 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# Matching short predicate:
# 75.8 ms ± 11.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# ------------
# Using `np.any`
# Long predicate (lenght=1999995):
#6.69 s ± 360 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Mid-sized predicate (length=19995)
# 73.3 ms ± 5.22 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Short predicate:
# 4.97 ms ± 793 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)


# ---------------
# Current implementation (master branch):

# Long predicate which matches: (length=2M)
# 6.24 s ± 451 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Long predicate (lenght=1999995):
# 5.21 s ± 330 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Mid-sized predicate (length=19995)
# 213 ms ± 15.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Short predicate:
# 57.9 ms ± 6.64 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# Matching short predicate:
# 63.2 ms ± 3.57 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

Obviously some margin of error needs to be taken into account with these results, esp. for large predicates.

### Conclusion
If the values provided as `in` predicates are _not_ in the actual Parquet file and the current predicate logic returns True, this will be faster than the current predicate logic, the difference being relatively big for `len(values) <= 50K`, however, at `len(values)> 1M` this will actually be slower.
On the other hand, if the current predicate logic returns False, this will probably be slower in all cases (however, this difference is relatively small). 
If the values provided as  `in` predicates are in between the Parquet statistics, this will also be slower than the current predicate logic, although again the difference is relatively small.
 
Note that, in any case, when predicate values are larger than ~1M, this implementation will be slower than the current one.
Thus, I think @crepererum suggestion about using this for small predicates, and using the current implementation for large predicates makes sense.

fjetter:
The results are quite striking and I cannot see any obvious errors in the benchmark setup which could account for large deviations (Sure, it's only one data type, data setup is constant, etc. but that shouldn't have a significant impact).

Therefore, I'd also be fine with the for-loop-only implementation with a dedicated comment about the performance and a link to this issue. The measurements above indicate that even for a list of 20k items the for loop performs better. A list of 20k items doesn't count as a _mid-sized_ list anymore but is so large that I would start questioning the approach of using predicate pushdown by itself (We have lists up to ~200 items in production).

Side note: sorting the list and performing some sort of binary search might improve this even further but at some point I would argue that the performance is good enough 😁

lr4d:
Updated changelog

PR #132, opened 2019-09-23
Add explicit readthedocs configuration

fjetter:
By default, readthedocs will use setuptools install to set up the doc build.
This may lead to build errors for non-python dependencies since we get some
coupling to their packaging implementations (e.g. without this it was tried
to compile a rust package). This way we can force it to explicitly use pre
packaged wheels.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,14 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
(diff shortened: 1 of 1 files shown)


lr4d:
If I go to the module reference pages (e.g. https://kartothek.readthedocs.io/en/latest/_rst/kartothek.io.eager.html), I am unable to see the source of the functions when hitting `[source]`. Do you know what causes this?

fjetter:
Source is probably not working because I changed the default branch back to master.

lr4d:
It would be nice if it did

PR #126, opened 2019-08-25
`store_dataframes_as_dataset` accepts dicts and list of tuples equivalents

SvoONs:
# Description:

This PR makes `io/eager/store_dataframes_as_dataset` accept `dfs` arguments in the following formats:
`{"data": {"table": df}}, {"table": df}, {"data": [("table", df)]}` and `[("table", df)]`.

Docstrings have not been adapted yet.

- [x] Closes #63 
- [x] Tests added / passed
- [x] Passes `./format_code.sh`
- [x] Changelog entry

PR #135, opened 2019-09-23
Added secondary_indices as a default arg

### LINKED ISSUE #46: Keyword argument for secondary_indices is not used consistently in the code base
We introduced the keyword argument `secondary_indices` to allow writing pipelines to calculate an index on the fly instead of having the user to provide this as an input.
This keyword has been introduced only inconsistently, yet and many pipelines still don't use it. We should enable all writing pipelines to use this keyword argument to eventually deprecate the explicit user input.

aniruddhgoteti:
# Description:

Added a consistant arg `secondary_indices' to all writing pipelines

- [x] Closes #46 
- [x] Tests added / passed
- [x] Passes `./format_code.sh`
- [x] Changelog entry

### DIFF
--- a/kartothek/io/eager.py
+++ b/kartothek/io/eager.py
@@ -18,7 +18,7 @@ from kartothek.core.naming import (
     PARQUET_FILE_SUFFIX,
     get_partition_file_prefix,
 )
-from kartothek.core.uuid import gen_uuid
+from kartothek.io.iter import store_dataframes_as_dataset__iter
 from kartothek.io_components.delete import (
     delete_common_metadata,
     delete_indices,
@@ -42,10 +42,7 @@ from kartothek.io_components.utils import (
     sort_values_categorical,
     validate_partition_keys,
 )
-from kartothek.io_components.write import (
-    raise_if_dataset_exists,
-    store_dataset_from_partitions,
-)
+from kartothek.io_components.write import raise_if_dataset_exists
 
 
 @default_docs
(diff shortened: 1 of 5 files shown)


kartothek/io/dask/bag.py
```diff
@@ -56,6 +56,7 @@ def read_dataset_as_metapartitions_bag(
     factory=None,
     dispatch_by=None,
     partition_size=None,
+    secondary_indices=None,
```
fjetter: This is a read function. No need for the secondary indices keyword

kartothek/io/dask/bag.py
```diff
@@ -239,7 +246,7 @@ def store_bag_as_dataset(
 
 @default_docs
 def build_dataset_indices__bag(
-    store, dataset_uuid, columns, partition_size=None, factory=None
+    store, dataset_uuid, columns, partition_size=None, factory=None, secondary_indices=None,
```
fjetter: This pipeline is intended for building indices and the argument `columns` covers already that use case

kartothek/io/dask/dataframe.py
```diff
@@ -36,6 +36,7 @@ def read_dataset_as_ddf(
     dates_as_object=False,
     predicates=None,
     factory=None,
+    secondary_indices=None,
```
fjetter: Also a read pipeline

kartothek/io/dask/delayed.py
```diff
@@ -156,6 +156,8 @@ def merge_datasets_as_delayed(
     match_how="exact",
     label_merger=None,
     metadata_merger=None,
+    secondary_indices=None,
```
fjetter: This may actually be valid :)

kartothek/io/dask/delayed.py
```diff
@@ -273,6 +278,8 @@ def read_dataset_as_delayed_metapartitions(
     predicates=None,
     factory=None,
     dispatch_by=None,
+    secondary_indices=None,
```
fjetter: again read

aniruddhgoteti:
Also do I need to write entry in the changelog for this?

aniruddhgoteti:
> Why do you require the secondary indices keyword for read pipelines?

Misread only `write` pipelines as `all` pipelines

fjetter:
Can you describe the use case for having index build in read pipelines?

fjetter:
Yes, please also add a changelog entry

aniruddhgoteti:
> Please rebase with the latest master and update the changelog accordingly

Done

kartothek/io/eager.py
```diff
@@ -497,6 +502,9 @@ def store_dataframes_as_dataset(
         store=store, dataset_uuid=dataset_uuid, df_serializer=df_serializer
     )
 
+    if secondary_indices:
```
lr4d: you should build indices before you `store_dataframes`, as you require the data to be present at the metapartition (it is gone once you have stored it)

aniruddhgoteti: oh ya! Thanks

lr4d:
Looks good but we should add some tests. I will try to get this done for `store_dataset_as_dataframes` which is currently not using the generic tests because of its broken interface. See if you can do this for the other functions

lr4d:
Will have to wait until #126 is merged to get `store_dataset_as_dataframes` conforming to generic tests

kartothek/io/dask/delayed.py
```diff
@@ -156,6 +156,7 @@ def merge_datasets_as_delayed(
     match_how="exact",
     label_merger=None,
     metadata_merger=None,
+    secondary_indices=None,
```
fjetter: Actually, I thought about this pipeline again and I believe we don *not* want to allow index building here. Index building should only be supported for pipelines with immediate write which is not the case here. Sorry for the confusion. Can you please remove this again?

aniruddhgoteti: Sure

PR #137, opened 2019-09-24
Add minimal version requirement for urlquote

fjetter:
Only version 1.1 offers the urlquote.quoting module which is used in Kartothek


fjetter:
duplicate of #136

PR #138, opened 2019-09-25
Normalise properly to lists within normalise_arg

### LINKED ISSUE #56: Normalize input to `list` properly in `normalize_args`
Currently, if a user wants to partition along column "x", and since columns are unique, thinks it is appropriate to provide a set as the value to the argument e.g. `partition_on={"x"}`; this would be "normalized" to `[{"x"}]` ( a list of sets), and then an error will be raised at some other part of the code where this input is unexpected.
The expected behavior of the normalization would be to normalize sets, tuples or similar iterables to a list such as `["x"]` (a list of strings).

I do not consider this a bug as in the docs it states `partition_on` should be provided as a list.

Normalization function: `kartothek.io_components.utils.normalize_arg`

aniruddhgoteti:
# Description:

Normalise properly to lists within normalise_arg

- [x] Closes #56 
- [x] Changelog entry

### DIFF
--- a/kartothek/io_components/utils.py
+++ b/kartothek/io_components/utils.py
@@ -167,7 +167,15 @@ _ARGS_TO_TYPE = {"partition_on": list, "delete_scope": list, "secondary_indices"
 
 
 def normalize_arg(arg_name, old_value):
-    def _make_list(*_args):
+    def _make_list(_args):
+        if isinstance(_args, (str, bytes, int, float)):
+            return [_args]
+        if _args is None:
+            return []
+        if isinstance(_args, (set, frozenset, dict)):
+            raise ValueError(
+                "{} is incompatible for normalisation.".format(type(_args))
+            )
         return list(_args)
 
     type_to_normalize = {list: _make_list}
@@ -179,7 +187,7 @@ def normalize_arg(arg_name, old_value):
     if isinstance(old_value, _ARGS_TO_TYPE[arg_name]):
         return old_value
     elif old_value is None:
-        new_value = args_to_normalize[arg_name]()
+        new_value = _ARGS_TO_TYPE[arg_name]()
     elif not isinstance(old_value, _ARGS_TO_TYPE[arg_name]):
         new_value = args_to_normalize[arg_name](old_value)
     else:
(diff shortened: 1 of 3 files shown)


tests/io_components/test_utils.py
```diff
@@ -71,6 +71,7 @@ def testcombine_metadata_lists():
         (("a", "abc", None), ("a", ["abc"], [])),
         (("a", None, "abc"), ("a", [], ["abc"])),
         (("a", None, None), ("a", [], [])),
+        (("a", ("abc"), {"abc"}), ("a", ["a"], ["a"])),
```
fjetter: Expected should probably be `("a", ["abc"], ["abc"])`, shouldn't it?

aniruddhgoteti: Yes sorry, but I re-read the bug description and now raise an expection if sets or tuples are given as output. Am I going in the right direction?

kartothek/io_components/utils.py
```diff
@@ -168,6 +168,10 @@ def validate_partition_keys(
 
 def normalize_arg(arg_name, old_value):
     def _make_list(*_args):
+        if isinstance(_args, (str, bytes)):
+            return [_args]
+        if isinstance(_args, (set, tuple)):
```
lr4d: this is redundant, you already have `return list(_args)` below. Also note:
```
In [1]: def a(*args): 
   ...:     print(type(args)) 
   ...:     print(args) 
   ...:                                                                                                                                                      

In [2]: a(1)                                                                                                                                                 
<class 'tuple'>
(1,)

In [3]: a([1,2,3])                                                                                                                                           
<class 'tuple'>
([1, 2, 3],)
```

tests/io_components/test_utils.py
```diff
@@ -71,6 +71,7 @@ def testcombine_metadata_lists():
         (("a", "abc", None), ("a", ["abc"], [])),
         (("a", None, "abc"), ("a", [], ["abc"])),
         (("a", None, None), ("a", [], [])),
+        (("a", ("abc"), {"abc"}), ("a", ["abc"], ["abc"])),
```
fjetter: Actually we don't want to support sets here. Sets do not maintain order and most of the normalised features require strict ordering, e.g. partition on will behave differently when using differently ordered input

aniruddhgoteti: Done, Please verify

PR #140, opened 2019-09-25
Increase minimal version for urlquote

fjetter:
The version before this had packaging issues and race conditions when threading


xhochy:
Should we also update this in the conda recipe of the current release?

fjetter:
@xhochy Yes, I'll create a PR

PR #141, opened 2019-09-26
Input to `normalize_args` is properly normalized to `list`

### LINKED ISSUE #56: Normalize input to `list` properly in `normalize_args`

aniruddhgoteti:
# Description:

Normalise properly to lists within normalise_arg
Duplicate of PR #138 but due to its messed up commit history, this new PR

- [x] Closes #56 
- [x] Changelog entry

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Changelog
 
 Version 3.X.X (2019-09-XX)
 ==========================
+- Input to ``normalize_args`` is properly normalized to ``list``
 - ``MetaPartition.load_dataframes`` now raises if table in ``columns`` argument doesn't exist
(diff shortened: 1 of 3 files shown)


aniruddhgoteti:
@fjetter any more changes required in this PR?

PR #142, opened 2019-09-28
Provide default table name for kartothek.io.*read_table* methods

### LINKED ISSUE #121: Make `table` kwarg in `read_table*` use the default table name if unspecified
Currently, one can create a dataset without specifying a table name, and a default table name (`"table"`) will be assigned as the name of that table.
However, when one then tries to read that dataset using `read_table`, it becomes necessary to explicitly specify a table name:
```python
>>> from kartothek.io.eager import read_table
>>> read_table(dataset_uuid=dataset_uuid, store=store_factory)

TypeError: Parameter `table` is not optional.
```

I suggest the `table` parameter in `kartothek.io.*.read_table*` functions takes the default table name of kartothek if it is not specified, just as the store/update functions do.

This would fail for datasets created with old kartothek versions (due to the change in default table name in v3.0), but AFAIK most of the `read_table` implementations were not available at that time. In case of failure, in the error message it could be clarified that the default table name was used

SvoONs:
# Description:

This PR makes ``read_table`` methods in the ``kartothek.io`` module use the default table name as it is already specified for store/update methods in ``kartothek.io_components.metapartition`` if none is given.

- [x] Closes #121 
- [x] Changelog entry

### DIFF
--- a/kartothek/io/dask/delayed.py
+++ b/kartothek/io/dask/delayed.py
@@ -21,6 +21,7 @@ from kartothek.io_components.docs import default_docs
 from kartothek.io_components.gc import delete_files, dispatch_files_to_gc
 from kartothek.io_components.merge import align_datasets
 from kartothek.io_components.metapartition import (
+    SINGLE_TABLE,
     MetaPartition,
     parse_input_to_metapartition,
 )
@@ -386,7 +387,7 @@ def read_dataset_as_delayed(
 def read_table_as_delayed(
     dataset_uuid=None,
     store=None,
-    table=None,
+    table=SINGLE_TABLE,
     columns=None,
     concat_partitions_on_primary_index=False,
     predicate_pushdown_to_io=True,
@@ -415,9 +416,9 @@ def read_table_as_delayed(
 
     Parameters
     ----------
+    table: str, optional
+    The table to be loaded. If none is specified, the default 'table' is used.
     """
-    if table is None:
-        raise TypeError("Parameter `table` is not optional.")
     if not isinstance(columns, dict):
         columns = {table: columns}
     mps = read_dataset_as_delayed_metapartitions(
--- a/kartothek/io/testing/utils.py
+++ b/kartothek/io/testing/utils.py
@@ -6,6 +6,7 @@ import storefact
 from simplekv.decorator import StoreDecorator
 
 from kartothek.io.eager import store_dataframes_as_dataset
+from kartothek.io_components.metapartition import SINGLE_TABLE
 
 
 def create_dataset(dataset_uuid, store_factory, metadata_version):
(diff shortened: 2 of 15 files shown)


tests/io/eager/test_read.py
```diff
@@ -145,3 +145,16 @@ def test_read_table_simple_list_for_cols_cats(dataset, store_session):
     expected_df = expected_df.astype("category")
 
     pdt.assert_frame_equal(df, expected_df, check_dtype=False, check_like=True)
+
+
+def test_write_and_read_default_table_name(store_session):
```
lr4d: you will want to add/change tests that check that this behavior also works for the `read_table` functions. You may want to change https://github.com/JDASoftwareGroup/kartothek/blob/dcee179b7a69d022f5001bd309deba806c235c31/tests/io/dask/delayed/test_read.py#L27 so that no `table` kwarg is passed (and probably replace the test code which defines the tables as "`core`" to `SINGLE_TABLE`).

kartothek/io/dask/delayed.py
```diff
@@ -415,9 +416,11 @@ def read_table_as_delayed(
 
     Parameters
     ----------
+    table: str, optional
+    The table to be loaded. If none is specified, the default 'table' is used.
     """
     if table is None:
-        raise TypeError("Parameter `table` is not optional.")
+        table = SINGLE_TABLE
```
amerkel2: Why not just give `SINGLE_TABLE` as default value for the `table` parameter instead of doing this check for `None`?

SvoONs: ups... fixed it.

tests/io/eager/test_read.py
SvoONs: I added a new commit where I removed the table kwarg in above code snippet and refactored tests as far as required to use `SINGLE_TABLE` instead of `core`.

kartothek/io/testing/utils.py
```diff
@@ -20,13 +20,13 @@ def create_dataset(dataset_uuid, store_factory, metadata_version):
...
-            "indices": {"core": {val: ["cluster_2"] for val in df.TARGET.unique()}},
+            "data": [("table", df.copy(deep=True)), ("helper", df_helper)],
+            "indices": {"table": {val: ["cluster_2"] for val in df.TARGET.unique()}},
         },
         {
             "label": "cluster_2",
-            "data": [("core", df.copy(deep=True)), ("helper", df_helper)],
-            "indices": {"core": {val: ["cluster_2"] for val in df.TARGET.unique()}},
+            "data": [("table", df.copy(deep=True)), ("helper", df_helper)],
+            "indices": {"table": {val: ["cluster_2"] for val in df.TARGET.unique()}},
```
lr4d: Tiny nitpick: I would also use `SINGLE_TABLE` here so if it gets renamed tests do not need to be changed again (but some people may argue you shouldn't have that logic for a test)

SvoONs: Changed it to `SINGLE_TABLE`

SvoONs:
looks like `store_dataframes_as_dataset` doesn't accept a single `pandas.DataFrame` as an input anymore?

lr4d:
That's right. By making the interface consistent we now require the input to be an iterable of dicts or dataframes.

lr4d:
No, you're right. I thought it was never officially supported but it is in the docstring. Will create a PR to fix this

PR #147, opened 2019-10-02
Creating dataset with non existing to raise an error

### LINKED ISSUE #47: Creating a dataset with a non-existing column as explicit index does not raise
```python
# -*- coding: utf-8 -*-

import pandas as pd
from functools import partial
import storefact

from kartothek.io.eager import store_dataframes_as_dataset
from kartothek.core.index import ExplicitSecondaryIndex


def get_storefactory(store_config):
    return partial(storefact.get_store, **store_config)


store_config = {"type": "hfs", "path": "some_directory"}
store = get_storefactory(store_config)
partitions = [
    {
        "label": "part1",
        "data": [("core", pd.DataFrame({"p": [1, 2]}))],
        "indices": {"x": ExplicitSecondaryIndex("x", {1: ["part1"], 2: ["part2"]})},
    }
]
store_dataframes_as_dataset(
    dfs=partitions,
    store=store,
    metadata={"dataset": "metadata"},
    dataset_uuid="dataset_uuid",
)

# Returns: DatasetMetadata(uuid=dataset_uuid, tables=['core'], partition_keys=[], metadata_version=4, indices=['x'], explicit_partitions=True)

```

Kshitij68:
# Description:

- [ ] Closes #47 
- [ ] Changelog entry


kartothek/io_components/metapartition.py
```diff
@@ -1715,3 +1712,5 @@ def _ensure_valid_indices(secondary_indices, mp_indices):
                 f"Expected index columns: {secondary_indices}"
                 f"Provided index: {mp_indices}"
             )
+    else:
+        raise ValueError("Please pass `secondary_indices` and `data` ")
```
fjetter: I do not get this case. This function should ensure that a explicitly passed index dictionary does provide all indices which are compatible with the existing dataset.
Not sure if hitting this clause is a real scenario. Either way, the exception message sounds misleading for this case

Kshitij68: Yes, you are right. I have made the checks for `data` and `secondary_indices` independent of each other. 

Thanks for pointing out :)

tests/io/eager/test_write.py
```diff
@@ -273,3 +274,22 @@ def test_write_single_partition_different_partitioning(store_factory):
...
+def test_store_dataframes_as_dataset_does_not_allow_invalid_indices(store_factory):
+    partitions = [
+        {
+            "label": "part1",
+            "data": [("core", pd.DataFrame({"p": [1, 2]}))],
+            "indices": {"x": ExplicitSecondaryIndex("x", {1: ["part1"], 2: ["part2"]})},
+        }
+    ]
+
+    with pytest.raises(ValueError) as err:
```
fjetter: Rather use `with pytest.raises(ValueError, match="In table core, no column corresponding to index x") than asserting the exception youself

kartothek/io_components/metapartition.py
```diff
@@ -1696,15 +1694,15 @@ def parse_input_to_metapartition(
...
-def _ensure_valid_indices(secondary_indices, mp_indices):
-    if not secondary_indices:
-        if mp_indices:
-            raise ValueError(
-                "Incorrect indices provided for dataset.\n"
-                f"Expected index columns: {secondary_indices}"
-                f"Provided index: {mp_indices}"
-            )
-    else:
+def _ensure_valid_indices(mp_indices, secondary_indices=None, data=None):
```
fjetter: Can you move this function to `io_components/utils.py`? The behavior is closely entangled to `_ensure_compatible_indices` and should keep them close / refactor it to be less brittle

tests/io/eager/test_write.py
Kshitij68: Done!

kartothek/io_components/metapartition.py
Kshitij68: Done. 
Have added a TODO to do refactoring also.

PR #148, opened 2019-10-02
Remove passing test case marked as xfailed

Kshitij68:
# Description:

The test case was marked as xfailed.
Not adding a changelog entry as this seems irrelevant for end-user or developer


Kshitij68:
Realized that this won't work as we still are on py_arrow0.13 in CI.

PR #149, opened 2019-10-04
Prototype for table interface

fjetter:
This is a prototype for a new user interface which would help to get rid of the many inconsistencies we currently have between the functional approaches.

The core idea is to offer *one* entry point for a user via a single class (`Table`). Depending on what the user wants to do the class creates specialised instances which are able to read or write (never both). For examples, see the test implementations for `test_read` and `test_update`.

The classes are built in such a way that method chaining can be used to set options and finally execute the task (read or write/commit).

The implementation should be taken with a grain of salt and is not mature.

I'd like to only include features we definitely want to keep in the long run in this interface and once we feel comfortable with it we can deprecate all other functions as part of the public API.

This is also designed with a single-table specification in mind. To fully deprecate the multi-table specification, I believe, we need to offer an alternative but this is out of scope for this PR.

For this first iteration, I'd like to collect feedback on the interface and class structure in particular.


amerkel2:
I like the interface of having one entry point instead of many inconsistenc functions. I wonder what's the use case of setting the options on the reader/writer after instantiation, though. This could also create the potential for non-obvious behavior, e.g., what happens when I change the options between writing and committing?

lr4d:
What do you think about having a number of partial/wrapper functions to offer a single consistent entry point to users?

fjetter:
The motivation for separating initialisation and option setting was primarily to keep the method signatures small. I realised while implementing, though, that doing it this way introduces a lot of (potentially buggy) boilerplate code. Simply using the constructor of TableReader/Writer, i.e. the read/write method of Table would decrease this by a long shot and would allow us to also properly use the `attr` package for the class definition.

crepererum:
> This is also designed with a single-table specification in mind. To fully deprecate the multi-table specification, I believe, we need to offer an alternative but this is out of scope for this PR.

 If we move away from a multi-table approach (IMHO the right thing to do), then we have "dataset = table". So why are we naming the new interface "Table" then? Shouldn't it be called "Dataset"? I also think that if we design a new interface, we should do this after/with the "single-table-only" approach, because the multi-table stuff is all over the place (many things, including the inputs and outputs of our functions, are dicts now).

fjetter:
> If we move away from a multi-table approach (IMHO the right thing to do)

The way we're implementing the multi table approach atm is not sustainable in the long run. I could, however see a possibility for a multi-table write functionality since I could see value in this. On the same line there may be a multi-table alignment mechanism with *very simple* equi-join logics involved. Although related, I believe this is a topic for another thread/issue.

> then we have "dataset = table". So why are we naming the new interface "Table" then? Shouldn't it be called "Dataset"?

This is intentional. IMHO, the term "Table" is much better defined than the term "Dataset" and fits better to what kartothek actually stores once we abolish the multi-table specs. Wouldn't you agree?

> we should do this after/with the "single-table-only" 

I think we should do this simultaneously. By defining a clean user interface we can hide the messiness of the existing implementation and clean up lazily.
Once we announce this to be the new interface, the old one will receive deprecation warnings and I'd like to drop the actual code quite fast afterwards.


In any case, here I'd like to collect ideas about how the interface should look like. This is only a Draft PR for a reason and the implementation here is throw-away(-ish) code.

crepererum:
> This is intentional. IMHO, the term "Table" is much better defined than the term "Dataset" and fits better to what kartothek actually stores once we abolish the multi-table specs. Wouldn't you agree?

Personally not really, since we've jumped back and forth between these terms quite often now and I don't think calling it a table now out of a sudden would help.

xhochy:
> > This is intentional. IMHO, the term "Table" is much better defined than the term "Dataset" and fits better to what kartothek actually stores once we abolish the multi-table specs. Wouldn't you agree?
> 
> Personally not really, since we've jumped back and forth between these terms quite often now and I don't think calling it a table now out of a sudden would help.

Talking to people from Iceberg and ORC, `table (format)` is the more commonly known term what we are doing here in ktk. Often people associate with dataset a loose collection of information in files. This also matches what we want to handle in ktk but is not a well-suited name for interacting with other data engineering communities.

kartothek/core/factory.py
```diff
@@ -99,26 +106,33 @@ def _instantiate_metadata_cache(self):
                 # backwards compat
                 self._cache_metadata = self._ds_callable()
             else:
-                self._cache_metadata = DatasetMetadata.load_from_store(
-                    uuid=self.dataset_uuid,
-                    store=self.store,
-                    load_schema=self.load_schema,
-                    load_all_indices=self.load_all_indices_flag,
-                )
+                try:
```
xhochy: 🤬 You know I don't like `except KeyError` for existence checks.

PR #157, opened 2019-10-14
Refactor dispatch logic

### LINKED ISSUE #134: Dispatch by ignores predicates
### Problem description

Predicates are not respected if used in conjunction with `dipatch_by`

### Example code

```python

import storefact
import tempfile
import pandas as pd
import shutil
from kartothek.io.eager import store_dataframes_as_dataset, build_dataset_indices
from kartothek.io.dask.delayed import read_dataset_as_delayed


folder = tempfile.mkdtemp(prefix="ktk_test_")
s = storefact.get_store_from_url(f"hfs://{folder}")


df = pd.DataFrame({"a": range(10), "b": [0, 1] * 5, "c": [0, 1, 2, 3, 4] * 2})
store_dataframes_as_dataset(s, "test", [df], partition_on=["c"])
build_dataset_indices(lambda: s, "test", ["b"])

assert (
        len(
            read_dataset_as_delayed(
                dataset_uuid="test",
                store=lambda: s,
                predicates=[[("b", "in", [1])]],
                dispatch_by=["b"],
            )
        )
        == 1
    )

```

fjetter:
# Description:

This refactors the dispatch logic, such that the partition pruning uses the same logic for all code paths. 

- [x] Closes #134
- [x] Changelog entry


amerkel2:
Changes look good to me, but some tests are failing. One reason seems that `dispatch_by` is not impemented for reading as DDF, which no longer seems to be considered by the tests.

fjetter:
We have some other issues where the schema violation produces different output. Will need to look into this...

fjetter:
Some unintended changes to the MetaPartition slipped in which caused different exception messages. Will factor these changes out for another PR

fjetter:
I moved the `docs` submodule to `core` to make better use of the `default_docs` decorator

kartothek/serialization/_generic.py
```diff
@@ -167,6 +172,23 @@ def check_predicates(predicates):
                     )
 
 
+def remove_columns_from_predicates(
```
crepererum: the function name, the docs and the parameters are somewhat misleading here. It does not remove the columns that are in `columns` but keeps the columns that are exactly in there. Or the implementation is wrong (`if col in columns`).

kartothek/io_components/read.py
```diff
@@ -133,35 +124,49 @@ def dispatch_metapartitions_from_factory(
...
+    columns = _columns_in_predicates(predicates)
+
+    if set(dataset_factory.partition_keys).union(columns):
+        dataset_factory = dataset_factory.load_partition_indices()
 
     # Load the necessary indices
     for column in columns:
         if column in dataset_factory.indices:
             dataset_factory = dataset_factory.load_index(column)
+    return dataset_factory
```
crepererum: why does this function return the factoy and `_load_indices_dispatch_by` does not? Alos, all new functions here missing docstrings

tests/io_components/test_read.py
```diff
@@ -95,8 +94,7 @@ def test_dispatch_metapartitions_query_partition_on(
 @pytest.mark.parametrize(
     "predicates",
     [
-        [[("P", "==", 2)], [("TARGET", "==", 500)]],
-        [[("P", "in", [2])], [("TARGET", "in", [500])]],
+        [[("L", "in", [2])], [("TARGET", "in", [500])]],
```
fjetter: Please review this thoroughly. It seems to me that we had a bug in the partition pruning before this refactoring. The previously tested behaviour seems wrong to me.

kartothek/io/testing/read.py
```diff
@@ -267,7 +267,6 @@ def test_read_dataset_as_dataframes_predicate(
         [[("P", "in", [2]), ("TARGET", "==", 2)]],
         [[("P", "!=", 1), ("L", "==", 2)]],
         [[("P", "!=", 1), ("L", "in", [2])]],
-        [[("P", ">", 2)], [("TARGET", ">=", 2)]],
```
fjetter: Please review this thoroughly. It seems to me that we had a bug in the partition pruning before this refactoring. The previously tested behaviour seems wrong to me.

fjetter:
I could remove a solid chunk of unnecessary code. Please review carefully, though. I indicated two test cases which seemed to have tested erroneous behaviour before.

This PR also adds a few more functions to our _public_ stable API. If there are any comments about this, feel free to suggest changes.

kartothek/serialization/_generic.py
fjetter: I renamed the function name and added docs with an example

kartothek/io_components/read.py
fjetter: Refactored such that this function is not used anymore

amerkel2:
> This PR also adds a few more functions to our _public_ stable API (`serialization`). If there are any comments about this, feel free to suggest changes.

Since it's public API, shouldn't we mention this in the changelog?

fjetter:
@amerkel2 extended the changelog and added more documentation

@crepererum anything from your side?

PR #158, opened 2019-10-14
No schema validation in internal ops

fjetter:
# Description:

This should improve overall performance for very large datasets by introducing slots for objects which are created frequently.

The schema validation within the MP seemed quite unnecessary so I removed them. The change in exception messages stems from hashing the schemas at the very final reduction which isn't performed by default.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,10 @@ Version 3.X.X (2019-09-XX)
 - Fixed DeprecationWarning in pandas ``CategoricalDtype``
 - Fixed broken docstring for `store_dataframes_as_dataset`
 - Make ``kartothek.io.*read_table*`` methods use default table name if unspecified
+- Internal operations no longer perform schema validations. This will improve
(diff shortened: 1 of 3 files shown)


crepererum:
There are two changes in this PR:

- schema validation on internal ops: seems very legit, also saves a solid amount of tracking data in the schema wrapper
- slots: any numbers on this?

fjetter:
> slots: any numbers on this?

No numbers, yet. I'm also quite curious but haven't found the time, yet.

Do you see any downsides? Thee only thing I see is that it doesn't allow modifications to the object definition itself (e.g. define new attributes at runtime) which is something I would rather disable anyhow

crepererum:
> Do you see any downsides?

Complexity. If it doesn't do anything noticeable, we shouldn't have this in our code.

fjetter:
I removed the slots again. I tried it with a 50k partition dataset but couldn't measure an impact.

PR #160, opened 2019-10-15
pyarrow 0.15.0 compat

### LINKED ISSUE #155: pyarrow 0.15.0 compat
pyarrow nightly tests are currently broken, mostly around the predicate pushdown. We'll need to investigate and fix this so we can whitelist 0.15.0 as well.

crepererum:
Fixes #155


xhochy:
@crepererum Any other upstream issues besides ARROW-6872 in here?

kartothek/serialization/_parquet.py
```diff
@@ -56,9 +76,18 @@ def _reset_dictionary_columns(table):
     2. Massive performance issue due to non-fast path implementation https://issues.apache.org/jira/browse/ARROW-5089
     """
     for i in range(table.num_columns):
-        if pa.types.is_dictionary(table[i].type):
-            new_col = table[i].cast(table[i].data.chunk(0).dictionary.type)
-            table = table.remove_column(i).add_column(i, new_col)
+        col = table[i]
```
fjetter: Wondering if we should skip this entire function for arrow >=0.15.0. At least according to the ticket this has been resolved by now

crepererum: Depends. As far as I understand this, this would mean that the files produced under 0.15.0 would be difficult to be consumed by older versions. What's our policy on that?

crepererum:
> @crepererum Any other upstream issues besides ARROW-6872 in here?

No, that's the only one.

kartothek/serialization/_parquet.py
fjetter: The performance issue only stems from the arrow->parquet conversion. The issue was that there was no fast-path implemented for dict encoded columns and they needed to be decoded first before they were passed to parquet. 
That's not the critical performance issue but rather that the entire array was decoded for every row group (i.e. 100 row groups -> 100 times decoding instead of only partial decoding or all but only once).


afaik, there is / was never an issue reading them back again.

kartothek/serialization/_parquet.py
```diff
@@ -56,9 +76,18 @@ def _reset_dictionary_columns(table):
...
+            new_col = col.cast(new_type)
+            if ARROW_LARGER_EQ_0150:
+                field = table.field(i)
+                new_field = pa.field(
+                    field.name, new_type, field.nullable, field.metadata
+                )
+                table = table.remove_column(i).add_column(i, new_field, new_col)
+            else:
+                table = table.remove_column(i).add_column(i, new_col)
     return table
```
fjetter: I was rather thinking

```python
def _reset_dictionary_columns(table):
    if ARROW_LARGER_EQ_0150:
        for i in range(table.num_columns):
            if pa.types.is_dictionary(table[i].type):
                new_col = table[i].cast(table[i].data.chunk(0).dictionary.type)
                table = table.remove_column(i).add_column(i, new_col)
        col = table[i]
        if pa.types.is_dictionary(col.type):
            new_type = col.data.chunk(0).dictionary.type
            new_col = col.cast(new_type)
            table = table.remove_column(i).add_column(i, new_col)
    return table
```

Sorry if this wasn't clear

crepererum: I think you mean "you don't need to implement the arrow 0.15.0 compat for _reset_dictionary_columns since it isn't used under arrow 0.15.0". Your code example doesn't make much sense to me though.

crepererum: the current code is doing what you proposed I think. the function isn't called under arrow 0.15.0

fjetter:
Builds break because it cannot find dask 2.6.0 for osx oO

PR #164, opened 2019-10-17
Allow empty tables to be part of a partition

fjetter:
# Description:

Fix an issue where cross-table validation would fail if one of the tables is empty (empty files are allowed as long as there is a file)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,8 @@ Version 3.X.X (2019-09-XX)
   defer the validation in case of inconsistencies to the final commit. Exception
   messages will be less verbose in these cases as before.
 - Add support for pyarrow 0.15.0
+- Fix an issue where an empty dataframe of a partition in a multi-table dataset
+  would raise a schema validation exception
 
 
 Version 3.4.0 (2019-09-17)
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -19,6 +19,7 @@ Version 3.X.X (2019-09-XX)
   defer the validation in case of inconsistencies to the final commit. Exception
   messages will be less verbose in these cases as before.
 - Add support for pyarrow 0.15.0
+- Allow empty tables to be part of a partition
```
crepererum: I think, this was allowed before, since there is a very extensive null-handling in `_determine_schemas_to_compare`. The bug you're fixing here is about SHARED columns, i.e. when multiple tables are present. What I am not sure about is exact trigger that is required to hit this:

- a single partition of a multi-table dataset contains an empty DF (check is already done on the partition-level)
- all partitions of one table of a multi-table dataset are empty (check is done only in the reducer-part).

fjetter: The `_determine_schemas_to_compare` assumes you pass multiple, supposedly identical schemas. It the determines the schema with the most information, i.e. the least null columns and promotes this to a reference column. At the same time it determines and tracks the null columns per schema such that they can be ignored in the following comparison.

This shared_column reference check never hit this code path. To trigger the change I am currently proposing is e.g.

```
table1:
A: [1, 2, 3]
B: [None, None, None]

table2:
B: ["A", "B", "C"]
C: [1, 2, 3]
```

or similarly
```
table1:
A: [1, 2, 3]
B: ["A", "B", "C"]

table2:
B: []
C: []
```

I don't think we covered this edge case in the past

fjetter:
I rephrased the changelog entry

PR #146, opened 2019-10-01
Fix broken docstring parser and add test case

Kshitij68:
# Description:

Given that there are no doc-tests, it's a little hard to tell if this implementation is not going to lead unforeseen bugs.

- Closes #145 
- [x] Changelog entry

PR #166, opened 2019-10-17
Fix broken test case

Kshitij68:
# Description:

This fixes the broken test case that was produced in PR#146

- [x] Closes #146 
- [x] No changelog entry required


fjetter:
Build breaks due to flake8 

tests/io_components/test_docs.py:21:1: F401 'kartothek.io.eager.store_dataframes_as_dataset' imported but unused

PR #168, opened 2019-10-21
Hash buckets in update_ddf

fjetter:
# Description:

This allows for a deterministic bucketing based on hashing. For large scale data processing this requires a bit more ram (10-20%) but scales very well otherwise.
I would suggest to expose a kwarg `bucket_by` which exposes this parameter publicly and otherwise acts similar to partition_on

### DIFF
--- a/kartothek/io/dask/_update.py
+++ b/kartothek/io/dask/_update.py
@@ -2,9 +2,10 @@
 
 
 from functools import partial
+from typing import List
 
-import dask.array as da
 import numpy as np
+import pandas as pd
 
 from kartothek.io_components.metapartition import (
     MetaPartition,
@@ -13,6 +14,26 @@ from kartothek.io_components.metapartition import (
 from kartothek.io_components.utils import sort_values_categorical
 
 from ._utils import map_delayed
+
+_KTK_HASH_BUCKET = "__KTK_HASH_BUCKET"
+
+
+def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
+    """
+    Categorize each row of `df` based on the data in the columns `subset`
+    into `num_buckets` values. This is based on `pandas.util.hash_pandas_object`
+    """
+
+    if subset is None:
+        subset = df.columns
+    hash_arr = pd.util.hash_pandas_object(df[subset], index=False)
+    buckets = hash_arr % num_buckets
+
+    available_bit_widths = np.array([8, 16, 32, 64])
+    mask = available_bit_widths > np.log2(num_buckets)
+    bit_width = min(available_bit_widths[mask])
+    df[_KTK_HASH_BUCKET] = buckets.astype(f"uint{bit_width}")
+    return df
 
 
 def _update_dask_partitions_shuffle(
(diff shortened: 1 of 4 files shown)


kartothek/io/dask/_update.py
```diff
@@ -14,6 +15,26 @@
...
+
+def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
+    """
+    Categorize each row of `df` based on the data in the columns `subset`
+    into `num_buckets` values. This is based on `pandas.util.hash_pandas_object`
+    """
+
+    if subset is None:
+        subset = df.columns
+    hash_arr = pd.util.hash_pandas_object(df[subset], index=False)
```
crepererum: what determinism guarantees do we have here? is there any host-based / time-based seed or something involved?

fjetter: From my understanding it's deterministic and doesn't involve host based seeds but https://github.com/pandas-dev/pandas/blob/af1585d697aed306ec15c56f524d8481742ba6eb/pandas/core/util/hashing.py#L26

kartothek/io/dask/_update.py
```diff
@@ -14,6 +15,26 @@
...
+    buckets = hash_arr % num_buckets
+
+    available_bit_widths = np.array([8, 16, 32, 64])
+    mask = available_bit_widths > np.log2(num_buckets)
+    bit_width = min(available_bit_widths[mask])
```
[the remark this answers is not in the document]
fjetter: We shouldn't make this dependent on the size of the partition data frame. We much rather want to have a deterministic hashing behavior s.t. A given value will always end up in the same bucket no matter how large the df is.

fjetter:
I guess it would make sense to expose `bucket_by` as a parameter. Would possibly also make sense for other backends although they can obviously not write compact

xhochy:
Before this we would have even sized buckets, now they can be uneven if you have a lot of duplicate rows?

lr4d:
> In any case I prefer an approach where we have some kind of a deterministic behaviour which is not coupled to the internals of dask which is currently the case by relying on the internal partitioning.

Why do you want it detached of the internals of dask, instead of e.g. having this as a special case of `ddf.repartition`?

kartothek/io/dask/_update.py
lr4d: I have added a test to address this concern

kartothek/io/dask/_update.py
```diff
@@ -30,32 +51,35 @@ def _update_dask_partitions_shuffle(
     if ddf.npartitions == 0:
         return ddf
 
-    splits = np.array_split(
-        np.arange(ddf.npartitions), min(ddf.npartitions, num_buckets)
+    if num_buckets is not None:
+        if sort_partitions_by:
+            bucket_by = [sort_partitions_by]
```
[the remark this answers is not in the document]
amerkel2: sounds reasonable to me

kartothek/io/dask/_update.py
```diff
@@ -30,32 +51,37 @@ def _update_dask_partitions_shuffle(
...
 
-    splits = np.array_split(
-        np.arange(ddf.npartitions), min(ddf.npartitions, num_buckets)
+    if num_buckets is not None:
+        if isinstance(sort_partitions_by, list):
+            bucket_by = sort_partitions_by
+        elif isinstance(sort_partitions_by, str):
+            bucket_by = [sort_partitions_by]
+        else:
+            bucket_by = None
```
marco-neumann-jdas: this should raise an `ValueError`. The type is neither a list nor a string, but they user might have used a somewhat valid type (e.g. tuple) or something weird (like a random object) and we're silently ignoring it.

fjetter: Promoted the bucket_by to a user input as discussed above https://github.com/JDASoftwareGroup/kartothek/pull/168#discussion_r347443451

kartothek/io/dask/dataframe.py
```diff
@@ -119,6 +119,7 @@ def update_dataset_from_ddf(
     default_metadata_version=DEFAULT_METADATA_VERSION,
     partition_on=None,
     factory=None,
+    bucket_by=None,
```
marco-neumann-jdas: docs?

amerkel2: Add to docstring please

PR #169, opened 2019-10-22
Update docs and make test cases more robust

Kshitij68:
# Description:

Update docstrings 

Make test cases for default_docs more robust

- [x] Changelog entry

### DIFF
--- a/kartothek/core/docs.py
+++ b/kartothek/core/docs.py
@@ -20,14 +20,20 @@ _PARAMETER_MAPPING = {
         the default behavior and create a new label from all input labels
         (depending on the matches this might be more than two values)""",
     "metadata_merger": """
-    label_merger: callable, optional
+    metadata_merger: callable, optional
          By default partition metadata is combined using the :func:`~kartothek.io_components.utils.combine_metadata` function.
          You can supply a callable here that implements a custom merge operation on the metadata dictionaries
          (depending on the matches this might be more than two values).""",
+    "table": """
+    table: str, optional
+        The table to be loaded. If none is specified, the default 'table' is used.""",
     "tables": """
     tables : list of str
         A list of tables to be loaded. If None is given, all tables of
         a partition are loaded""",
+    "table_meta": """
+    table_meta: Dict[str, SchemaWrapper]
+        The dataset table schemas""",
     "columns": """
     columns : dict of list of string, optional
         A dictionary mapping tables to list of columns. Only the specified
@@ -46,41 +52,34 @@ _PARAMETER_MAPPING = {
     output_dataset_metadata: dict, optional
         Metadata for the merged target dataset. Will be updated with a
         `merge_datasets__pipeline` key that contains the source dataset uuids for
-        the merge.
-""",
+        the merge.""",
     "output_store": """
     output_store : callable
         Factory function producing a KeyValueStore.
         If given, the resulting dataset is written to this store. By default
-        the input store
-""",
+        the input store""",
     "metadata": """
     metadata : dict, optional
-        A dictionary used to update the dataset metadata.
-""",
+        A dictionary used to update the dataset metadata.""",
     "dataset_uuid": """
     dataset_uuid: str
-        The dataset UUID
-""",
+        The dataset UUID""",
     "metadata_version": """
     metadata_version: int, optional
-        The dataset metadata version
-""",
+        The dataset metadata version""",
     "partition_on": """
     partition_on: list
         Column names by which the dataset should be partitioned by physically.
         These columns may later on be used as an Index to improve query performance.
         Partition columns need to be present in all dataset tables.
-        Sensitive to ordering.
-""",
+        Sensitive to ordering.""",
     "predicate_pushdown_to_io": """
     predicate_pushdown_to_io: bool
         Push predicates through to the I/O layer, default True. Disable
         this if you see problems with predicate pushdown for the given
         file even if the file format supports it. Note that this option
         only hides problems in the storage layer that need to be addressed
-        there.
-""",
+        there.""",
     "delete_scope": """
     delete_scope: list of dicts
         This defines which partitions are replaced with the input and therefore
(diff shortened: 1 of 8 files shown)


kartothek/io/eager.py
```diff
@@ -393,6 +387,18 @@ def commit_dataset(
         Input partition to be committed.
 
     """
+    if output_dataset_uuid is not None:
+        warnings.warn(
+            "The keyword `concat_partitions_on_primary_index` is deprecated and will be removed in the next major release.",
```
lr4d: wrong keyword `concat_partitions_on_primary_index`

kartothek/io/eager.py
```diff
@@ -280,7 +274,7 @@ def read_table(
         >>> df = read_table(store, 'dataset_uuid', 'core')
 
     """
-    if concat_partitions_on_primary_index:
+    if concat_partitions_on_primary_index is not False:
```
lr4d: why was this change necessary?

kartothek/io/eager.py
```diff
@@ -608,6 +614,24 @@ def write_single_partition(
     -------
     An empty :class:`~kartothek.io_components.metapartition.MetaPartition` referencing the new files
     """
+    if metadata is not None:
+        warnings.warn(
+            "The keyword `metadata` is deprecated and will be removed in the next major release.",
+            DeprecationWarning,
+        )
+
+    if overwrite is not False:
```
lr4d: is this deprecated?

kartothek/io/eager.py
lr4d: is this deprecated?

kartothek/io/eager.py
```diff
@@ -393,6 +387,18 @@ def commit_dataset(
         Input partition to be committed.
 
     """
+    if output_dataset_uuid is not None:
+        warnings.warn(
+            "The keyword `concat_partitions_on_primary_index` is deprecated and will be removed in the next major release.",
+            DeprecationWarning,
+        )
+
+    if df_serializer is not None:
```
lr4d: is this deprecated?

tests/core/test_docs.py
```diff
@@ -76,9 +76,33 @@
...
+        store: str
+            This is an argument
+
+        """
+
+    resultant_docstrings = str(dummy_function.__doc__)
+    assert "This is a dummy_function" in resultant_docstrings
+    assert "This is an argument" in resultant_docstrings
+    assert (
+        "Factory dummy_function producing a KeyValueStore" not in resultant_docstrings
```
lr4d: Should be `Factory function producing a KeyValueStore.`. Also, can you add a short comment/docstring here to explain what you're doing?

kartothek/io/eager.py
Kshitij68: The argument is not used anywhere. Made it clear in the warning statement also.

kartothek/io/eager.py
Kshitij68: Passing `concat_partitions_on_primary_index` as anything should also raise a warning. 
If not making the above change, passing 0 will not raise the warning

kartothek/io/eager.py
Kshitij68: The argument has no use. Made it clear in the warning itself.

kartothek/io/eager.py
Kshitij68: The argument has no use and hence put the warning. Made the warning statement more clear

kartothek/io/eager.py
Kshitij68: same reasoning as others

kartothek/io/iter.py
```diff
@@ -200,21 +200,23 @@ def update_dataset_from_dataframes__iter(
...
     Parameters
     ----------
-    df_generator: Iterable[Union[pd.DataFrame, Dict[str, pd.DataFrame]]]
-        The dataframe(s) to be stored.
 
     Returns
     -------
     The dataset metadata object (:class:`~kartothek.core.dataset.DatasetMetadata`).
     """
+    if load_dynamic_metadata is True:
```
fjetter: should be `is not None` since the entire parameter should've been removed in `3.0`

kartothek/core/docs.py
```diff
@@ -129,24 +124,45 @@
...
+        Dask bag partition size. Use a larger numbers to decrease scheduler load and overhead, use smaller numbers for a
+        fine-grained scheduling and better resilience against worker errors.""",
+    "metadata_storage_format": """
+    metadata_storage_format: str
+        Optional list of datastorage format to use. Currently supported is `.json` & `.msgpack.zstd"`""",
+    "df_generator": """
+    df_generator: Iterable[Union[pd.DataFrame, Dict[str, pd.DataFrame]]]
+        The dataframe(s) to be stored""",
+    "central_partition_metadata": """
+    central_partition_metadata: TODO
```
fjetter: this keyword should no longer exist (should've been removed in 3.0 already). In any case, it doesn't have any effect anymore. Can you find the usage and add also warnings, please?

kartothek/core/docs.py
```diff
@@ -129,24 +124,45 @@
...
-    partition_size: int
-        Amount of metapartitions to cluster into one dask partition.
-""",
+    partition_size: Optional[int]
+        Dask bag partition size. Use a larger numbers to decrease scheduler load and overhead, use smaller numbers for a
+        fine-grained scheduling and better resilience against worker errors.""",
+    "metadata_storage_format": """
+    metadata_storage_format: str
+        Optional list of datastorage format to use. Currently supported is `.json` & `.msgpack.zstd"`""",
+    "df_generator": """
```
fjetter: Not sure if I would count this as a "general" doc string. We usually kept very specific parameter documentation with the function it refers to. Is this used at more than one places?

kartothek/io/iter.py
Kshitij68: Aah, stupid mistake. Thanks.

I would recommend it being `is not True` as the default value is `True`

kartothek/core/docs.py
Kshitij68: Yes, it is used in `store_dataframes_as_dataset__iter` and `update_dataset_from_dataframes__iter`

kartothek/core/docs.py
Kshitij68: You are referring to `central_partition_metadata` right? This was used `update_dataset_from_dataframes__iter`. I have depreciated from that place.

lr4d:
@Kshitij68 build seems to be failing due to formatting. You can run the pre-commit hooks to fix this

PR #177, opened 2019-11-09
Add an optional argument to in dask bag API to pass n_partitions

### LINKED ISSUE #176: kartothek Dask bag API to have option for passing number of partitions
### Problem description

Please describe the current behavior, why it is a problem and what the expected behavior should be.

Currently `read_dataset_as_metapartitions_bag`, `read_dataset_as_dataframe_bag` & `build_dataset_indices__bag` have an option to pass `partition_size`. However, it is possible to create dask bags by passing `n_partitions` also.  
It would be nice if the user had an option to pass `partition_size` OR `n_partitions` 

Note: Dask allows passing both. Should we also allow?

Kshitij68:
# Description:

- [x] Closes #176 
- [x] Changelog entry

### DIFF
--- a/kartothek/core/docs.py
+++ b/kartothek/core/docs.py
@@ -146,6 +146,12 @@ _PARAMETER_MAPPING = {
     "partition_size": """
     partition_size: int
         Amount of metapartitions to cluster into one dask partition.
+        Note: Do not pass this argument and `n_partitions` together.
+""",
+    "n_partitions": """
+    n_partitions: int
+        The number of desired partitions
+        Note: Do not pass this argument and `partition_size` together.
 """,
 }
 
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -56,6 +56,7 @@ def read_dataset_as_metapartitions_bag(
     factory=None,
     dispatch_by=None,
     partition_size=None,
+    n_partitions=None,
 ):
     """
     Retrieve dataset as `dask.bag` of `MetaPartition` objects.
@@ -67,6 +68,11 @@ def read_dataset_as_metapartitions_bag(
     -------
     A dask.bag object containing the metapartions.
     """
+    if partition_size and n_partitions:
+        raise ValueError(
+            "Please provide only one of the argument `partition_size` or `n_partitions`"
+        )
+
     ds_factory = _ensure_factory(
         dataset_uuid=dataset_uuid,
         store=store,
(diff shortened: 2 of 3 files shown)


kartothek/core/docs.py
```diff
@@ -146,6 +146,12 @@
     "partition_size": """
     partition_size: int
         Amount of metapartitions to cluster into one dask partition.
+        Note: Do not pass this argument and `n_partitions` together.
```
fjetter: Maybe we can put a link to the actual Dask docs in here since that probably explains the intention of the parameters best. 
Ideally we'd be using [intersphinx](http://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) but as a first step a plain link would suffice

kartothek/io/dask/bag.py
```diff
@@ -67,6 +68,11 @@ def read_dataset_as_metapartitions_bag(
     -------
     A dask.bag object containing the metapartions.
     """
+    if partition_size and n_partitions:
+        raise ValueError(
+            "Please only one of the argument `partition_size` or `n_partitions`"
```
fjetter: Should be "Please *provide* only one [...]"

fjetter:
Tests fail due to #178

Kshitij68:
@fjetter I propose depreciating keyword argument `partition_size` and introducing `partition_config`. This new parameter will accept dictionary with key as `partition_size` or `n_partitions` (value will be corresponding value obviously). 

Primary reason being that we already have too many arguments and I want to avoid introducing new ones.

fjetter:
We intend to offer a new API soon-ish anyway. I wouldn't want to start deprecating anything at the moment until we know how it will/should look like.
IMHO, wrapping the parameters in a dict doesn't really reduce complexity since the user still needs to pick the parameters.

lr4d:
Re-triggered the build job that failed, it's green now.

Kshitij68:
Thanks @lr4d :).  I'll try to finish up this long-standing PR asap.

kartothek/io/dask/bag.py
Kshitij68: Done!

kartothek/core/docs.py
Kshitij68: I had copied the docstrings from dask.

Do you want me to provide the link to the docstrings in code-base or dask documentation?

lr4d:
closing as stale for now. Feel free to re-open

PR #180, opened 2019-11-19
Introduce type annotations for dask submodule

fjetter:
# Description:

This introduces type hints for the dask.dataframe update module

### DIFF
--- a/kartothek/core/factory.py
+++ b/kartothek/core/factory.py
@@ -81,11 +81,18 @@ class DatasetFactory(DatasetMetadataBase):
         self.is_loaded = False
         self.load_dataset_metadata = load_dataset_metadata
         self.load_all_indices_flag = load_all_indices
+        self._exists = None
 
     def __repr__(self):
         return "<DatasetFactory: uuid={} is_loaded={}>".format(
             self.dataset_uuid, self.is_loaded
         )
+
+    @property
+    def exists(self):
+        if self._exists is None:
+            self._exists = DatasetMetadata.exists(self.dataset_uuid, self.store)
+        return self._exists
 
     @property
     def store(self):
@@ -125,13 +132,18 @@ class DatasetFactory(DatasetMetadataBase):
         # attribute is None, it still falls back to this call
         if name in self._nullable_attributes:
             return object.__getattribute__(self, name)
-        self._instantiate_metadata_cache()
-        ds = getattr(self, "dataset_metadata")
-        return getattr(ds, name)
+        if self.exists:
+            self._instantiate_metadata_cache()
+            ds = getattr(self, "dataset_metadata")
+            return getattr(ds, name)
+        else:
+            raise AttributeError(
+                f"Dataset doesn't yet exist and 'DatasetFactory' object has no attribute '{name}'"
+            )
 
     def __getstate__(self):
         # remove cache
-        return {k: v for k, v in self.__dict__.items() if not k.startswith("_cache_")}
+        return {k: v for k, v in self.__dict__.items() if not k.startswith("_cache")}
 
     def __setstate__(self, state):
         self.__init__(
--- a/kartothek/core/types.py
+++ b/kartothek/core/types.py
@@ -1,0 +1,14 @@
+from typing import Callable
+
+from simplekv import KeyValueStore
+
+from kartothek.serialization import (
+    ConjunctionType,
+    LiteralType,
+    LiteralValue,
+    PredicatesType,
+)
+
+StoreFactory = Callable[[], KeyValueStore]
+
+__all__ = ["ConjunctionType", "LiteralType", "LiteralValue", "PredicatesType"]
--- a/kartothek/io/dask/_update.py
+++ b/kartothek/io/dask/_update.py
@@ -1,32 +1,36 @@
-# -*- coding: utf-8 -*-
-
-
 from functools import partial
+from typing import List, Optional, cast
 
 import dask.array as da
+import dask.dataframe as dd
 import numpy as np
+import pandas as pd
+from dask.delayed import Delayed
 
+from kartothek.core.factory import DatasetFactory
 from kartothek.io_components.metapartition import (
     MetaPartition,
     parse_input_to_metapartition,
 )
 from kartothek.io_components.utils import sort_values_categorical
+from kartothek.serialization import DataFrameSerializer
 
 from ._utils import map_delayed
 
+DelayedList = List[Delayed]
+
 
 def _update_dask_partitions_shuffle(
-    ddf,
-    table,
-    secondary_indices,
-    metadata_version,
-    partition_on,
-    store_factory,
-    df_serializer,
-    dataset_uuid,
-    num_buckets,
-    sort_partitions_by,
-):
+    ddf: dd.DataFrame,
+    table: str,
+    secondary_indices: Optional[List[str]],
+    metadata_version: int,
+    partition_on: List[str],
+    ds_factory: DatasetFactory,
+    df_serializer: Optional[DataFrameSerializer],
+    num_buckets: int,
+    sort_partitions_by: Optional[str],
+) -> da.Array:
     if ddf.npartitions == 0:
         return ddf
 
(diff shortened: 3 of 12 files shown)


kartothek/io/dask/_update.py
```diff
@@ -38,17 +40,18 @@ def _update_dask_partitions_shuffle(
 
     final = []
     for lower, upper in zip(lower_bounds, upper_bounds):
-        chunk_ddf = ddf.partitions[lower : upper + 1]
+        lower_int = cast(int, lower)
```
marco-neumann-jdas: I think this might be technically wrong, since `upper` and `lower` are from a numpy array and might be numpy integers and not python integers. You can safely use `int(lower)` to actually enforce the type. `cast` only changes the mypy type w/o any actual runtime effect (also, it is unchecked and there is no checked cast in python 🤦‍♂ ).

kartothek/io/dask/dataframe.py
```diff
@@ -25,17 +28,17 @@
...
+    dataset_uuid: Optional[str] = None,
+    store: Optional[Callable] = None,
+    table: Optional[str] = None,
+    columns: Optional[List[str]] = None,
+    concat_partitions_on_primary_index: bool = False,
+    predicate_pushdown_to_io: bool = True,
+    categoricals: Optional[List[str]] = None,
+    label_filter: Optional[Callable] = None,
+    dates_as_object: bool = False,
+    predicates: PredicatesType = None,
```
marco-neumann-jdas: I think mypy will complain about this as well: `Optional[...]` is missing for `predicates` and `factory`.

kartothek/io/dask/_update.py
fjetter: This is super weird. If I do `int(lower)` I get
```
No overload variant of "int" matches argument type "_T"
```

kartothek/io/dask/dataframe.py
fjetter: We defined `PredicatesType` as `Optional[List[ConjunctionType]]`

fjetter: Therefore `None` would be a valid predicate. We might iterate this but for starters I would like to simply capture what we have

kartothek/io/dask/_update.py
marco-neumann-jdas: 🤦‍♂ python typing

fjetter:
For the update I needed to fix some internal interfaces which led me to extend the factory for a bit cleaner interface. These changes are also in alignment with what is proposed in #149

kartothek/core/types.py
```diff
@@ -0,0 +1,14 @@
...
+from kartothek.serialization import (
+    ConjunctionType,
+    LiteralType,
+    LiteralValue,
+    PredicatesType,
+)
+
+StoreFactory = Callable[[], KeyValueStore]
+
+__all__ = ["ConjunctionType", "LiteralType", "LiteralValue", "PredicatesType"]
```
fjetter: Not sure how we should deal with this. The store factory is something I encountered very frequently and the `PredicatesType` as well. Maybe it would make sense to define a type here for every kwarg we expose?

kartothek/io/dask/_update.py
fjetter: The int function is apparently defined as 

```
def int(self, x: Union[str, bytes, SupportsInt] = ...) -> int
```

Therefore, the Numpy types would need to use `SupportsInt`

still...

fjetter:
Upon request I'll split the factory changes from the type annotations

kartothek/core/types.py
marco-neumann-jdas: I am ok with the common complex types, but I think we should avoid the indirection if it is not really needed (like `Optional[str]` is not too complex IMHO).

kartothek/io/dask/delayed.py
```diff
@@ -97,8 +107,11 @@ def delete_dataset__delayed(dataset_uuid=None, store=None, factory=None):
 
 @default_docs
 def garbage_collect_dataset__delayed(
-    dataset_uuid=None, store=None, chunk_size=100, factory=None
-):
+    dataset_uuid: Optional[str] = None,
+    store: Optional[StoreFactory] = None,
+    chunk_size=100,
```
marco-neumann-jdas:
```suggestion
    chunk_size: int = 100,
```

kartothek/io_components/utils.py
```diff
@@ -111,7 +112,7 @@ def _instantiate_store(store):
 
 
 def _ensure_compatible_indices(dataset, secondary_indices):
-    if dataset:
+    if dataset.exists:
```
marco-neumann-jdas: Is there a test of this?

kartothek/io_components/utils.py
```diff
@@ -341,30 +340,34 @@ def sort_values_categorical(df, column):
...
         The dataset to be validated
     expected_table: Optional[str]
         Ensure that the table in the dataset is the same as the given one.
     """
-
-    if len(dataset.tables) > 1:
-        raise TypeError(
-            "Expected single table dataset but found dataset with tables: `{}`".format(
-                dataset.tables
+    if (isinstance(dataset, DatasetFactory) and dataset.exists) or isinstance(
```
marco-neumann-jdas: This additional guard wasn't there before. Is there a test for this behavior?

kartothek/io_components/write.py
```diff
@@ -55,7 +55,7 @@ def persist_common_metadata(partition_list, update_dataset, store, dataset_uuid)
         for tab, tm in mp.table_meta.items():
             tm_dct[tab].add(tm)
 
-    if update_dataset:
+    if update_dataset and update_dataset.exists:
```
marco-neumann-jdas: test?

kartothek/io_components/write.py
```diff
@@ -106,7 +106,7 @@ def store_dataset_from_partitions(
 ):
     store = _instantiate_store(store)
 
-    if update_dataset:
+    if update_dataset and update_dataset.exists:
```
marco-neumann-jdas: test?

kartothek/core/factory.py
```diff
@@ -99,20 +106,23 @@ def _instantiate_metadata_cache(self):
...
                 self._cache_metadata = self._ds_callable()
             else:
-                self._cache_metadata = DatasetMetadata.load_from_store(
-                    uuid=self.dataset_uuid,
-                    store=self.store,
-                    load_schema=self.load_schema,
-                    load_all_indices=self.load_all_indices_flag,
-                )
+                self._exists = DatasetMetadata.exists(self.dataset_uuid, self.store)
+                if self._exists:
```
marco-neumann-jdas: What happens w/o this existence check?

kartothek/core/factory.py
```diff
@@ -99,20 +106,23 @@ def _instantiate_metadata_cache(self):
...
+                    )
             if not self.load_dataset_metadata:
                 self._cache_metadata.metadata = {}
         self.is_loaded = True
         return self
 
     @property
     def dataset_metadata(self):
-        self._instantiate_metadata_cache()
+        if self._exists is not False:
```
marco-neumann-jdas: What happens w/o this existence check?

kartothek/io_components/utils.py
fjetter: This is used in all writing functions where we migrated to the factory model by using `_ensure_factory`. 
Explicit unit test doesn't exist. Hell breaks loose if I change this, though. 

I would consider it _tested_ but I can add more tests if you think it is necessary. If so, do you think a unit test is sufficient or something larger?

kartothek/io_components/utils.py
fjetter: The function `check_single_table_dataset` didn't have any unit tests. I added some

kartothek/core/factory.py
fjetter: Simplified it. There is no need to check for existence here.

kartothek/core/factory.py
fjetter: I removed all unnecessary existents checks in the factory. Behavior should be now easier to understand, I hope.

kartothek/io_components/write.py
fjetter: Again no dedicated unit test but its used all over the place. What test do you have in mind?

kartothek/io_components/utils.py
marco-neumann-jdas: So why did this work before your PR then?

lr4d:
closing as stale

PR #188, opened 2019-12-09
Improve predicate TypeError messages

jonashaag:
Improve error messages when passing predicate values of wrong type.

Example old:

```
TypeError: Unexpected type encountered. Expected O but got M.
```

Example new:

```
TypeError: Unexpected type for predicate: Column 'col1' has pandas type 'object',
but predicate value Timestamp('2020-01-01 00:00:00') has pandas type 'datetime64[ns]'
(Python type '<class 'pandas._libs.tslibs.timestamps.Timestamp'>').
```

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -258,6 +258,7 @@ def _normalize_predicates(parquet_file, predicates, for_pushdown):
             col, op, val = literal
             col_idx = parquet_file.reader.column_name_idx(col)
             pa_type = schema[col_idx].type
+            column_name = schema[col_idx].name
 
             if pa.types.is_null(pa_type):
                 # early exit, the entire conjunction evaluates to False
@@ -265,7 +266,10 @@ def _normalize_predicates(parquet_file, predicates, for_pushdown):
                 break
 
             if op == "in":
-                values = [_normalize_value(l, pa_type) for l in literal[2]]
+                values = [
+                    _normalize_value(l, pa_type, column_name=column_name)
+                    for l in literal[2]
+                ]
                 if for_pushdown and values:
                     normalized_value = [
                         _timelike_to_arrow_encoding(value, pa_type) for value in values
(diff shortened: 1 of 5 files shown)


kartothek/serialization/_parquet.py
```diff
@@ -347,14 +348,20 @@ def _normalize_value(value, pa_type):
         elif isinstance(value, bytes):
             value = value.decode("utf-8")
             return datetime.datetime.strptime(value, "%Y-%m-%d").date()
-        elif isinstance(value, datetime.date) and not isinstance(
-            value, datetime.datetime
-        ):
-            return value
+        elif isinstance(value, datetime.date):
+            if isinstance(value, datetime.datetime):
+                raise TypeError(
```
jonashaag: Special case for matching datetime predicate values against date columns. Generic error message below does not handle this nicely

fjetter:
@marco-neumann-jdas So far we've been very strict regarding typing. I'm wondering if we should start relaxing this and start casting for certain types. E.g. if the date time doesn't have timezone information and can be converted lossless to a date this would be fine. The other way round may be even easier. Thoughts?

fjetter:
The only thing that bothers me here is that we pass the column name deep down to a function which doesn't even know about the concept of a column.
Would it be possible to define a function which checks the validity top-level, similar to `check_predicates`? The signature would be something like `compat_predicates_schema(predicates: PredicatesType, schema: SchemaWrapper)`

fjetter:
Failing CI due to black/flake8

marco-neumann-jdas:
@fjetter let's not mix the general casting rules discussion with the valid error message improvements proposed here.

jonashaag:
> The only thing that bothers me here is that we pass the column name deep down to a function which doesn't even know about the concept of a column.

Agreed, it's not nice. I also tried using `try...catch` with a custom exception type that is thrown at the lower levels without column info, and is re-thrown at higher levels with column info added. The `column_name` solution is much more pragmatic than that approach though (if you can't related -- try to code it up and you will see :-D)

jonashaag:
It's a pattern similar to what is used here: https://docs.python.org/2/c-api/exceptions.html#c.Py_EnterRecursiveCall (not that the CPython code base is a role model for nice code but maybe for pragmatic code ;-))

Fixed black.

jonashaag:
Fixed test.

PR #191, opened 2019-12-16
Remove _KTK_HASH_BUCKET if exists

fjetter:
# Description:

The update function would not remove the internal column

- [ ] Closes #xxxx
- [x] Changelog entry

### DIFF
--- a/tests/io/dask/dataframe/test_update.py
+++ b/tests/io/dask/dataframe/test_update.py
@@ -171,6 +171,12 @@ def test_update_shuffle_buckets(
         range(unique_secondaries)
     )
 
+    assert set(dataset.table_meta["core"].names) == {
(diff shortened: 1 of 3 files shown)


tests/io/dask/dataframe/test_update.py
```diff
@@ -171,6 +171,13 @@ def test_update_shuffle_buckets(
         range(unique_secondaries)
     )
 
+    assert set(dataset.table_meta["core"].names) == {
+        "primary",
+        "secondary",
+        "sorted_column",
+    }
+    assert _KTK_HASH_BUCKET not in dataset.table_meta["core"].names
```
lr4d: This is redundant given the previous statement but it's nice to have it explicit

PR #197, opened 2020-01-07
Nice error message if partition_on and secondary_indices overlap

jonashaag:
Before this change, user would get error like "Column X missing from dataframe" (because X has been made a partition key already and removed).

### DIFF
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -21,7 +21,7 @@ from kartothek.io_components.metapartition import (
     parse_input_to_metapartition,
 )
 from kartothek.io_components.read import dispatch_metapartitions_from_factory
-from kartothek.io_components.utils import normalize_args
+from kartothek.io_components.utils import normalize_args, raise_if_indices_overlap
 from kartothek.io_components.write import (
     raise_if_dataset_exists,
     store_dataset_from_partitions,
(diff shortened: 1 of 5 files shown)


kartothek/io/dask/bag.py
```diff
@@ -207,6 +207,12 @@ def store_bag_as_dataset(
     if not overwrite:
         raise_if_dataset_exists(dataset_uuid=dataset_uuid, store=store)
 
+    partition_secondary_overlap = set(partition_on) & set(secondary_indices)
```
xhochy: Can you turn this into a generic check instead of duplicating it three times?

```python
def check_partition_secondary_overlap(partition_on: Iterable[str], secondary_indices: Iterable[str]) -> None:
    __tracebackhide__ = True

    partition_secondary_overlap = set(partition_on) & set(secondary_indices)
    if partition_secondary_overlap:
        raise RuntimeError(
            f"Cannot create secondary index on partition columns: {partition_secondary_overlap}"
        )
```

jonashaag: Unfortunately rest of code is duplicated as well and I’m not sure where to put all that stuff

fjetter: While that's not a _good_ place, the `io_components.utils` module is our bin for collecting this stuff

jonashaag: OK, will move it there.

PR #199, opened 2020-01-10
Raise an error in `dataset.query` when column queried is not an index column

### LINKED ISSUE #198: unexpected deletion of all partitions of a dataset without warning
### Problem description

1. Kartothek deletes all partitions of a dataset when a non-existing column is provided in the `delete_scope`.
2. Kartothek deletes all partitions of a dataset (without warning) when a `delete_scope` is provided for a column which does not have indices.

```python
from kartothek.io.iter import store_dataframes_as_dataset__iter
from kartothek.io.eager import update_dataset_from_dataframes, delete_dataset
import pandas as pd
import storefact
from functools import partial

# Case 1: delete_scope specifies column which does not exist
df = pd.DataFrame({"A": [0] * 12})

store_factory = partial(
    storefact.get_store_from_url, f"hfs://{TemporaryDirectory().name}"
)
dataset_uuid = "uuid"

dm = store_dataframes_as_dataset__iter(
    [df],
    store=store_factory,
    dataset_uuid=dataset_uuid,
)
print(dm.partitions)

dm = update_dataset_from_dataframes(
    [], store=store_factory, dataset_uuid=dataset_uuid, delete_scope=[{"UNKNOWN_COLUMN": 6}]
)
assert dm.partitions == {}

# Case 2: no partitioning, no secondary indices: delete_scope deletes all partitions w/o warning or checking if
# any of the values provided actually appear in Parquet metadata / data
df = pd.DataFrame({"A": [0] * 12})

store_factory = partial(
    storefact.get_store_from_url, f"hfs://{TemporaryDirectory().name}"
)
dataset_uuid = "uuid"

dm = store_dataframes_as_dataset__iter(
    [df],
    store=store_factory,
    dataset_uuid=dataset_uuid,
)
print(dm.partitions)

dm = update_dataset_from_dataframes(
    [], store=store_factory, dataset_uuid=dataset_uuid, delete_scope=[{"A": -1000}]
)
assert dm.partitions == {}
```

### Used versions
master branch

### Related issues
https://github.com/JDASoftwareGroup/kartothek/issues/196

lr4d:
This fixes https://github.com/JDASoftwareGroup/kartothek/issues/198

- [ ] Changelog entry

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,7 +29,7 @@ Improvements
 Bug fixes
 ^^^^^^^^^
 * Fix an issue where :func:`~kartothek.io.dask.dataframe.update_dataset_from_ddf` would create a column named "_KTK_HASH_BUCKET" in the dataset
-
+* Fix https://github.com/JDASoftwareGroup/kartothek/issues/198
 
 Version 3.6.1 (2019-12-11)
 ==========================
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -311,6 +311,20 @@ class DatasetMetadataBase(CopyMixin):
         for column, value in kwargs.items():
             if column in combined_indices:
                 candidate_set &= set(combined_indices[column].query(value))
+            else:
+                if (
+                    # FIXME: `self.tables[0]`. What if column exists in another table?
+                    not self.table_meta[self.tables[0]]
+                    .internal()
+                    .field_by_name(column)
+                ):
+                    raise ValueError(
+                        f"Column provided (`{column}`) does not appear in dataset"
+                    )
+                # Rather than returning all partitions, we abort the execution at this stage to be safe
+                raise ValueError(
+                    f"Column provided (`{column}`) does not appear in dataset indices"
+                )
 
         return list(candidate_set)
 
--- a/kartothek/io/testing/update.py
+++ b/kartothek/io/testing/update.py
@@ -159,9 +159,8 @@ def test_update_dataset_with_partitions__reducer_delete_only(
     )
     dataset = dataset.load_index("p", store)
 
-    empty_part = []
     dataset_updated = bound_update_dataset(
-        [empty_part],
+        [],
         store=store_factory,
         dataset_uuid="dataset_uuid",
         delete_scope=[{"p": 1}],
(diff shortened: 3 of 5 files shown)


CHANGES.rst
```diff
@@ -29,7 +29,7 @@ Improvements
 Bug fixes
 ^^^^^^^^^
 * Fix an issue where :func:`~kartothek.io.dask.dataframe.update_dataset_from_ddf` would create a column named "_KTK_HASH_BUCKET" in the dataset
-
+* Fix https://github.com/JDASoftwareGroup/kartothek/issues/198
```
fjetter: GitHub repos usually use the convention to mark issues as `GH###`

kartothek/core/dataset.py
```diff
@@ -311,6 +311,20 @@ def query(self, indices: List[IndexBase] = None, **kwargs) -> List[str]:
         for column, value in kwargs.items():
             if column in combined_indices:
                 candidate_set &= set(combined_indices[column].query(value))
+            else:
+                if (
+                    # FIXME: `self.tables[0]`. What if column exists in another table?
+                    not self.table_meta[self.tables[0]]
```
fjetter: As long as we support multi table datasets we can not make this assumption. You'd need to check all tables

kartothek/core/dataset.py
```diff
@@ -311,6 +311,20 @@ def query(self, indices: List[IndexBase] = None, **kwargs) -> List[str]:
...
+                    .internal()
+                    .field_by_name(column)
+                ):
+                    raise ValueError(
+                        f"Column provided (`{column}`) does not appear in dataset"
+                    )
+                # Rather than returning all partitions, we abort the execution at this stage to be safe
+                raise ValueError(
```
fjetter: _Technically_ this is a breaking change in behaviour :/
I do agree that this is a sane behaviour, though.

fjetter: I would actually propose to add a _new_ function / functionality to query by predicates (essentially call dispatch_metapartitions but only return the partition names). This would amount to the same functionality and would allow to deal with this in a non-breaking way, especially since I would like to get rid of the query-by-dict syntax in general

fjetter (changes requested):

PR #208, opened 2020-01-28
Decrease parquet version to 1.0

fjetter:
@xhochy I was just browsing a few arrow tickets I opened ages ago and stumbled upon a comment claiming that one shouldn't use parquet 2.0 in production. Is this still the case?

https://issues.apache.org/jira/browse/PARQUET-458?focusedCommentId=16931914&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16931914
https://issues.apache.org/jira/browse/ARROW-5089

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -68,7 +68,7 @@ def _reset_dictionary_columns(table):
 
 
 class ParquetSerializer(DataFrameSerializer):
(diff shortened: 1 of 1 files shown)


xhochy:
This is mainly related to the new `DataPageV2` which is not used by the code path as they are not implemented in Arrow. The version "2.0" flag here only adds support for the unsigned integer types.

PR #210, opened 2020-01-31
Revert rest_dictionary_table

fjetter:
# Description:

Statistics are completely messed up. This introduces a trade off between performance when serialising (both memory and CPU time) vs speed when querying. Assuming that retrieval speed is important when using a chunk size this trades off in favour of the statistics

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@ Next release
 Improvements
 ^^^^^^^^^^^^
 
-* Performance improvements for categorical data when using pyarrow>=0.15.0
+* Cast categorical columns to plain data type for all arrow types, see https://issues.apache.org/jira/browse/ARROW-7732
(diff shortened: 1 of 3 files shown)


fjetter:
Actually, I'm on the fence here after thinking a bit about it. After all, the predicate pushdown statistics can only work for a few columns but the categorical would help for large data frames. Work around would be to not use categorical for columns that matter.
I incline to close this PR again

marco-neumann-jdas:
So is that a generic issue with parquet statistics then? If so, is there an upstream ticket discussion this?

fjetter:
Yes, it's an issue with parquet statistics. We didn't encounter this before because for arrow<0.15.0 we cast categoricals back to plain Dtype, upstream issue https://issues.apache.org/jira/browse/ARROW-7732

PR #211, opened 2020-02-03
Pin pandas 1.0.0 due to an upstream bug

fjetter:
# Description:

Pandas 1.0.1 will be out in a few days. This currently breaks important functionality (nightlies are already green-ish)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
 env:
   global:
     - KTK_TESTS=1
+    - KTK_DOCS=0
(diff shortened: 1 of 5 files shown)


fjetter:
Had to fix a few deprecations along the way. In particular it is quite tricky because our docs raise if they encounter deprecation warnings. I created a dedicated doc build job to handle this

lr4d:
Please check that all build jobs are running as expected when changing the travis file next time, this has introduced 3 new issues with the build. (I realized something was seriously off when I saw the arrow nightlies passing)

PR #212, opened 2020-02-03
fix GITHUB_CLIENT tokens

marco-neumann-jdas:


florian-jetter-jdas:
Jobs pass the previous keyerror. Tests themselves will be fixed by #211

PR #215, opened 2020-02-05
Run tests for arrow 16

fjetter:
# Description:

I suggest to soon-ish drop support for 0.13.0 and 0.14.1 to not blow up the test matrix. In the same step we can also add py3.8.
For now, I'd like to keep as is and do at least one more release before we cut

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,11 @@ env:
     - CONDA_CREATE_ARGS="pyarrow==0.13.0"
     - CONDA_CREATE_ARGS="pyarrow==0.14.1"
     - CONDA_CREATE_ARGS="pyarrow==0.15.0"
+    - CONDA_CREATE_ARGS="pyarrow==0.16.0"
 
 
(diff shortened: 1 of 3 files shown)


fjetter:
Not yet available on Conda...

lr4d:
> Not yet available on Conda...

It's available now. I've restarted the build, though I'd suggest we first merge #221 and rebase to verify that the os x job also runs smoothly

lr4d:
The doc jobs is failing, looks like some issue with the mypy pre-commit hook. This job is also not using the correct python version (the env variable `TRAVIS_PYTHON_VERSION` is unset, I presume because the environment is `sh`)

lr4d:
> The doc jobs is failing, looks like some issue with the mypy pre-commit hook. This job is also not using the correct python version (the env variable `TRAVIS_PYTHON_VERSION` is unset, I presume because the environment is `sh`)

should be fixed by #223

PR #222, opened 2020-02-10
Use proper generator for map_delayed

fjetter:
# Description:

This explicit loop and list creation is pretty slow. See below measurements for a dataset with about 150k partitions, especially when calling this many times this becomes quite an overhead.

The changes in delete are due to partial not properly working if one of the kwargs is a delayed object. Since this is only private the change should be fine

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Next release
 Improvements
 ^^^^^^^^^^^^
 
+* Decrease scheduling overhead for dask based pipelines
 * Performance improvements for categorical data when using pyarrow>=0.15.0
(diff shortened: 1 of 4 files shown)


lr4d:
What's the intention of the casting to list?

lr4d:
> What's the intention of the casting to list?

To evaluate the `map` object I presume. Why not have it inside `map_delayed` in the first place?

fjetter:
> What's the intention of the casting to list?
The read interface is still list, not a generator. Can revisit this with the next major release

> Why not have it inside map_delayed in the first place?
Internally we sometimes concatenate the map_delayed calls which may become quite costly

PR #224, opened 2020-02-11
fix lack of testing in build jobs

lr4d:
So our code hasn't been properly tested for the last few PR's as `1 == true` does not evaluate to true in bash (surprise).


fjetter:
🙈

PR #226, opened 2020-02-13
Reconstruct dask index during loading

fjetter:
# Description:

This is a feature I wanted to build for quite some time now and with the `dispatch_by` logic it is, in fact, rather simple

With the `dask_index_on` kwarg it is possible to set the Dask index on a given index column directly which allows for very efficient joins on this column.

### DIFF
--- a/kartothek/core/index.py
+++ b/kartothek/core/index.py
@@ -132,6 +132,16 @@ class IndexBase(CopyMixin):
             class_=type(self).__name__, attrs=", ".join(repr_str)
         )
 
+    def observed_values(self) -> np.array:
+        """
+        Return an array of all observed values
+        """
+        return np.fromiter(
+            (self.normalize_value(self.dtype, x) for x in self.index_dct.keys()),
+            count=len(self.index_dct),
+            dtype=self.dtype.to_pandas_dtype(),
+        )
+
     @staticmethod
     def normalize_value(dtype: pa.DataType, value: Any) -> Any:
         """
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -36,6 +36,7 @@ def read_dataset_as_ddf(
     dates_as_object=False,
     predicates=None,
     factory=None,
+    dask_index_on=None,
 ):
     """
     Retrieve a single table from a dataset as partition-individual :class:`~dask.dataframe.DataFrame` instance.
@@ -46,6 +47,10 @@ def read_dataset_as_ddf(
 
     Parameters
     ----------
+    dask_index_on: str
+        Reconstruct (and set) a dask index on the provided index column.
+
+        For details on performance, see also `dispatch_by`
     """
     ds_factory = _ensure_factory(
         dataset_uuid=dataset_uuid,
@@ -73,9 +78,18 @@ def read_dataset_as_ddf(
         label_filter=label_filter,
         dates_as_object=dates_as_object,
         predicates=predicates,
+        dispatch_by=dask_index_on,
     )
-
-    return dd.from_delayed(delayed_partitions, meta=meta)
+    if dask_index_on:
+        divisions = ds_factory.indices[dask_index_on].observed_values()
+        divisions.sort()
+        divisions = list(divisions)
+        divisions.append(divisions[-1])
+        return dd.from_delayed(
+            delayed_partitions, meta=meta, divisions=divisions
+        ).set_index(dask_index_on, divisions=divisions)
+    else:
+        return dd.from_delayed(delayed_partitions, meta=meta)
 
 
 def _get_dask_meta_for_dataset(
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,6 +1,3 @@
-# -*- coding: utf-8 -*-
-
-
 import dask
 import dask.dataframe as dd
 import numpy as np
@@ -15,6 +12,7 @@ from kartothek.io_components.utils import (
     _ensure_compatible_indices,
     check_single_table_dataset,
     normalize_arg,
+    normalize_args,
     validate_partition_keys,
 )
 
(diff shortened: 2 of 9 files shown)


fjetter:
Tests fail due to improper docs formatting. Will fix this, shouldn't stop reviews

lr4d:
I get that it allows for efficient joins, but do we want to force the number of partitions to always be equal to the number of unique values for the index column when using this kwarg?

kartothek/core/index.py
```diff
@@ -132,6 +132,16 @@ def __repr__(self) -> str:
             class_=type(self).__name__, attrs=", ".join(repr_str)
         )
 
+    def observed_values(self) -> np.array:
+        """
+        Return an array of all observed values
+        """
+        return np.fromiter(
+            (self.normalize_value(self.dtype, x) for x in self.index_dct.keys()),
+            count=len(self.index_dct),
```
lr4d: redundant?

kartothek/io_components/read.py
```diff
@@ -75,7 +75,7 @@ def dispatch_metapartitions_from_factory(
 
         # Group the resulting MetaParitions by partition keys or a subset of those keys
         merged_partitions = base_df.groupby(
-            by=list(dispatch_by), sort=False, as_index=False
+            by=list(dispatch_by), sort=True, as_index=False
```
lr4d: when is this used? I tried reverting to `sort=False` and the test below was still passing

lr4d: Also note, `dispatch_by` is a string here

fjetter:
> but do we want to force the number of partitions to always be equal to the number of unique values for the index column when using this kwarg?

We have certain constraints regarding the dask internal implementation. In particular, we are not allowed to have duplicate division entries (except the last ones).

Compressing multiple ktk partitions into one dask partition could be a future improvement (Difficulty here is to decide which to merge, if not trivially done)

kartothek/io_components/read.py
fjetter: > when is this used? I tried reverting to sort=False and the test below was still passing

Should add a test case then. I rely on having the partitions sorted when setting the divisions.

- [ ] Add test case for sorted/unsorted partitions

fjetter: > Also note, dispatch_by is a string here

Should be normalised

- [ ] Add normalisation for dask_index_on

kartothek/core/index.py
fjetter: What exactly? The fromiter or the count? I used the fromiter to somehow reduce memory footprint, the count is an optional kwarg which improves performance since Numpy can allocate all memory up front

lr4d: I meant the count, but never mind then

lr4d:
> Compressing multiple ktk partitions into one dask partition could be a future improvement (Difficulty here is to decide which to merge, if not trivially done)

Right, that was what I was going for. My rationale was that, if we have an index with a very large number of unique values (one could argue this shouldn't happen in the first place), then we will get tiny dask partitions, which will affect performance.

One could have a check where we look at if the number of unique values is larger than _n_, and if it is, trivially merge these by setting the divisions as steps of the unique values (e.g. every 10'th value), to then have a number of partitions equal to some constant

fjetter:
> One could have a check where we look at if the number of unique values is larger than n, and if it is, trivially merge these by setting the divisions as steps of the unique values

What if it is a huge dataset and the unique values actually make sense? What would be `n`? I wouldn't want to implement some kind of heuristic here since no matter what we implement, we will hit weird edge cases.

In fact, the heavy fragmentation is something dask can deal with (think `repartition`). The issue would rather be that if we use an improper column, we rely heavily on predicate pushdown which makes the whole thing slower than actually shuffling the data.

Therefore, the only merge I would consider actually worth investigating is to merge partitions with consecutive values originating from the same file. Maybe this could be implemented as a dask optimization actually... either way it will require a bit of work and I consider this out of scope for this PR

The heavy fragmentation of the Dask Dataframe is something I'm willing to accept with the appropriate warning. After all, this is the same for the `dispatch_by` functionality. If used with a partition column (or a column from which the partition column is derived from) this is still very powerful

kartothek/io/dask/dataframe.py
```diff
@@ -73,9 +78,18 @@ def read_dataset_as_ddf(
...
     )
-
-    return dd.from_delayed(delayed_partitions, meta=meta)
+    if dask_index_on:
+        divisions = ds_factory.indices[dask_index_on].observed_values()
+        divisions.sort()
+        divisions = list(divisions)
+        divisions.append(divisions[-1])
+        return dd.from_delayed(
+            delayed_partitions, meta=meta, divisions=divisions
```
lr4d: `divisions` shouldn't be specified here should it?

fjetter: Technically, at the moment, it doesn't do anything but I don't know if this will change

kartothek/io/dask/dataframe.py
```diff
@@ -73,9 +78,18 @@ def read_dataset_as_ddf(
...
+        ).set_index(dask_index_on, divisions=divisions)
```
fjetter: I believe I will need to add the sorted kwarg here to avoid a shuffle

- [ ] Add sorted kwarg and try to figure out if we can test for shuffle

kartothek/io/dask/dataframe.py
lr4d: please remove it

lr4d: @fjetter this is logicaly incorrect, no?

fjetter: no, it isn't, why would it be?

lr4d: Because at that point (i.e. before setting any index) the `divisions` would be referencing the default pandas integer/range index, not any custom index column. However, the variable `divisions` created references the divisions for the custom index column, which is not yet set on the dataframe (set in the next line, where it does make sense).

lr4d: Although the code runs fine, it shouldn't be there AFAI see

Issue #235, opened 2020-03-03
Let merge_datasets_as_delayed merge >2 datasets and filter by predicates


mlondschien:
It would be nice to be able to supply `kartothek.io.dask.delayed.merge_datasets_as_delayed` with a list of `dataset_uuids` to merge an arbitrary number of datasets. 

This could be implemented by 

 - defining a new `kartothek.dask.delayed.concat_datasets_as_delayed` or similar which takes a list `dataset_uuids` as an argument or
 - allow  `kartothek.dask.delayed.merge_datasets_as_delayed` to be supplied with a list for `left`

to not break existing usages of the function. 

The `match_how="left"` would need to be replaced by `match_how="first"`. I am not sure how to translate `match_how="prefix"`. What is a typical use case here? Additionally it would be nice to supply a `merge_func` via `merge_tasks` that takes more than two dataframes as input. This would require a similar change as above, either defining a `MetaPartition.concat_dataframes` or similar method or allowing `MetaPartition.merge_dataframes` to be supplied with a list for `left`.

Questions concerning the current implementation of `merge_datasets_as_delayed`:
 - For merging with `match_how="exact"`, the keys of the partitions are compared to match partitions. However, since the keys include the name of the stored parquet files, they will never match. What is the idea here?
 - It is annoying to have to supply the names of the tables in merge_tasks. Why not supply the `merge_func` with all available dataframes if no labels (currently `left` and `right`) are supplied?

Additionally it would be nice to be able to supply `merge_datasets_as_delayed` with a `predicates` argument that filters by partitions before merging and then uses `filter_df_from_predicates` after merging.

fjetter:
Firstly, thanks for the interest and sorry for the delayed response.

Secondly, the entire currently existing alignment logic is about as old as the library and was never really refactored. I'm open to breaking some eggs in this situation (We have a few other things which should be addressed in terms of UX so friendly, breaking release would be appropriate in the near future, anyhow)

> he keys of the partitions are compared to match partitions. However, since the keys include the name of the stored parquet files, they will never match. What is the idea here?

As I said, this implementation is dated back a long time. In the very first iterations we used hard coded file names (e.g. partition_0, partition_1) which were provided by the user application. In this context this makes a lot of sense and allows for the easiest alignment, obviously.
I consider providing these labels as a bad practice and want to remove this in the next breaking release so we can safely consider this merge option as deprecated.

> It is annoying to have to supply the names of the tables in merge_tasks

[...]

PR #239, opened 2020-03-10
Let merge_datasets_as_delayed merge >2 datasets, filter by predicates and subset column

mlondschien:
# Description:

This Draft PR tries to implement changes as suggested in #235. The current philosophy is to
 * Implement an additional function (e.g. `merge_many_datasets_as_delayed`) not to break the existing API
 * rewrite `align_datasets` to efficiently cope with multiple datasets
 * Implement an additional `merge_many_dataframes` or similar to the `MetaPartition` class to allow for merging of dataframes.

The new `merge_many_datasets_as_delayed` can be supplied with `predicates` to subset rows and `columns` to subset columns. To account for the case where a column is defined in multiple datasets, `columns` needs to be a dictionary of type `{uuid: table: [columns]}`.

### DIFF
--- a/kartothek/io/dask/delayed.py
+++ b/kartothek/io/dask/delayed.py
@@ -39,6 +39,7 @@ from kartothek.io_components.write import (
     store_dataset_from_partitions,
 )
 
+from ._generic import check_predicates
 from ._update import _update_dask_partitions_one_to_one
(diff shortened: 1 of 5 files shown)


mlondschien:
Closing this as its superseded by #243 .

PR #240, opened 2020-03-10
fix arrow nightly builds

### LINKED ISSUE #237: Nightlies fail with arrow integration
The current nightlies fail due to a new arrow version.

e.g. https://travis-ci.com/JDASoftwareGroup/kartothek/jobs/296114927

fjetter:
# Description:

This closes the nightly build failures. The issue is that a parquet schema roundtrip adds additional metadata to the fields. This can currently cause problems, especially for nested data types.

We can wait with merging until see https://github.com/apache/arrow/pull/6569 is fixed/resolved. Maybe the nested data types are already resolved there. Then the fixes here would boil down to the fix in `_diff_schemas`

- [x] Closes #237
- [x] Changelog entry
- [x] Wait until https://github.com/apache/arrow/pull/6569 is resolved

### DIFF
--- a/kartothek/core/common_metadata.py
+++ b/kartothek/core/common_metadata.py
@@ -58,7 +58,7 @@ class SchemaWrapper:
     def _schema_compat(self):
         # https://issues.apache.org/jira/browse/ARROW-5104
         schema = self.__schema
-        if self.__schema is not None and self.__schema.metadata is not None:
+        if self.__schema is not None and self.__schema.pandas_metadata is not None:
             pandas_metadata = schema.pandas_metadata
             index_cols = pandas_metadata["index_columns"]
             if len(index_cols) > 1:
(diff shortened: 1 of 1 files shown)


kartothek/core/common_metadata.py
```diff
@@ -547,8 +547,10 @@ def _remove_diff_header(diff):
 
 def _diff_schemas(first, second):
     # see https://issues.apache.org/jira/browse/ARROW-4176
-    first_pyarrow_info, _ = str(first).split("metadata\n--------")
-    second_pyarrow_info, _ = str(second).split("metadata\n--------")
+
+    # Pyarrow >0.16.0 do not include the metadata anymore
+    first_pyarrow_info = str(first).split("metadata\n--------")[0]
+    second_pyarrow_info = str(second).split("metadata\n--------")[0]
```
wesm: This isn't set in stone yet.

fjetter: The code here should at least not raise anymore. 
I could offer https://github.com/JDASoftwareGroup/kartothek/pull/247 to not break arrow integration due to minor changes like these format changes

wesm: Yeah, I question whether a str/repr should be regarded as stable from commit to commit. This output is designed for humans, not machines, at least that is how it has been developed so far. Would it be better to check only the data contained in the schema (field names, types, metadata keys/values)?

fjetter: We only process this diff because the output with metadata was way to verbose and the diff was essentially useless. This function strips the metadata and offers a better human readable diff. I now realize that we should probably just diff `schema.remove_metadata()` which should achieve the same but should be much more stable.

I'll take another look at #247 because I do agree that we shouldn't commit to the repr as a public API.

That is/was the intention of https://issues.apache.org/jira/browse/ARROW-4176 (I'll link this comment for more context)

PR #243, opened 2020-03-11
Add merge_many_datasets_as_delayed

### LINKED ISSUE #235: Let merge_datasets_as_delayed merge >2 datasets and filter by predicates

xhochy:
This implements a merge that works on multiple datasets. For the moment, I have kept the code separate from the existing merge. Aligning partitions is done depending on `match_how` as in some cases, we can speed it up greatly by using e.g. the indices dataframes.

Fixes #235 

Missing:

- [ ] predicate support
- [ ] `exact` matching 
- [ ] merge with existing code


kartothek/io_components/merge.py
```diff
@@ -1,8 +1,18 @@
...
 
 from kartothek.core.dataset import DatasetMetadata
+from kartothek.core.factory import DatasetFactory
 from kartothek.io_components.metapartition import MetaPartition
-from kartothek.io_components.utils import _instantiate_store
+from kartothek.io_components.read import dispatch_metapartitions_from_factory
+from kartothek.io_components.utils import _instantiate_store, _make_callable
+
+if TYPE_CHECKING:
+    from simplekv import KeyValueStore
```
lr4d:
```suggestion
    from simplekv import KeyValueStore  # noqa: F401
```

linting fails: `kartothek/io_components/merge.py:14:5: F401 'simplekv.KeyValueStore' imported but unused`

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+    columns: Optional[List[Dict[str, List[str]]]] = None,
+):
+    """
+    A dask.delayed graph to perform the merge of two full kartothek datasets.
+
+    Parameters
+    ----------
+    dataset_uuids : List[str]
+    match_how : Union[str, Callable]
+        Define the partition label matching scheme.
```
marco-neumann-jdas: Why is the whole thing label-based and not index-based?

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+        Available implementations are:
+
+        * first : The partitions of the first dataset are considered to be the base
```
marco-neumann-jdas: related to the question above: do we really need different string-based join modes?

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+                  partitions and **all** partitions of the remaining datasets are
+                  joined to the partitions of the first dataset. This should only be
+                  used if all but the first dataset contain very few partitions.
```
marco-neumann-jdas: What does "few" mean? What happens if this is not the case? Please give the user more guidance and try to provide a more failure-proof API.

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+        If True, an exact match of partition labels between the to-be-merged
+        datasets is required in order to merge.
+        If False (Default), the partition labels of the dataset with fewer
+        partitions are interpreted as prefixes.
+    merge_tasks : List[Dict]
+        A list of merge tasks. Each item in this list is a dictionary giving
+        explicit instructions for a specific merge.
+        Each dict should contain key/values:
+
+        * 'output_label' : The table for the merged dataframe
```
marco-neumann-jdas: what about the `tables` key from the example below?

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+        * `merge_func`: A callable with signature
+                        `merge_func(dfs, merge_kwargs)` to
+                        handle the data preprocessing and merging.
+        * 'merge_kwargs' : The kwargs to be passed to the `merge_func`
```
marco-neumann-jdas: not required, use a partial instead.

kartothek/io/dask/delayed.py
```diff
@@ -230,6 +283,91 @@ def merge_datasets_as_delayed(
...
+        * exact : All partition labels of each dataset need to be an exact match.
+        * callable : A callable with signature func(labels: List[str]) which
+                        returns a boolean to determine if the partitions match.
+
+        If True, an exact match of partition labels between the to-be-merged
+        datasets is required in order to merge.
+        If False (Default), the partition labels of the dataset with fewer
+        partitions are interpreted as prefixes.
+    merge_tasks : List[Dict]
+        A list of merge tasks. Each item in this list is a dictionary giving
```
marco-neumann-jdas: does a merge task drop/consume its input tables? if not, I think this might be a memory issue.

PR #244, opened 2020-03-11
Type DatasetFactory

xhochy:


xhochy:
@fjetter Does this need a changelog entry?

kartothek/core/factory.py
```diff
@@ -39,12 +45,12 @@ class DatasetFactory(DatasetMetadataBase):
...
     def __init__(
         self,
-        dataset_uuid,
-        store_factory,
-        load_schema=True,
-        load_all_indices=False,
-        load_dataset_metadata=True,
-    ):
+        dataset_uuid: str,
+        store_factory: Callable[[], "KeyValueStore"],
```
lr4d: `store_factory()` is called a bit down, wouldn't this fail for `KeyValueStore`?

xhochy: We call `_check_callable` which should fail for just `KeyValueStore`.

lr4d: my bad, didn't read the types properly

xhochy:
pre-commit mystery again: locally they pass.

xhochy:
Mystery solved: The hooks run in CI with Python 3.6 and locally with Python 3.7. I have used Python 3.7+ features and thus some typing things are not correctly resolved in the CI.

fjetter:
> Does this need a changelog entry?

No

PR #250, opened 2020-03-16
Reset index with range index

xhochy:
We are the sole owner of the `DataFrame` in this case and can thus make in-place operations. For larger DataFrames, this was `7s` vs `5us` for me.

cc @mlondschien @jtilly

### DIFF
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -839,7 +839,9 @@ class MetaPartition(Iterable):
                     ind_col = ind_col.dt.date
             index_cols.append(ind_col)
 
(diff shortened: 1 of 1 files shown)


fjetter:
Interesting. Did you check if there are other places where this might be relevant?

xhochy:
> Interesting. Did you check if there are other places where this might be relevant?

No, I only checked the path behind `kartothek.io.eager.read_table` with a quite large DataFrame. After the PRs now all time is spent in `pd.concat` and `read_parquet`, these are obviously necessary. In my case, they did only make up 30% before because the DataFrame I had at hand was expensive to copy (40% float, 40% object, the rest other numeric columns).

PR #251, opened 2020-03-16
Only concat non-empty index_cols

xhochy:
Even running a no-op concat takes some time, so drop this.

cc @mlondschien @jtilly

### DIFF
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -852,7 +852,11 @@ class MetaPartition(Iterable):
         if cleaned_original_columns != original_columns:
             # indexer call is slow, so only do that if really necessary
             df = df.loc[:, cleaned_original_columns]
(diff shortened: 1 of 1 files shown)


kartothek/io_components/metapartition.py
```diff
@@ -852,7 +852,11 @@ def _reconstruct_index_columns(
         if cleaned_original_columns != original_columns:
             # indexer call is slow, so only do that if really necessary
             df = df.loc[:, cleaned_original_columns]
-        return pd.concat(index_cols + [df], axis=1, sort=False, join="inner")
+
+        if len(index_cols) > 0:
+            df = pd.concat(index_cols + [df], axis=1, sort=False, join="inner")
```
fjetter: I'm ok with the change but do we understand what's causing this? Why is concat slow if the input is a single item? That should be a no-op. Is this worth reporting to pandas?

PR #252, opened 2020-03-16
Only subset columns if table_columns are a strict subset

xhochy:
cc @mlondschien @jtilly


fjetter:
Is this for performance reasons or what's the issue?

xhochy:
Yes, this and all other PRs from me from yesterday are solely about performance and don't make a difference in behaviour.

PR #254, opened 2020-03-17
Add warning about shuffle method

fjetter:
# Description:

This adds a warning about the data shuffling method since this is a surprise and hard to debug.


lr4d:
hm I no longer have merge rights

Issue #263, opened 2020-03-25
Pandas 1.0 nullable data type produces ArrowTypeError


ged-steponavicius:
### Problem description

pyarrow pa.Schema.from_pandas does not work with new pandas 1.0 nullable integer type which breaks store_dataframes_as_dataset

### Example code 
```
import pandas as pd
from functools import partial
from tempfile import TemporaryDirectory
from storefact import get_store_from_url
from kartothek.io.eager import store_dataframes_as_dataset,update_dataset_from_dataframes

dataset_dir = TemporaryDirectory()

store_factory = partial(get_store_from_url, f"hfs://{dataset_dir.name}")
store = get_store_from_url( f"hfs://{dataset_dir.name}")

df = pd.DataFrame([{'int_col':1},{'int_col':2}])
df['int_col'] = df['int_col'].astype(pd.Int64Dtype())

dm = store_dataframes_as_dataset(
    store_factory,
    "sample_ds", 
    [df]
    )

```

error: ArrowTypeError: Did not pass numpy.dtype object

schema = pa.Schema.from_pandas(df) - does not work
schema = pa.Table.from_pandas(df).schema - works fine

pyarrow 0.16
kartothek 3.8.1

lr4d:
Thanks for the report @ged-steponavicius . I see you also opened this issue in the Arrow JIRA tracker, I'll just link that here for posterity https://issues.apache.org/jira/browse/ARROW-8223

xhochy:
This is a duplicate of https://github.com/JDASoftwareGroup/kartothek/issues/234 and will be fixed in the soon-to-be-released `pyarrow=0.17.0` and there is already a failing unit test in https://github.com/JDASoftwareGroup/kartothek/blob/5cba248cbfe9f7375ea3656850bbc98c7fb1746c/kartothek/io/testing/read.py#L733-L762 (it doesn't fail anymore with the Arrow nightlies).

lr4d:
Thanks for the input @xhochy , I'll close this for now

PR #264, opened 2020-03-25
replaced pa.Schema.from_pandas with pa.Table.from_pandas(df).schema

ged-steponavicius:
fixes https://github.com/JDASoftwareGroup/kartothek/issues/263

works with pyarrow 0.15+


lr4d:
@ged-steponavicius Thanks for your efforts again. Could you add a test for the pandas nullable type? As I presume you already know, the `pa.Table.from_pandas(df)` code fails with `pyarrow==0.13.0` with this dtype

btw @fjetter , given these incompatibilities between pyarrow <0.15 and >0.15, can we drop support for the older versions?

lr4d:
If there's no other way around this, I guess we should drop support for pyarrow<0.17 when 0.17 is released?

PR #266, opened 2020-03-30
some small performance improvements

marco-neumann-jdas:
See commits for details.

### DIFF
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -756,7 +756,7 @@ class MetaPartition(Iterable):
                     )
 
                 if list(df.columns) != table_columns:
-                    df = df.loc[:, table_columns]
+                    df = df.reindex(columns=table_columns, copy=False)
             new_data[table] = df
         return self.copy(data=new_data)
 
@@ -801,7 +801,7 @@ class MetaPartition(Iterable):
         ]
         if cleaned_original_columns != original_columns:
             # indexer call is slow, so only do that if really necessary
-            df = df.loc[:, cleaned_original_columns]
+            df = df.reindex(columns=cleaned_original_columns, copy=False)
 
         for pos, (primary_key, value) in enumerate(key_indices):
             # If there are predicates, don't reconstruct the index if it wasn't requested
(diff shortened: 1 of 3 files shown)


kartothek/io_components/metapartition.py
```diff
@@ -686,7 +686,17 @@ def load_dataframes(
...
+                if table_columns is not None:
+                    table_columns_to_io_candidates = sorted(
+                        set(self.table_meta[table].names) - set(table_columns)
+                    )
+                    if table_columns_to_io_candidates:
+                        table_columns_to_io = [table_columns_to_io_candidates[0]]
+                    else:
+                        table_columns_to_io = None
+                else:
+                    table_columns_to_io = None
```
fjetter: What's the background for this change? Can you add a comment in the code? I cannot see any details in the commit message

marco-neumann-jdas: "only load a single dummy column in MetaPartition.load_dataframes" is the commit message. This case happens if the user does not request any columns but we still need at least a single dummy column to reconstruct a non-broken DF. W/o this change, we just loaded ALL available columns which is way too much.

fjetter: Ok let me rephrase it: I suspected the "else we load everything" answer already but I'm not sure if this is clear in the code once a bit of time passes. Also, I'm pretty sure this was possible at some point. Do you know if anything in arrow changed in that regard?

fjetter:
Can you also add a changelog describing in which areas we can expect performance improvements? e.g.

* Read performance improvements for ...
   * partitioned datasets
   * queries with empty columns

or something like that.

kartothek/io_components/metapartition.py
lr4d: I took a brief look at the serialization module and it looks like if we pass `colums=[]` we get back a dataframe with an index the size of the number of rows of the Parquet file. Wouldn't that be faster? (Or is that not suitable for index reconstruction afterwards?)

marco-neumann-jdas: Let me explain it based on @lr4d question:

if you pass an empty column list, you want the set of rows (at least the DF len) to be restored. The issue is that not all serializers have dealed with this edge case correctly in the past. It was implemented here because it is the only location where the reader knows about potential dummy columns. That said, I removed the special case handling all together because to my knownledge, all serializers can now deal with this correctly.

lr4d: Thanks for the explanation @marco-neumann-jdas

PR #267, opened 2020-03-31
Pull Arrow nightlies from arrow-nightlies

xhochy:
Permanently moved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,9 @@ install:
   - conda activate test-env
   - >
     if [[ "${ARROW_NIGHTLY-0}" == 1 ]]; then
+        # Install both arrow-cpp and pyarrow to make sure that we have the
+        # latest nightly of both packages. It is sadly not guaranteed that the
+        # nightlies and the latest release would otherwise work together.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -44,7 +44,7 @@ install:
   - conda activate test-env
   - >
     if [[ "${ARROW_NIGHTLY-0}" == 1 ]]; then
-        travis_retry conda install -c ursa-labs/label/nightly pyarrow
+        travis_retry conda install -c arrow-nightlies arrow-cpp pyarrow
```
lr4d: do we need `arrow-cpp`?

xhochy: Locally I had the case that only `pyarrow` was pulled from the repo but `arrow-cpp` stayed on the conda-forge version. The two must though exactly match in the version used, so both should be pulled. The correct behaviour would be that `arrow-cpp` as a dependency would be automatically also be pulled from the repo but this is not yet happening consistently.

lr4d: maybe add that as a comment for posterity

xhochy:
@lr4d @fjetter Merge?

lr4d:
Thanks @xhochy

PR #272, opened 2020-04-07
Bunch of fixes for index type handling

marco-neumann-jdas:
See commits and changelog.

### DIFF
--- a/kartothek/core/index.py
+++ b/kartothek/core/index.py
@@ -885,6 +885,7 @@ def _index_dct_to_table(index_dct: IndexDictType, column: str, dtype: pa.DataTyp
         else:
             keys = np.asarray(list(keys_it))
 
+    # TODO: Remove work-around
     # This is because of ARROW-1646:
     #   [Python] pyarrow.array cannot handle NumPy scalar types
     # Additional note: pyarrow.array is supposed to infer type automatically.
--- a/kartothek/core/index.py
+++ b/kartothek/core/index.py
@@ -846,20 +846,44 @@ def _parquet_bytes_to_dict(column: str, index_buffer: bytes):
 
 
 def _index_dct_to_table(index_dct: IndexDictType, column: str, dtype: pa.DataType):
-    keys = index_dct.keys()
-    if (dtype is None) and (len(keys) > 0):
-        probe = next(iter(keys))
+    keys_it = index_dct.keys()
+
+    # find possible type probe
+    if len(keys_it) > 0:
+        probe = next(iter(keys_it))
+        has_probe = True
+    else:
+        probe = None
+        has_probe = False
+
+    # type inference
+    if (dtype is None) and has_probe:
         if isinstance(probe, np.datetime64):
             dtype = pa.timestamp(
                 "ns"
             )  # workaround pyarrow type inference bug (ARROW-2554)
         elif isinstance(probe, pd.Timestamp):
-            keys = [d.to_datetime64() for d in keys]
             dtype = pa.timestamp(
                 "ns"
             )  # workaround pyarrow type inference bug (ARROW-2554)
-        elif isinstance(probe, np.bool_):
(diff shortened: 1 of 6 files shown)


kartothek/core/index.py
```diff
@@ -846,20 +846,44 @@ def _parquet_bytes_to_dict(column: str, index_buffer: bytes):
 
 
 def _index_dct_to_table(index_dct: IndexDictType, column: str, dtype: pa.DataType):
```
fjetter, line 848: I hate this function so much :(

tests/conftest.py
```diff
@@ -31,6 +31,9 @@
 pd.options.mode.chained_assignment = "raise"
 
 
+pytest.register_assert_rewrite("kartothek.io.testing")
```
fjetter, line 34: I don't really get the docs of `pytest.register_assert_rewrite`. What I suspect: Will this make the output better readable?

kartothek/io_components/metapartition.py
```diff
@@ -1233,10 +1233,20 @@ def build_indices(self, columns):
...
+                meta.field(col).type
+                for meta in self.table_meta.values()
+                if col in meta.names
+            )[0]
+
             new_index = ExplicitSecondaryIndex(
-                column=col, index_dct={value: [self.label] for value in possible_values}
+                column=col,
+                index_dct={value: [self.label] for value in possible_values},
+                dtype=dtype,
```
fjetter, line 1247: We should've seen the lack of dtype as a warning somewhere, I believe. Did you evaluate whether this pops up somewhere else?

tests/conftest.py
marco-neumann-jdas: Yes. For example:

```python
assert some_dict == other_dict
```

in lib dir (not within `tests/`) only raises an `AssertionError` so the tests fails (this is correct), but you don't get a visual diff between the two dicts. This is only possible if pytest can rewrite the assertion (basically it replaces the bytecode snipped) to hook into it. See https://docs.pytest.org/en/latest/writing_plugins.html#assertion-rewriting

I find it extremely helpful to get the full pytest feature set (aka visual diffs) in our test suite.

kartothek/io_components/metapartition.py
marco-neumann-jdas: So normally this is OK, the index will infer the dtype during the store operation from `index_dct` (this even works for ints under python 3, since numpy ints have their own type now) or during load (from the parquet file).

But there is one edge case during index build where this does not work (see last commit): when the `index_dct` is empty. In that case, the index constructor even raises, since the type inference would obviously be bollocks. I am somewhat surprised that nobody ran into this ("create index where one partition has an empty DF").

Also, who emits the warning?

kartothek/core/index.py
```diff
@@ -846,20 +846,44 @@ def _parquet_bytes_to_dict(column: str, index_buffer: bytes):
...
+            # numpy might create object ndarrays here, which pyarrow might (for some reason) convert fo floats
+            keys = list(keys_it)
+        elif (
+            dtype == pa.timestamp("ns")
+            and has_probe
+            and isinstance(probe, pd.Timestamp)
+        ):
+            keys = np.asarray([d.to_datetime64() for d in keys_it])
+        else:
+            keys = np.asarray(list(keys_it))
```
lr4d, line 887:
```suggestion
    # TODO: Remove work-around
```

kartothek/io_components/metapartition.py
fjetter: Sorry, I was mistaken. I thought we disallowed this / deprecated this but that's not the case.

tests/conftest.py
fjetter: good to know, thanks!

fjetter:
Builds are failing for arrow 0.13.0

PR #273, opened 2020-04-08
Fix some deprecations and other warnings

marco-neumann-jdas:
See commit msgs for details.

There is one notable case left where we test the roundtrip of byte-string columns through pyarrow, likely a leftover from python 2. Since I would like to raise in the serializer in that case and also adjust the tests, this will be done in another PR.

### DIFF
--- a/kartothek/serialization/_generic.py
+++ b/kartothek/serialization/_generic.py
@@ -378,7 +378,7 @@ def _ensure_type_stability(
         error messages.
     """
 
-    value_dtype = pd.Series(value).dtype
+    value_dtype = pd.Series(value if is_list_like(value) else [value]).dtype
     array_like, array_value_type = _handle_categorical_data(array_like, require_ordered)
     array_like, array_value_type = _handle_null_arrays(array_like, value_dtype)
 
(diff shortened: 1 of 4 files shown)


kartothek/serialization/_generic.py
```diff
@@ -378,7 +378,7 @@ def _ensure_type_stability(
         error messages.
     """
 
-    value_dtype = pd.Series(value).dtype
+    value_dtype = pd.Series(value if is_list_like(value) else [value]).dtype
```
fjetter, line 381: Is this user facing, i.e. is it relevant enough to mention in a changelog?

marco-neumann-jdas: Do we mention fixed warnings in the changelog? I'm OK with this, just wanted to clarify.

fjetter: If it's just about the warning, I don't care at all. I thought it was a fix since you added another test

marco-neumann-jdas: Valid assumption/question. Added the other test to make sure it is working (like a safety net). The test status didn't change due to the actual change here.

PR #274, opened 2020-04-08
fix test_arrow_compat test

marco-neumann-jdas:
- wrong if condition preventing actual testing
- remove 'datetime64_ts' column from 0.12.1 and 0.13.0 reference files
- add 0.16.0 reference data + test fixture


tests/serialization/test_arrow_compat.py
```diff
@@ -49,6 +49,8 @@ def test_arrow_compat(arrow_version, reference_store, mocker):
     restored = ParquetSerializer().restore_dataframe(
         store=reference_store, key=arrow_version + ".parquet", date_as_object=True
     )
-    if arrow_version == "0.14.1" and not ARROW_LARGER_EQ_0141:
+
+    if arrow_version in ("0.14.1", "0.15.0", "0.16.0") and not ARROW_LARGER_EQ_0141:
```
fjetter, line 53: Wouldn't the `ARROW_LARGER_EQ_0141` be sufficient?

marco-neumann-jdas: No, under pyarrow 0.13.1 (lib) the fix is not required for the 0.12 and 0.13 reference files (`arrow_version` => stored reference file, `ARROW_LARGER_EQ_0141` => current lib/runtime version)

Issue #259, opened 2020-03-19
BUG: Timestamp with timezone produce Schema violation


ged-steponavicius:
Pandas DateTime fields with timezone do not survive the round trip and produce an error on update_dataset_from_dataframes

```
import pandas as pd
import numpy as np

from functools import partial
from tempfile import TemporaryDirectory
from storefact import get_store_from_url
from kartothek.io.eager import store_dataframes_as_dataset,update_dataset_from_dataframes

dataset_dir = TemporaryDirectory()

store_factory = partial(get_store_from_url, f"hfs://{dataset_dir.name}")

df = pd.DataFrame([{'part':1,'date':pd.to_datetime('now'),'date_utc':pd.to_datetime('now',utc=True),'date_tz':pd.to_datetime('now')},
                   {'part':2,'date':pd.to_datetime('now'),'date_utc':pd.to_datetime('now',utc=True),'date_tz':pd.to_datetime('now')}])
df['date_tz'] = df['date_tz'].dt.tz_localize(tz='US/Eastern')


dm = store_dataframes_as_dataset(
    store_factory,
    "sample_ds", 
    [df],
    overwrite=True,
    partition_on=['part']

 )

dm = update_dataset_from_dataframes(
    [df.loc[df['part']==2]],
    store_factory,
    "sample_ds",
    partition_on=['part'],
    delete_scope=[{'part':2}]

 )

```
error: 

ValueError: Schemas for table 'table' of dataset 'sample_ds' are not compatible!
```
Schema violation

Origin schema: {table/part=2/c826fc19c19f48a4bf98dc519ed2bde2}
Origin reference: {sample_ds/table/_common_metadata}

Diff:
Arrow schema:
@@ -1,4 +1,4 @@

 part: int64
 date: timestamp[ns]
-date_tz: timestamp[ns, tz=UTC]
+date_tz: timestamp[ns, tz=US/Eastern]
 date_utc: timestamp[ns, tz=UTC]
```

The core of an issue appears to be down to 2 factors:
1)  _bytes2schema from common_metadata
```
if f.type == pa.timestamp("us"):
        f = pa.field(f.name, pa.timestamp("ns"))
```
field type with tz != pa.timestamp("us"), granularity remains unchaged

2) _schema2bytes from common_metadata
`pq.write_metadata(schema, buf, version="2.0", coerce_timestamps="us")`
coerce_timestamps appears to drop tz information, sets tz to UTC

another test to confirm:
```
fields = [
[...]

PR #260, opened 2020-03-19
fixed schema violation for timestamp with tz

ged-steponavicius:
# Description:
fixes issue https://github.com/JDASoftwareGroup/kartothek/issues/259

PR #278, opened 2020-04-14
WIP: Tz bug fix rebased

lr4d:
Supersedes https://github.com/JDASoftwareGroup/kartothek/pull/260

I tried to rebase this branch on top of master but it was quite the adventure to do so.
I excluded the changes to the changelog and it seems like some removal of comments is also missing, but the important parts should be there.

cc @ged-steponavicius


kartothek/core/common_metadata.py
```diff
@@ -78,6 +77,35 @@ def _schema_compat(self):
 
             schema = schema.remove_metadata()
             md = {b"pandas": _dict_to_binary(pandas_metadata)}
+            # https://github.com/JDASoftwareGroup/kartothek/issues/259
+
+            if schema is not None:
```
marco-neumann-jdas, line 82: why can it be `None` here? 3 lines above we just call `schema = schema.remove_metadata()`. Can this return `None`?

kartothek/core/common_metadata.py
```diff
@@ -78,6 +77,35 @@ def _schema_compat(self):
...
+                fields = []
+                for f in schema:
```
marco-neumann-jdas, line 84: short comment before the entire massaging would be nice. This function just does more and more and it is hard to follow along.

kartothek/core/common_metadata.py
```diff
@@ -380,24 +408,14 @@ def store_schema_metadata(schema, dataset_uuid, store, table):
...
-        f = schema[idx]
-
-        # schema data recovered from parquet always contains timestamp data in us-granularity, but pandas will use
-        # ns-granularity, so we re-align the two different worlds here
-        if f.type == pa.timestamp("us"):
-            f = pa.field(f.name, pa.timestamp("ns"))
-
-        fields.append(f)
-    return pa.schema(fields, schema.metadata)
+    return schema
```
marco-neumann-jdas, line 418: a direct return from the line above would work now as well.

reference-data/arrow-compat/batch_generate_references.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
marco-neumann-jdas, line 1: thanks a lot for adding this tool!

tests/serialization/test_arrow_compat.py
```diff
@@ -53,4 +52,7 @@ def test_arrow_compat(arrow_version, reference_store, mocker):
     if arrow_version in ("0.14.1", "0.15.0", "0.16.0") and not ARROW_LARGER_EQ_0141:
         orig = orig.astype({"null": float})
 
+    if LooseVersion(arrow_version) < "0.13.0":
```
marco-neumann-jdas, line 55: can you add a short comment on the reason? Is this because timezones cannot be preserved with these old versions?

lr4d: Because the changes from @ged-steponavicius fixes https://github.com/JDASoftwareGroup/kartothek/issues/259, enabling proper compatibilty of datetimes with timezones i.e. these columns, and we currently only support pyarrow >= 0.13.0

fjetter: Then why do we need to check `< "0.13.0":` That's dead code, isn't it?

lr4d:
This PR now also includes the appropriate changelog entry and I added some commits to update the reference Parquet files.

@ged-steponavicius feel free to push to this branch if you'd like to address the review comments

tests/serialization/test_arrow_compat.py
lr4d: `arrow_version` here refers to the arrow version with which the reference data file was created

lr4d:
```suggestion
    if LooseVersion(arrow_version) < "0.13.0":  # gh-259: compatibilty of datetimes with timezones only supported in kartothek versions with pyarrow >= 0.13.0
```

marco-neumann-jdas: suggestion looks good to me, thanks :)

PR #286, opened 2020-05-29
Compress payload columns during shuffle

fjetter:
This compresses the payload columns into a single bytestream before shuffle. This has several benefits

* network/disk IO significantly reduced
* intermediate (de)serialization is cheaper

I'm very confident that there is still a lot of room for improvement here but this simple implementation showed already about an order of magnitude better performance on small scale tests. I'll also see if we can get something similar into dask right away.

Credits go to @marco-neumann-jdas for the idea

### DIFF
--- a/kartothek/io/dask/_update.py
+++ b/kartothek/io/dask/_update.py
@@ -1,18 +1,24 @@
 import logging
 from functools import partial
-from typing import List
+from typing import Callable, List, Optional, Union
 
+import dask.array as da
+import dask.dataframe as dd
 import numpy as np
 import pandas as pd
+from dask.delayed import Delayed
+from simplekv import KeyValueStore
 
 from kartothek.io_components.metapartition import (
     MetaPartition,
     parse_input_to_metapartition,
 )
 from kartothek.io_components.utils import sort_values_categorical
+from kartothek.serialization import DataFrameSerializer
 
 from ._utils import map_delayed
 
+StoreFactoryType = Callable[[], KeyValueStore]
 _logger = logging.getLogger()
 
 _KTK_HASH_BUCKET = "__KTK_HASH_BUCKET"
@@ -37,8 +43,7 @@ def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
     return df.assign(**{_KTK_HASH_BUCKET: buckets.astype(f"uint{bit_width}")})
 
 
-def _pack_payload(partition: pd.DataFrame, group_key: List[str]) -> pd.DataFrame:
-    # TODO: benchmark how good plain parquet would do
+def pack_payload_pandas(partition: pd.DataFrame, group_key: List[str]) -> pd.DataFrame:
     try:
         # Technically distributed is an optional dependency
         from distributed.protocol import serialize_bytes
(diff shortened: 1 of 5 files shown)


fjetter:
The coverage failure comes from not testing the importerror

kartothek/io/dask/_update.py
```diff
@@ -35,6 +37,40 @@ def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
...
+        from distributed.protocol import deserialize_bytes
+    except ImportError:
+        _logger.warning(
+            "Shuffle payload columns cannot be compressed since distributed is not installed."
+        )
+        return partition
+
+    return pd.concat(
+        partition[_PAYLOAD_COL].map(lambda part: deserialize_bytes(part)).values,
+        copy=False,
```
marco-neumann-jdas, line 70: does this work w/o `ignore_index=False` aka are we sure that the index is unique here?

kartothek/io/dask/_update.py
```diff
@@ -51,14 +87,16 @@ def _update_dask_partitions_shuffle(
     if ddf.npartitions == 0:
         return ddf
 
+    group_cols = partition_on[:]
```
marco-neumann-jdas, line 90: is this an implicit copy? If so, I would prefer `copy.copy` instead for readability/clarity.

kartothek/io/dask/_update.py
```diff
@@ -35,6 +37,40 @@ def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
     return df.assign(**{_KTK_HASH_BUCKET: buckets.astype(f"uint{bit_width}")})
 
 
+def _pack_payload(partition: pd.DataFrame, group_key: List[str]) -> pd.DataFrame:
```
marco-neumann-jdas, line 40: I think we could use some short explanation either in the `_[un]pack_payload` docstrings or on the module-level (top-docstring) that explains what's going on there. Esp. replacing the entire group DF with a single row and packing the content into a binary columns only makes sense to me after reading it twice. I could also volunteer to do some ascii-art-drawing / example stuff if you want. I just think the entire module actually uses a quite simple / streamlined approach (which is nice) but from just reading the code, it requires substantial brain power to understand it.

kartothek/io/dask/_update.py
fjetter: I thought about it and am not entirely sure how dask depends on these indices. I'll try to come up with a few test cases to check this.

kartothek/io/dask/_update.py
fjetter: sure, can do

kartothek/io/dask/_update.py
fjetter: I would just post an example like

pd.DataFrame({"group_col": [1, 2], "payloadA": ..., "payloadB": ...})

becomes

pd.DataFrame({"group_col": [1, 2], "payload_bytes": b"long byte string"})


if your internal ascii artist comes up with something nicer, I wouldn't mind of course :)

fjetter: I can also add a module doc string and also explain the bucketing/hashing to wrap things up

fjetter:
When packing this into dask, the test suite lights up like a christmas tree due to many edge cases. will investigate a bit before we can merge this

kartothek/io/dask/_update.py
```diff
@@ -51,14 +87,16 @@ def _update_dask_partitions_shuffle(
...
 
+    group_cols = partition_on[:]
     if num_buckets is not None:
         meta = ddf._meta
         meta[_KTK_HASH_BUCKET] = np.uint64(0)
         ddf = ddf.map_partitions(_hash_bucket, bucket_by, num_buckets, meta=meta)
-        group_cols = [partition_on[0], _KTK_HASH_BUCKET]
-    else:
-        group_cols = [partition_on[0]]
+        group_cols.append(_KTK_HASH_BUCKET)
```
lr4d, line 95: why this change in logic?

kartothek/io/dask/_update.py
```diff
@@ -35,6 +37,40 @@ def _hash_bucket(df: pd.DataFrame, subset: List[str], num_buckets: int):
...
+        # Technically distributed is an optional dependency
+        from distributed.protocol import deserialize_bytes
+    except ImportError:
+        _logger.warning(
+            "Shuffle payload columns cannot be compressed since distributed is not installed."
+        )
+        return partition
+
+    return pd.concat(
+        partition[_PAYLOAD_COL].map(lambda part: deserialize_bytes(part)).values,
```
lr4d, line 69: style np
```suggestion
        partition[_PAYLOAD_COL].map(deserialize_bytes).values,
```

kartothek/io/dask/_update.py
fjetter: unrelated. this was a ill informed "optimization" when I initially created this. It is, however, quite harmful and creates possible very large hotspots. Consider a partitioning like
`["year", "month", "day"]` which would data data of each year on a single worker even though the data could be chunked much more granular.

Generally speaking, the less this is restricted, the more homogeneous the intermediate jobs will distribute the data and we should allow as much freedom here as possible. in the end the load balancing is done via a hash bucketing more or less identical to how we calculate buckets -> more input values will yield a better distribution in buckets (statistically speaking, there are of course exceptions)

kartothek/io/dask/_update.py
fjetter: oops. slipped in from an earlier iteration

kartothek/io/dask/_update.py
fjetter: At least dask itself is sensitive to this (e.g. for `set_index`). I guess within the kartothek context we get away with `ignore_index=True` (We don't properly preserve DF indices everywhere, anyhow)

kartothek/io/dask/_update.py
lr4d: Thanks for the explanation. I think this may be more appropriate in a different commit, do you agree?

kartothek/io/dask/_update.py
marco-neumann-jdas: I think an example is already very very good, because it's something most people can relate to :)

fjetter: Added documentation and type annotations to the most relevant functions. Also added an explicit example how the packing will impact a partition.

kartothek/io/dask/_update.py
fjetter: If anything, this could be a separate PR. Since this is internal and the user facing thing will be "Significant improvement" in performance I kept it the way it is

PR #287, opened 2020-05-29
Add a dispatch_by argument to load_dataset_as_ddf

mlondschien:
The `karthothek.io.dask.dataframe.load_dataset_as_ddf` does not have a `dispatch_by` argument. It does, however, have a `dask_index_on` argument, which is passed to `dispatch_by` to `kartothek.io.delayed.load_dataset_as_delayed` internally. As dask does not support multiindices, we cannot pass a list of columns to `dask_index_on`, as would be possible with `dispatch_by`. As a proposed solution this PR adds a `dispatch_by` argument to `load_dataset_as_ddf`, which, if no `dask_index_on` is supplied, is passed to `load_dataset_as_delayed`.

### DIFF
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -36,6 +36,7 @@ def read_dataset_as_ddf(
     predicates=None,
     factory=None,
     dask_index_on=None,
+    dispatch_by=None,
 ):
     """
     Retrieve a single table from a dataset as partition-individual :class:`~dask.dataframe.DataFrame` instance.
@@ -47,7 +48,8 @@ def read_dataset_as_ddf(
     Parameters
     ----------
     dask_index_on: str
-        Reconstruct (and set) a dask index on the provided index column.
+        Reconstruct (and set) a dask index on the provided index column. Cannot be used
+        in conjunction with `dispatch_by`.
 
         For details on performance, see also `dispatch_by`
     """
@@ -55,6 +57,13 @@ def read_dataset_as_ddf(
         raise TypeError(
             f"The paramter `dask_index_on` must be a string but got {type(dask_index_on)}"
         )
+
+    if dask_index_on is not None and dispatch_by is not None:
+        raise ValueError(
+            "`read_dataset_as_ddf` got parameters `dask_index_on` and `dispatch_by`. "
+            "Note that `dispatch_by` can only be used if `dask_index_on` is None."
+        )
+
     ds_factory = _ensure_factory(
         dataset_uuid=dataset_uuid,
         store=store,
@@ -81,7 +90,7 @@ def read_dataset_as_ddf(
         label_filter=label_filter,
         dates_as_object=dates_as_object,
         predicates=predicates,
-        dispatch_by=dask_index_on,
+        dispatch_by=dask_index_on if dask_index_on else dispatch_by,
     )
     if dask_index_on:
         divisions = ds_factory.indices[dask_index_on].observed_values()
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -58,7 +58,7 @@ def read_dataset_as_ddf(
             f"The paramter `dask_index_on` must be a string but got {type(dask_index_on)}"
         )
 
-    if dask_index_on is not None and dispatch_by is not None:
+    if dask_index_on is not None and dispatch_by is not None and len(dispatch_by) > 0:
         raise ValueError(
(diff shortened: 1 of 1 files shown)


fjetter:
Allowing the dispatch by is fine for me if it is useful. I didn't add it intentionally since it can only be used to control the internal structure of the dataframe but I can understand that it can offer benefits by not having to reshuffle the data if the physical layout suits the dispatch.

However, since you're mentioning the multiindex, dask cannot really leverage this information. So when performing expensive operations like a shuffle/groupby which would usually benefit from having an index, this would not help much.

mlondschien:
Yes. We are only interested in `dispatch_by` to avoid reshuffling the data.

fjetter:
Just to clarify my point. Let's assume you have the following physical files

```
# A

ColA    ColB    Payload
0       0       a
0       0       b
0       1       c
0       1       d

# B

ColA    ColB    Payload
1       0       a
1       0       b
1       1       c
1       1       d
```

using `dispatch_by=["ColA", "ColB"]` would create four dask partitions, one for every observed tuple. In this particular case, this may use predicate pushdown assuming that the files were written using row group chunk sizes.


If you now perform

`ddf.groupby(["ColA", "ColB"]).apply(lambda x: x)`

dask will still perform a full shuffle since it doesn't know anything about the internal structure. I would not even dare to assume that we would save network IO since the intermediate shuffle steps use a hash algorithm which would mix the data.

The only way I can see this to be really benefitial is if you do stuff like

`ddf.map_partitions(function_which_assumes_the_groups_are_as_expected)`

where you could save the reshuffle.

If you perform stuff like this, I would probably propose to use the bag interface, see https://kartothek.readthedocs.io/en/latest/_rst/kartothek.io.dask.bag.html#kartothek.io.dask.bag.read_dataset_as_dataframe_bag

For example

```
kartothek.io.dask.bag.read_dataset_as_dataframe_bag(
)

```

mlondschien:
Thanks! This is basically what we are doing. Our use-case is as follows:

We have a dataset partitioned on columns `["A", "B", "chunk"]`, where `A`, `B` are actually columns we are interested in, and `chunk` is a (semi-)random column taking values 0 to 255 which we use to get reasonably sized partitions (~10k rows each) and draw subsets when experimenting. The chunks are constructed in such a way that we are always fine with doing `ddf.map_partitions(...)`. 

Our dataset contains a third column (say `C`), which is zero for most (~99%) of the observations. We are often (most of the time) only interested in drawing those observations which have a non-zero value for `C`, and would like to load this (fast). If we partition our dataset on `C`, since `chunk` is semi-random (and needs to be), we get, for each value of `A` and `B`, a normal sized (~10k observations) partition for observations with zero in `C` and a tiny partition (~100 observations) with nonzero values in `C`. If we then load (with the eager backend) using predicates `[[("C", ">", 0)]]`, the dataset takes quite long to load due to all the `pd.concat` happening. 

Our solution for this is to simply load the dataset using the predicate `[[("C", ">", 0)]]`, shuffle, and write a new dataset that is not partitioned on `chunk`. This reduces the amount of `pd.concat` required by a factor of 256 when loading. Since the shuffling does not always work smoothely, it is much nicer to use set `dispatch_by=["A", "B"]` in `load_dataset_as_...`. This PR intends to add this argument to `load_dataset_as_ddf` as to not use workarounds (e.g. using delayed objects, as we try to consistently use dask dataframes).

mlondschien:
I'm not seeing any tests for e.g. the `dispatch_by` argument of `read_dataset_as_delayed` (or the `dask_index_on` argument of `read_dataset_as_ddf`). Could you outline what exactly you would like to test here @lr4d ?

fjetter:
The dispatch by functionality is tested in `tests/io_components/test_read.py`. afaik, we do not have full end-to-end tests for this anywhere yet and I believe this is fine w/out additional tests.

@lr4d what caveat are you referring to?

lr4d:
@fjetter what you refer to in an earlier post where you mention:
> dask will still perform a full shuffle since it doesn't know anything about the internal structure

@mlondschien I was thinking of a simple test that ensures the partitions are in the way we expect them to be, but since we are not testing this in a generic way atm, I guess we can leave this out for now

lr4d:
Actually, @mlondschien can you add a changelog entry for this?

fjetter:
will add the changelog myself and release afterwards. thanks @mlondschien

mlondschien:
Sorry, couldn't get to it fast enough :). Thank you!

PR #289, opened 2020-06-02
Add default table to load_dataset_as_ddf and update_dataset_from_ddf

### LINKED ISSUE #288: read_dataset_as_ddf and update_dataset_from_ddf raise if optional table argument is not provided.
### Problem description

The `read_dataset_as_ddf` function has an optional argument `table`. However if not provided, a `KeyError` will raise [here](https://github.com/JDASoftwareGroup/kartothek/blob/f7dc8f985414801b115842a169805344cf1378a5/kartothek/io/dask/dataframe.py#L104). Similarly, the `table` argument of the `update_dataset_from_ddf` function is optional according to the docs, but will result in a `TypeError` [here](https://github.com/JDASoftwareGroup/kartothek/blob/f7dc8f985414801b115842a169805344cf1378a5/kartothek/io/dask/dataframe.py#L230).

### Possible solutions

A simple solution would be to pass `"table"` (or rather `kartothek.io_components.metapartition.SINGLE_TABLE`) as a default value. This however would raise the question why the function is called `read_dataset_as_ddf` rather than `read_table_as_ddf`, especially as `read_table_as_delayed` is called within `read_dataset_as_ddf`, so maybe there exists a less obvious solution.

mlondschien:
Closes #288

### DIFF
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -9,7 +9,7 @@ from kartothek.core.common_metadata import empty_dataframe_from_schema
 from kartothek.core.docs import default_docs
 from kartothek.core.factory import _ensure_factory
 from kartothek.core.naming import DEFAULT_METADATA_VERSION
-from kartothek.io_components.metapartition import parse_input_to_metapartition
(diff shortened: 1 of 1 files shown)


mlondschien:
It appears that I am not authorized to merge even after approval :(.

PR #292, opened 2020-06-03
Allow pyarrow==0.17.X

fjetter:
We're currently still stuck on 0.13.X in some places which is why I don't want to drop the support for the old versions, yet. We're moving on towards 17 and I hope the release following this will be the last one with support for <=15


fjetter:
Builds were successful. No idea why the github integration behaves weird again

PR #293, opened 2020-06-04
fix unpack_payload and add tests

marco-neumann-jdas:
Not sure if we should make this function public and if so: where it should live.


fjetter:
I still hope this will end up in dask but I don't mind making this public. Even though I don't really like `utils`  I can see this in a `kartothek.io.dask.dataframe.utils` module since it doesn't tie in to anything else properly

PR #297, opened 2020-06-09
Change to new NumFOCUS nightly wheels location

xhochy:
Since three months they are only uploaded there.


xhochy:
We don't test on Python 3.8 but `pandas` only publishes nightly builds for Python 3.8.

PR #298, opened 2020-06-09
Test with Python 3.8

xhochy:


xhochy:
Requires at least https://github.com/JDASoftwareGroup/kartothek/pull/299

xhochy:
Travis is now also happy. I applied some of Travis' linter suggestion here to get this working, mostly different names but no difference in behaviour.

PR #300, opened 2020-06-12
parametrize value error for external testing backends

usha-nemani-jdas:
# Description:
- Parametrized ValueError in kartothek.io.testing.update.py to support external testing

### DIFF
--- a/kartothek/io/testing/update.py
+++ b/kartothek/io/testing/update.py
@@ -662,7 +662,14 @@ def test_update_first_time_with_secondary_indices(
     )
 
 
-def test_partition_on_null(store_factory, bound_update_dataset):  # gh-262
+@pytest.fixture
+def expected_error():
+    return ValueError
+
+
+def test_partition_on_null(
+    store_factory, bound_update_dataset, expected_error
+):  # gh-262
     keys = ["a", "b", "c", np.nan]
     values = range(len(keys))
     d = dict(zip(keys, values))
@@ -673,7 +680,8 @@ def test_partition_on_null(store_factory, bound_update_dataset):  # gh-262
     )
 
     with pytest.raises(
-        ValueError, match=r"Original dataframe size .* on a column with null values."
+        expected_error,
+        match=r"Original dataframe size .* on a column with null values.",
     ):
         bound_update_dataset(
             [{"data": {"table": df}}],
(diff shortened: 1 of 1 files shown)


kartothek/io/testing/update.py
```diff
@@ -662,7 +662,14 @@ def test_update_first_time_with_secondary_indices(
     )
 
 
-def test_partition_on_null(store_factory, bound_update_dataset):  # gh-262
+@pytest.fixture
+def expected_error():
```
lr4d, line 666: Please add a comment here, to explain the rationale of this logic

kartothek/io/testing/update.py
```diff
@@ -673,7 +680,8 @@ def test_partition_on_null(store_factory, bound_update_dataset):  # gh-262
     )
 
     with pytest.raises(
-        ValueError, match=r"Original dataframe size .* on a column with null values."
```
lr4d, line 676: I'd also be fine with just having `Exception` here instead of `ValueError`. @usha-nemani-jdas would that fix your issue?

kartothek/io/testing/update.py
usha-nemani-jdas: When I call metaparttion.partition_on using yamal_pipeline, it throws YamalError. 
test_partition_on_null  need to have a common fixture to support various testing backends.

kartothek/io/testing/update.py
usha-nemani-jdas: Exception will fix my issue, since it is parent of ValueError and YamalError . It would support all other Error Types as well. Will do code changes accordingly.

KrishnarajAjayGharpure: I am not comfortable with the idea of generalizing it to an `Exception` when we know what type of exception to expect...

lr4d: We do not have a proper custom exception for this error case anyway, which imo would be the most correct way of handling it. I'm fine with the tradeoff for test simplicity in this case as we are also matching the error string

lr4d: @KrishnarajAjayGharpure are you okay if we merge this?

KrishnarajAjayGharpure: @lr4d - you make a good point about the string matching. However, I think this change is small enough to not overly worry and merge in the interest of expediency.

PR #304, opened 2020-06-24
Remove dispatch of metadata and indices

fjetter:
# Description:

I stumbled upon a dataset where the user stores user metadata of about 10kb in the dataset which is perfectly fine imo. However, the dataset itself also has about 100k partitions which generates about 1GB of dummy payload attached to the task graph although this is not used anywhere. This is quite a nonsense feature in the first place and a remnant of the very early days so I'd like to remove this. On the off chance that anybody is using the metapartitions pipeline (soon to be private, I'd suggest) I introduced this feature toggle. All ordinary reader pipelines should only use the simple dispatch

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,14 +2,30 @@
 Changelog
 =========
 
-Version 3.9.1 (2020-06-XX)
-==========================
+Version 3.10.0 (2020-06-XX)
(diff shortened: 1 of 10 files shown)


fjetter:
build failure due to docs...

PR #305, opened 2020-06-26
wip: use distributed.Queue as storage for metadata poc

lr4d:
With the intention of improving resilience,
> instead of storing the intermediate results (partition metadata) on dask workers we could submit those to a more central instance (an event bus or even simpler the dask scheduler)
this way the jobs would be safe from worker failures

cc @fjetter

### DIFF
--- a/kartothek/io/dask/_update.py
+++ b/kartothek/io/dask/_update.py
@@ -280,6 +280,7 @@ def update_dask_partitions_one_to_one(
     df_serializer: DataFrameSerializer,
     dataset_uuid: str,
     sort_partitions_by: Optional[str],
+    metadata_storage_factory=None,
 ) -> List[Delayed]:
     """
     Perform an ordinary, partition wise update where the usual partition
@@ -307,13 +308,32 @@ def update_dask_partitions_one_to_one(
     if secondary_indices:
         mps = map_delayed(MetaPartition.build_indices, mps, columns=secondary_indices)
 
-    return map_delayed(
+    mps = map_delayed(
         MetaPartition.store_dataframes,
         mps,
         store=store_factory,
         df_serializer=df_serializer,
         dataset_uuid=dataset_uuid,
     )
+
+    if metadata_storage_factory:
+
+        def send_mp_to_queue(mp, metadata_storage_factory):
+            # We can't msgpack serialize Metapartition by itself, so we serialize each mp dict
+            # We need to instantiate a worker client to be able to push to a queue
+            from distributed import get_client
+
+            c = get_client()  # noqa
+            metadata_store = metadata_storage_factory()
+            for _mp in mp.metapartitions:
+                metadata_store.put(_mp)
+
+        tasks = map_delayed(
+            send_mp_to_queue, mps, metadata_storage_factory=metadata_storage_factory
+        )
+        return tasks  # We need to pass this so it gets executed
+    else:
+        return mps
 
 
 def _store_partition(
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -149,6 +149,7 @@ def update_dataset_from_ddf(
     partition_on=None,
     factory=None,
     bucket_by=None,
+    metadata_resilience=True,
 ):
     """
     Update a dataset from a dask.dataframe.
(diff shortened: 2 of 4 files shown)


lr4d:
cc @marco-neumann-jdas @pacman82 in case you want to take a look at this

marco-neumann-jdas:
What kind of resilience do you get here? Against which exact failure scenario does this change help?

kartothek/io/dask/_update.py
```diff
@@ -307,14 +308,33 @@ def update_dask_partitions_one_to_one(
...
+
+        def send_mp_to_queue(mp, metadata_storage_factory):
+            # We can't msgpack serialize Metapartition by itself, so we serialize each mp dict
+            # We need to instantiate a worker client to be able to push to a queue
+            from distributed import get_client
+
+            c = get_client()  # noqa
+            metadata_store = metadata_storage_factory()
+            for _mp in mp.metapartitions:
+                metadata_store.put(_mp)
```
marco-neumann-jdas, line 329: this task might get retried. Do you do any de-duplication of the results in the end?

lr4d: good point

- [ ] make `partition_list` a set in `update_dataset_from_partitions` when `metadata_storage_factory`, to prevent potential duplicates

lr4d:
>What kind of resilience do you get here? Against which exact failure scenario does this change help?

This would help when a worker which has stored the data dies while holding the metadata, and would imply re-computing all dependencies for that data. Since the metadata is quite small, we can just move all metadata objects to a central place (i.e. the scheduler) to avoid having to re-compute the graph after data has been stored
An example when this can occur is at the last part of a "dataset update" graph, when most workers have already stored the data and the cluster may be getting downscaled because most workers are no longer necessary. However, removing a worker which is idle but holds metadata would imply re-computing all dependencies for that data, whereas with this implementation, the cluster could be downscaled up to a single worker (which would make the final metadata commit) without issues

marco-neumann-jdas:
> This would help when a worker which has stored the data dies while holding the metadata, and would imply re-computing all dependencies for that data. 

How does that work then? Dask doesn't know that you're bypassing its dependency system, so it will re-trigger the computation. And you don't know that all you can safely abort that and run some kind of manual collect+store operation w/o looking very carefully at the scheduler dashboard, because you cannot know that all dependencies where executed at least once. In case the cluster start "flapping", you don't know which parts of the dependency chain are recomputed because the got lost and which parts did not get computed in the first place. And every manual guess makes very strong assumption about the scheduling logic (which IIRC is an implementation detail of dask/distributed) and risks data loss.

lr4d:
> How does that work then? Dask doesn't know that you're bypassing its dependency system, so it will re-trigger the computation.

I remember this coming up when I started to experiment on this but I guess I didn't address that very properly here. I'll have a look when I have some spare time. Appreciate your feedback

fjetter:
> How does that work then? Dask doesn't know that you're bypassing its dependency system, so it will re-trigger the computation.

This could only be addressed by letting every task check whether a key/result was already logged on the scheduler. This way, the task would be rescheduled but it would be a "no-op"


We may also raise this as an issue in distributed. If this is nicely integrated in the scheduler that would be much better, of course. I can see various arguments pro/con so this will probably trigger a small debate. There are tasks in dask/distributed itself, though, which would benefit. Prime example is the parquet storage system of dask itself...

OTOH, this topic occasionally pops up and the "dask way" of dealing with this would probably be a duplication of results across the cluster but that doesn't properly work atm

lr4d:
I got this working in a hacky way by calling `compute` on the tasks before the final metadata commit, thereby bypassing the dask dependency management

i.e. we first execute one graph which includes everything up until the metadata is at the `distributed.Queue`
image
And, finally, we perform the metadata commit with the information in the queue

marco-neumann-jdas:
Instead of adding more complexity to the payload code with store interactions, retries, no-op checks etc, could we rather implement this as some kind of graph rewrite?

lr4d:
>could we rather implement this as some kind of graph rewrite

I guess that may be the "cleanest" given the current state of distributed, and probably shouldn't be too difficult

Ideally, though, I would see this kind of functionality properly supported in distributed, as @fjetter mentions
> We may also raise this as an issue in distributed. If this is nicely integrated in the scheduler that would be much better, of course

PR #306, opened 2020-07-02
Feat/collect parqet statistics

NeroCorleone:
The idea is to collect the parquet metadata for each MetaPartition and derive interesting numbers from it. First shot is to have an interface function which only returns a dataframe with interesting numbers.
We could think about to extending this to something report-like including nice (and meaningful) plots, if we want to.

### DIFF
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from functools import partial
 
 import dask.bag as db
+from toolz.itertoolz import random_sample
 
 from kartothek.core import naming
 from kartothek.core.docs import default_docs
@@ -20,14 +21,15 @@ from kartothek.io_components.metapartition import (
     MetaPartition,
     parse_input_to_metapartition,
 )
-from kartothek.io_components.read import dispatch_metapartitions_from_factory
+from kartothek.io_components.read import (
+    dispatch_metapartitions,
+    dispatch_metapartitions_from_factory,
+)
 from kartothek.io_components.utils import normalize_args, raise_if_indices_overlap
 from kartothek.io_components.write import (
     raise_if_dataset_exists,
     store_dataset_from_partitions,
 )
-from kartothek.io_components.read import dispatch_metapartitions
-from toolz.itertoolz import random_sample
 
 
 def _store_dataset_from_partitions_flat(mpss, *args, **kwargs):
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -26,6 +26,8 @@ from kartothek.io_components.write import (
     raise_if_dataset_exists,
     store_dataset_from_partitions,
 )
+from kartothek.io_components.read import dispatch_metapartitions
+from toolz.itertoolz import random_sample
 
 
 def _store_dataset_from_partitions_flat(mpss, *args, **kwargs):
@@ -288,3 +290,25 @@ def build_dataset_indices__bag(
             update_indices_from_partitions, dataset_metadata_factory=ds_factory
         )
     )
+
+
+def collect_dataset_statistics(store, dataset_uuid, table_name, predicates=None, sample_size=0.3):
+    """
+    WIP: Retrieve some statistics, maybe even plot them, generate a report...?
+    """
+    # TODO not sure if this is the right place for this function
+
+    mp_iterator = dispatch_metapartitions(dataset_uuid=dataset_uuid, store=store, predicates=predicates)
+    bag = db.from_sequence(random_sample(sample_size, mp_iterator))
+
+    def _get_parquet_stats(mp, store=store, table_name=table_name):
+        """
+        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
+        """
+        # TODO include more calculations here?
+        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
+        pq_metadata["rows_per_row_group"] = pq_metadata.["num_rows"] / pq_metadata["num_row_groups"]
+        return pq_metadata
+
+    df_stats = bag.map(_get_parquet_stats, store=store, table_name=table_name).to_dataframe().compute()
+    return df_stats
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -292,25 +292,3 @@ def build_dataset_indices__bag(
             update_indices_from_partitions, dataset_metadata_factory=ds_factory
         )
     )
-
-
-def collect_dataset_statistics(store, dataset_uuid, table_name, predicates=None, sample_size=0.3):
-    """
-    WIP: Retrieve some statistics, maybe even plot them, generate a report...?
-    """
-    # TODO not sure if this is the right place for this function
-
-    mp_iterator = dispatch_metapartitions(dataset_uuid=dataset_uuid, store=store, predicates=predicates)
-    bag = db.from_sequence(random_sample(sample_size, mp_iterator))
-
-    def _get_parquet_stats(mp, store=store, table_name=table_name):
-        """
-        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
-        """
-        # TODO include more calculations here?
-        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
-        pq_metadata["rows_per_row_group"] = pq_metadata.["num_rows"] / pq_metadata["num_row_groups"]
-        return pq_metadata
-
-    df_stats = bag.map(_get_parquet_stats, store=store, table_name=table_name).to_dataframe().compute()
-    return df_stats
--- a/kartothek/io/dask/bag.py
+++ b/kartothek/io/dask/bag.py
@@ -3,7 +3,6 @@ from collections import defaultdict
 from functools import partial
 
 import dask.bag as db
-from toolz.itertoolz import random_sample
 
 from kartothek.core import naming
 from kartothek.core.docs import default_docs
@@ -21,10 +20,7 @@ from kartothek.io_components.metapartition import (
     MetaPartition,
     parse_input_to_metapartition,
 )
-from kartothek.io_components.read import (
-    dispatch_metapartitions,
-    dispatch_metapartitions_from_factory,
-)
+from kartothek.io_components.read import dispatch_metapartitions_from_factory
 from kartothek.io_components.utils import normalize_args, raise_if_indices_overlap
 from kartothek.io_components.write import (
     raise_if_dataset_exists,
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,6 +1,7 @@
 import dask
 import dask.dataframe as dd
 import numpy as np
+from toolz.itertoolz import random_sample
 
 from kartothek.core.common_metadata import empty_dataframe_from_schema
 from kartothek.core.docs import default_docs
@@ -10,6 +11,7 @@ from kartothek.io_components.metapartition import (
     SINGLE_TABLE,
     parse_input_to_metapartition,
 )
+from kartothek.io_components.read import dispatch_metapartitions
 from kartothek.io_components.update import update_dataset_from_partitions
 from kartothek.io_components.utils import (
     _ensure_compatible_indices,
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
         metadata=metadata,
         metadata_merger=metadata_merger,
     )
+
+
+def collect_dataset_statistics(
+    store, dataset_uuid, table_name, predicates=None, frac=0.3
+):
+    """
+    Retrieve some statistics, maybe even plot them, generate a report...?
+
+    Parameters
+    ----------
+    store
+      A factory function providing a KeyValueStore
+    dataset_uuid
+      The dataset's unique identifier
+    table_name
+      Name of the kartothek table for which to retrieve the statistics
+    predicates
+      Kartothek predicates to apply filters on the data for which to gather statistics
+    frac
+      Fraction of the data to use for gathering statistics. `frac == 1` will use the whole dataset.
+
+    Returns
+    -------
+    A dask.dataframe containing information about dataset statistics.
+
+    """
+
+    def _get_parquet_stats(mp, store=store, table_name=table_name):
+        """
+        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
+        """
+        # TODO include more calculations here?
+        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
+        return pq_metadata
+
+    mp_iterator = dispatch_metapartitions(
+        dataset_uuid=dataset_uuid, store=store, predicates=predicates
+    )
+    dfs = [dask.delayed(_get_parquet_stats)(mp) for mp in mp_iterator]
+    df = dd.from_delayed(dfs)
+    df = df.compute()
+    df.reset_index(inplace=True, drop=True)
+
+    return df
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -7,6 +7,7 @@ import numpy as np
 import pandas as pd
 from simplekv import KeyValueStore
 
+from kartothek.core._compat import ARROW_LARGER_EQ_0141
 from kartothek.core.common_metadata import empty_dataframe_from_schema
 from kartothek.core.docs import default_docs
 from kartothek.core.factory import DatasetFactory, _ensure_factory
@@ -368,6 +369,8 @@ def collect_dataset_metadata(
       If no metadata could be retrieved, raise an error.
 
     """
+    if not ARROW_LARGER_EQ_0141:
+        raise RuntimeError("This function requires `pyarrow>=0.14.1`.")
     if not 0.0 < frac <= 1.0:
         raise ValueError(
             f"Invalid value for parameter `frac`: {frac}."
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,13 +1,15 @@
 import random
+from typing import Callable, Optional
 
 import dask
 import dask.dataframe as dd
 import numpy as np
 import pandas as pd
+from simplekv import KeyValueStore
 
 from kartothek.core.common_metadata import empty_dataframe_from_schema
 from kartothek.core.docs import default_docs
-from kartothek.core.factory import _ensure_factory
+from kartothek.core.factory import DatasetFactory, _ensure_factory
 from kartothek.core.naming import DEFAULT_METADATA_VERSION
 from kartothek.io_components.metapartition import (
     _METADATA_SCHEMA,
@@ -24,6 +26,7 @@ from kartothek.io_components.utils import (
     normalize_args,
     validate_partition_keys,
 )
+from kartothek.serialization import PredicatesType
 
 from ._update import update_dask_partitions_one_to_one, update_dask_partitions_shuffle
 from ._utils import _maybe_get_categoricals_from_index
@@ -310,13 +313,13 @@ def update_dataset_from_ddf(
 
 
 def collect_dataset_metadata(
-    store,
-    dataset_uuid,
-    table_name=SINGLE_TABLE,
-    predicates=None,
-    frac=1.0,
-    factory=None,
-):
+    store: Optional[Callable[[], KeyValueStore]] = None,
+    dataset_uuid: Optional[str] = None,
+    table_name: str = SINGLE_TABLE,
+    predicates: Optional[PredicatesType] = None,
+    frac: float = 1.0,
+    factory: Optional[DatasetFactory] = None,
+) -> dd.DataFrame:
     """
     Collect parquet metadata of the dataset. The `frac` parameter can be used to select a subset of the data.
 
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,6 +1,5 @@
 import logging
 import random
-import warnings
 
 import dask
 import dask.dataframe as dd
@@ -12,6 +11,7 @@ from kartothek.core.docs import default_docs
 from kartothek.core.factory import _ensure_factory
 from kartothek.core.naming import DEFAULT_METADATA_VERSION
 from kartothek.io_components.metapartition import (
+    _METADATA_SCHEMA,
     SINGLE_TABLE,
     MetaPartition,
     parse_input_to_metapartition,
@@ -352,7 +352,7 @@ def collect_dataset_metadata(
 
     Returns
     -------
-    A pandas.DataFrame containing the following information about dataset statistics:
+    A delayed pandas.DataFrame containing the following information about dataset statistics:
        * `partition_label`: File name of the parquet file, unique to each physical partition.
        * `row_group_id`: Index of the row groups within one parquet file.
        * `row_group_byte_size`: Byte size of the data within one row group.
@@ -395,15 +395,9 @@ def collect_dataset_metadata(
                 for mp in mps
             ]
         )
-        df = df.compute()
-        df.sort_values(
-            by=["partition_label", "row_group_id"], ignore_index=True, inplace=True
-        )
     else:
-        warnings.warn(
-            f"Can't retrieve metadata for empty dataset with uuid `{dataset_uuid}`"
-        )
-        df = pd.DataFrame(columns=MetaPartition._METADATA_COLUMNS)
-        df = df.astype(MetaPartition._METADATA_SCHEMA._asdict())
+        df = pd.DataFrame(columns=_METADATA_SCHEMA.keys())
+        df = df.astype(_METADATA_SCHEMA)
+        df = dask.delayed(df)
 
     return df
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -313,7 +313,7 @@ def update_dataset_from_ddf(
 
 
 def collect_dataset_metadata(
-    store_factory,
+    store,
     dataset_uuid,
     table_name=SINGLE_TABLE,
     predicates=None,
@@ -332,7 +332,7 @@ def collect_dataset_metadata(
 
     Parameters
     ----------
-    store_factory
+    store
       A factory function providing a KeyValueStore
     dataset_uuid
       The dataset's unique identifier
@@ -374,7 +374,7 @@ def collect_dataset_metadata(
         )
     dataset_factory = _ensure_factory(
         dataset_uuid=dataset_uuid,
-        store=store_factory,
+        store=store,
         factory=factory,
         load_dataset_metadata=False,
     )
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,4 +1,3 @@
-import logging
 import random
 
 import dask
@@ -29,8 +28,6 @@ from kartothek.io_components.utils import (
 from ._update import update_dask_partitions_one_to_one, update_dask_partitions_shuffle
 from ._utils import _maybe_get_categoricals_from_index
 from .delayed import read_table_as_delayed
-
-_logger = logging.getLogger()
 
 
 @default_docs
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -9,9 +9,10 @@ from kartothek.core.factory import _ensure_factory
 from kartothek.core.naming import DEFAULT_METADATA_VERSION
 from kartothek.io_components.metapartition import (
     SINGLE_TABLE,
+    MetaPartition,
     parse_input_to_metapartition,
 )
-from kartothek.io_components.read import dispatch_metapartitions
+from kartothek.io_components.read import dispatch_metapartitions_from_factory
 from kartothek.io_components.update import update_dataset_from_partitions
 from kartothek.io_components.utils import (
     _ensure_compatible_indices,
@@ -306,7 +307,7 @@ def update_dataset_from_ddf(
 
 
 def collect_dataset_statistics(
-    store, dataset_uuid, table_name, predicates=None, frac=0.3
+    store, dataset_uuid, table_name, predicates=None, frac=0.3, factory=None,
 ):
     """
     Retrieve some statistics, maybe even plot them, generate a report...?
@@ -321,29 +322,48 @@ def collect_dataset_statistics(
       Name of the kartothek table for which to retrieve the statistics
     predicates
       Kartothek predicates to apply filters on the data for which to gather statistics
+
+      .. warning::
+          Filtering will only be applied for predicates on indices.
+          The evaluation of the predicates therefore will therefore only return an approximate result.
+
     frac
       Fraction of the data to use for gathering statistics. `frac == 1` will use the whole dataset.
+    factory
+       A DatasetFactory holding the store and UUID to the source dataset.
 
     Returns
     -------
     A dask.dataframe containing information about dataset statistics.
 
+    Raises
+    ------
+    ValueError
+      If no metadata could be retrieved, raise an error.
+
     """
-
-    def _get_parquet_stats(mp, store=store, table_name=table_name):
-        """
-        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
-        """
-        # TODO include more calculations here?
-        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
-        return pq_metadata
-
-    mp_iterator = dispatch_metapartitions(
-        dataset_uuid=dataset_uuid, store=store, predicates=predicates
+    dataset_factory = _ensure_factory(
+        dataset_uuid=dataset_uuid,
+        store=store,
+        factory=factory,
+        load_dataset_metadata=False,
     )
-    dfs = [dask.delayed(_get_parquet_stats)(mp) for mp in mp_iterator]
+    mp_iterator = random_sample(
+        frac,
+        dispatch_metapartitions_from_factory(dataset_factory, predicates=predicates),
+    )
+    dfs = [
+        dask.delayed(MetaPartition.get_parquet_metadata)(
+            mp, store=dataset_factory.store_factory, table_name=table_name
+        )
+        for mp in mp_iterator
+    ]
+    if not dfs:
+        raise ValueError(
+            f"Could not retrieve metadata for dataset {dataset_uuid}. Perhaps the value {frac} for the parameter `frac` is too small?"
+        )
+
     df = dd.from_delayed(dfs)
     df = df.compute()
     df.reset_index(inplace=True, drop=True)
-
     return df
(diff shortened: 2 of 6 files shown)


tests/io_components/test_metapartition.py
```diff
@@ -1645,3 +1645,32 @@ def test_parse_input_schema_formats():
...
+    pq_metadata = meta_partition.get_parquet_metadata(store=store, table_name="core")
+    assert pq_metadata["number_rows"] == 0
+    assert pq_metadata["number_row_groups"] == 1
+
+
+def test_get_parquet_metadata_empty_metapartition(store):
+    # TODO not sure if this makes sense
+    mp = MetaPartition(label="test_label")
+    meta_partition = mp.store_dataframes(store=store, dataset_uuid="dataset_uuid",)
+    pq_metadata = meta_partition.get_parquet_metadata(store=store, table_name="core")
```
NeroCorleone, line 1672: Assertion missing here.

I am wondering: what is the difference between 
```
mp = MetaPartition(label="some_label", data={"core": pd.DataFrame()},) # empty df
```
and 
```
mp = MetaPartition(label="test_label") # no data at all
```
? And can it happen that `dispatch_metapartitions` return a MetaPartition with no data at all (not empty df)?

fjetter: the dispatch should return MPs without data. That's translates into "data was not yet loaded". If there is a dict but it is empty it means "data is loaded but it is empty"

kartothek/io/dask/bag.py
```diff
@@ -288,3 +292,25 @@ def build_dataset_indices__bag(
             update_indices_from_partitions, dataset_metadata_factory=ds_factory
         )
     )
+
+
+def collect_dataset_statistics(store, dataset_uuid, table_name, predicates=None, sample_size=0.3):
```
fjetter, line 297: docsting

fjetter: I would call the `sample_size` `frac` leaning onto https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html

kartothek/io/dask/bag.py
```diff
@@ -288,3 +292,25 @@ def build_dataset_indices__bag(
...
+    """
+    WIP: Retrieve some statistics, maybe even plot them, generate a report...?
+    """
+    # TODO not sure if this is the right place for this function
```
fjetter, line 301: It returns a dataframe, so technically speaking this should live in the `io.dask.dataframe` module by current convention

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1549,32 @@ def delete_from_store(self, dataset_uuid, store):
...
+                "number_row_groups": pq_metadata.num_row_groups,
+                "serialized_size": pq_metadata.serialized_size,
+            }
+            for rg_nb in range(stats["number_row_groups"]):
+                stats[f"row_group_{rg_nb}_byte_size"] = pq_metadata.row_group(
+                    rg_nb
+                ).total_byte_size
+            return stats
+        else:
+            LOGGER.info("No files in MetaPartition, no metadata to track.")
```
fjetter, line 1575: This can not really happen. You may raise a runtime error

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1549,32 @@ def delete_from_store(self, dataset_uuid, store):
...
+        if self.files:
+            fd = store.open(self.files[table_name])
+            pq_metadata = pa.parquet.ParquetFile(fd).metadata
+            stats = {
+                "number_rows": pq_metadata.num_rows,
+                "number_row_groups": pq_metadata.num_row_groups,
+                "serialized_size": pq_metadata.serialized_size,
+            }
+            for rg_nb in range(stats["number_row_groups"]):
+                stats[f"row_group_{rg_nb}_byte_size"] = pq_metadata.row_group(
```
fjetter, line 1570: This will probably not work. It is highly likely that the files in a dataset have different numbers of rowgroups. This would then return you a different schema for every partition. If you want to expose the RG information you might need to blow up the data, e.g.

| partition label | number of rows | number of row groups | row group ID |  row group byte size | 
| - |  - | - | - |  - | 
| "label" |  10000 | 2 | 1 |  5 | 
| "label" |  10000 | 2 | 2 |  20 | 

wherre the partition-level rows are duplicated

The primary key of this table is then `("partition label", "row group ID")`

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1549,32 @@ def delete_from_store(self, dataset_uuid, store):
...
+
+        # TODO docstrings
+        """
+        if not isinstance(table_name, str):
+            raise TypeError("Expecting a string for parameter `table_name`.")
+
+        if self.files:
+            fd = store.open(self.files[table_name])
+            pq_metadata = pa.parquet.ParquetFile(fd).metadata
+            stats = {
```
fjetter, line 1564: @xhochy is there, by any chance, an arrow functionality for this kind of thing?

tests/io_components/test_metapartition.py
```diff
@@ -1645,3 +1645,32 @@ def test_parse_input_schema_formats():
...
+
+
+# TODO test for row group size > 1
+# TODO add test for one empty metapartition
```
fjetter, line 1676: this can not happen, the dispatch always provides files, otherwise it doesn't dispatch anything. just raise an exception in MetaPartition

tests/io_components/test_metapartition.py
```diff
@@ -1645,3 +1645,32 @@ def test_parse_input_schema_formats():
...
+
+
+def test_get_parquet_metadata_empty_metapartition(store):
+    # TODO not sure if this makes sense
+    mp = MetaPartition(label="test_label")
+    meta_partition = mp.store_dataframes(store=store, dataset_uuid="dataset_uuid",)
+    pq_metadata = meta_partition.get_parquet_metadata(store=store, table_name="core")
+
+
+# TODO test for row group size > 1
```
fjetter, line 1675: This is where the schema differences between the partitions will raise

kartothek/io_components/metapartition.py
fjetter: What I mean is `ParquetFile` -> Metadata in sane format for analytics

kartothek/io_components/metapartition.py
NeroCorleone: Makes sense. I'll keep the row group information and blow up the data.

tests/io/dask/dataframe/test_stats.py
```diff
@@ -0,0 +1,49 @@
...
+
+    expected = pd.DataFrame(
+        data={
+            "partition_label": ["cluster_1"],
+            "row_group_id": [0],
+            "number_rows": [1],
+            "number_row_groups": [1],
+        }
+    )
+    # TODO this fails, might be an issue with the predicate passing to the metapartitions in `dispatch_metapartitions`?
```
NeroCorleone, line 48: I would have expected that`mp_iterator = dispatch_metapartitions(dataset_uuid=dataset_uuid, store=store, predicates=predicates)` only returns those metapartitions with `P == 1`, but that doesn't seem to be the case (iterator contains all metapartitions, from https://github.com/JDASoftwareGroup/kartothek/blob/master/tests/conftest.py#L197).

NeroCorleone: Or fo we evaluate predicates in the `dispatch_metapartitions` functionality only on the index level? That is at least how I read https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/io_components/read.py#L90-L100

fjetter: This is only evaluated on index level

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
         metadata=metadata,
         metadata_merger=metadata_merger,
     )
+
+
+def collect_dataset_statistics(
+    store, dataset_uuid, table_name, predicates=None, frac=0.3
```
fjetter, line 309: can you add a `factory` as a kwarg, please? we started to move the interface towards `kartothek.core.factory.DatasetFactory` a while ago for reading pipelines (eliminates the need to load metadata and indices multiple times)

You cann then also use `dispatch_metapartitiosn_from_factory` together with a utils function `ensure_factory` or similar (see other read implementations

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
...
+    Parameters
+    ----------
+    store
+      A factory function providing a KeyValueStore
+    dataset_uuid
+      The dataset's unique identifier
+    table_name
+      Name of the kartothek table for which to retrieve the statistics
+    predicates
+      Kartothek predicates to apply filters on the data for which to gather statistics
```
fjetter, line 323: This should include a warning that the predicates are not applied exactly but only on index level. Whatever is returned by this function is therefore only an approximation

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
...
+    frac
+      Fraction of the data to use for gathering statistics. `frac == 1` will use the whole dataset.
+
+    Returns
+    -------
+    A dask.dataframe containing information about dataset statistics.
+
+    """
+
+    def _get_parquet_stats(mp, store=store, table_name=table_name):
```
fjetter, line 333: Please use global scope to define the function, i.e. outside of the function context. This is more efficient for serialization

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
...
+        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
+        """
+        # TODO include more calculations here?
+        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
+        return pq_metadata
+
+    mp_iterator = dispatch_metapartitions(
+        dataset_uuid=dataset_uuid, store=store, predicates=predicates
+    )
+    dfs = [dask.delayed(_get_parquet_stats)(mp) for mp in mp_iterator]
```
fjetter, line 344: You can probably just call

`dask.delayed(MetaPartition.get_parquet_metadata, store=store, table_name=table_name)`

without the need for another function definition

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +303,47 @@ def update_dataset_from_ddf(
...
+
+    def _get_parquet_stats(mp, store=store, table_name=table_name):
+        """
+        WIP: Retrieve plain parquet metadata, derive some interesting numbers from them
+        """
+        # TODO include more calculations here?
+        pq_metadata = mp.get_parquet_metadata(store=store, table_name=table_name)
+        return pq_metadata
+
+    mp_iterator = dispatch_metapartitions(
```
fjetter, line 341: The sampling isn't implemented at the moment, is it? We could probably already push this down into `dispatch_metapartitions`. This may be useful for other applications as well 🤔

tests/io/dask/dataframe/test_stats.py
NeroCorleone: Fixed.

kartothek/io/dask/dataframe.py
NeroCorleone: Just added it as a kwarg for now, didn't want to remove the `store` and `dataset_uuid` parameter as default for now.

kartothek/io/dask/dataframe.py
NeroCorleone: Done

kartothek/io/dask/dataframe.py
NeroCorleone: Removed it entirely in favor of your proposed solution.

kartothek/io/dask/dataframe.py
NeroCorleone: ... why not make it easier, right? 😁

kartothek/io/dask/dataframe.py
NeroCorleone: Yes, apparently I excluded it for testing reasons and forgot to put it back in. Fixed for now.

NeroCorleone: We probably could push that down to the dispatch function, do you have another use case in mind already?

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1549,67 @@ def delete_from_store(self, dataset_uuid, store):
...
+          A DataFrame with relevant parquet metadata
+        """
+        if not isinstance(table_name, str):
+            raise TypeError("Expecting a string for parameter `table_name`.")
+
+        if callable(store):
+            store = store()
+
+        if self.files:
+            fd = store.open(self.files[table_name])
```
marco-neumann-jdas, line 1576: File descriptors (even though they're likely virtual here) should always be guarded:

```suggestion
            with store.open(self.files[table_name]) as fd:
```

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +304,66 @@ def update_dataset_from_ddf(
...
+      Name of the kartothek table for which to retrieve the statistics
+    predicates
+      Kartothek predicates to apply filters on the data for which to gather statistics
+
+      .. warning::
+          Filtering will only be applied for predicates on indices.
+          The evaluation of the predicates therefore will therefore only return an approximate result.
+
+    frac
+      Fraction of the data to use for gathering statistics. `frac == 1` will use the whole dataset.
```
marco-neumann-jdas, line 331: That's slightly imprecise. It's the fraction of partitions, which has two implications:

- for very few partitions the effect of the fraction parameter is quite coarse
- for a skewed distribution (size over partitions) sampling large partitions in and out might effect the result more than for even distributions, so the user should be aware of that

Also: is the returned data what was collected or do you extrapolate based on the fraction (e.g. for `frac=0.1`, does the user get roughly 10% of the true dataset size or do you calculate `collected_size / frac` and return that)

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +304,66 @@ def update_dataset_from_ddf(
         metadata=metadata,
         metadata_merger=metadata_merger,
     )
+
+
+def collect_dataset_statistics(
+    store, dataset_uuid, table_name, predicates=None, frac=0.3, factory=None,
+):
+    """
+    Retrieve some statistics, maybe even plot them, generate a report...?
```
marco-neumann-jdas, line 313: Just return the data would be enough. The user can do with that whatever they want (or the query planner).

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +304,66 @@ def update_dataset_from_ddf(
...
+          The evaluation of the predicates therefore will therefore only return an approximate result.
+
+    frac
+      Fraction of the data to use for gathering statistics. `frac == 1` will use the whole dataset.
+    factory
+       A DatasetFactory holding the store and UUID to the source dataset.
+
+    Returns
+    -------
+    A dask.dataframe containing information about dataset statistics.
```
marco-neumann-jdas, line 337: Please describe all columns that you're returning.

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +304,66 @@ def update_dataset_from_ddf(
...
+    )
+    dfs = [
+        dask.delayed(MetaPartition.get_parquet_metadata)(
+            mp, store=dataset_factory.store_factory, table_name=table_name
+        )
+        for mp in mp_iterator
+    ]
+    if not dfs:
+        raise ValueError(
+            f"Could not retrieve metadata for dataset {dataset_uuid}. Perhaps the value {frac} for the parameter `frac` is too small?"
```
marco-neumann-jdas, line 363: Or the dataset is empty? I think guesses in the exception message are not a really good UX.

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1549,67 @@ def delete_from_store(self, dataset_uuid, store):
...
+                    "number_rows_total",
+                    "number_row_groups",
+                    "serialized_size",
+                    "number_rows_per_row_group",
+                ],
+            )
+
+        else:
+            raise RuntimeError(
+                "This should not happen: MetaPartition with no data associated."
```
marco-neumann-jdas, line 1610: It can: if the metapartition was created from data (DF) and was not preserved/stored yet.

fjetter: Theoretically yes. In this particular workflow not and in any case, this exception should be fine. Maybe change the wording / remove the "this should not happen" thing. At least for the workflows where this is / should be used in this _should_ not happen

kartothek/io/dask/dataframe.py
fjetter: Regarding the normalization of the result you mentioned we also had this discussion and, quite frankly, I am not sure what the better API is. Theoretically everything is available for the user to normalize the data him/herself and with the implications / risks you already mentioned regarding too small sample sizes / inhomogeneous partitions, we figured that we shouldn't normalize it here.
otoh, normalizing it is currently a bit awkward since somewhere we need to get the total number of partitions which is available but not easily.

Input / suggestions welcome

kartothek/io/dask/dataframe.py
fjetter: > Yes, apparently I excluded it for testing reasons and forgot to put it back in. Fixed for now.

Either way, you should proabably add a test for it ;)

> do you have another use case in mind already?
Sampling actual data would be also a use case. Essentially https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.sample but the sampling would be on IO layer

kartothek/io/dask/dataframe.py
fjetter: I could make a point that `frac>0` should never return an empty value, i.e. we should always "round up" / ensure there is always at least one partition. That's probably not that easy with the itertoolz sampler but would also not be overly complicated to achieve. (One could make the opposite argument as well)

Opinions? Suggestions? As long as the behaviour is not surprising / documented, I'm fine either way

kartothek/io/dask/dataframe.py
NeroCorleone: >  It's the fraction of partitions, which has two implications:
Good point, both implications. I think, I should extend the docstring a bit more.

I have thought of the interface as "Please give me metadata about `frac` % of my data" -- so no normalization. Perhaps naming this whole thing `collect_statistics` is misleading... We are not really doing any statistics (beyond the sampling and calculating average number of rows per row group) and essentially offer the user a possibility to easily collect and view metadata.

kartothek/io/dask/dataframe.py
NeroCorleone: I agree, returning empty data here feels weird.
I guess the intention was saying something like "Hey user, your choice for `frac` is really off"...

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +304,66 @@ def update_dataset_from_ddf(
...
+
+def collect_dataset_statistics(
+    store, dataset_uuid, table_name, predicates=None, frac=0.3, factory=None,
+):
+    """
+    Retrieve some statistics, maybe even plot them, generate a report...?
+
+    Parameters
+    ----------
+    store
```
NeroCorleone, line 317: - [ ] rename to factory function

kartothek/io/dask/dataframe.py
NeroCorleone: - [ ] Adjust docstring: only df

kartothek/io/dask/dataframe.py
marco-neumann-jdas: Since you're likely returning a row per partition, I think it is the better API to NOT normalize here 👍 

Regarding the name: I think "statistics" is OK, at least I cannot think of anything better.

kartothek/io/dask/dataframe.py
marco-neumann-jdas: So even if we agree on rounding up and rejecting `frac == 0`, `dfs` might still be empty in case of an empty dataset. And an empty dataset is valid data. It's an edge case for sure but asking the user to guard each and every call to this function with `if len(ds.partitions) == 0` is not a very usable API and I think we'll run into bugs all over the place (even in our own code) when we do that.

So what I would rather propose:

- raise if `frac` is not in `(0, 1]` (half closed interval)
- drop the `if not dfs` check and write a test case for an empty DS. You'll likely need to tell dask the dtype for each and every column (because it cannot infer it, which is suboptimal anyways and requires it to eagerly load+test the first partition)

kartothek/io_components/metapartition.py
marco-neumann-jdas: 👍 for changing the wording to something like "MetaPartition.get_parquet_metadata requires the MetaPartition to have files. If you have in-memory data, you must store it first".

kartothek/io/dask/dataframe.py
lr4d: AFAI see Parquet calls what we're accessing here "metadata", see https://parquet.apache.org/documentation/latest/

kartothek/io/dask/dataframe.py
NeroCorleone: +1 for checking the input for`frac`.

Just to make sure, I understood your proposal: You are in favor of *not* raising for an empty dataset, but instead returning an empty result df and maybe issue a warning?

marco-neumann-jdas: > Just to make sure, I understood your proposal: You are in favor of not raising for an empty dataset, but instead returning an empty result df and maybe issue a warning?

yes.

tests/io/dask/dataframe/test_stats.py
```diff
@@ -0,0 +1,291 @@
...
+        store=store_factory,
+        dataset_uuid="dataset_uuid",
+        dfs=[{"table1": df1, "table2": df2}],
+        partition_on=["A"],
+    )
+
+    # df_stats = collect_dataset_metadata(
+    #     store_factory=store_factory, dataset_uuid="dataset_uuid", table_name="table2",
+    # )
+    # TODO assertions
```
NeroCorleone, line 273: - [ ] Assertion missing

tests/io/dask/dataframe/test_stats.py
```diff
@@ -0,0 +1,291 @@
...
+            table_name="table",
+        )
+    expected = pd.DataFrame(columns=METADATA_COLUMNS)
+    expected = expected.astype(dict(zip(METADATA_COLUMNS, METADATA_DTYPES)))
+    pd.testing.assert_frame_equal(expected, df_stats)
+
+
+def test_collect_dataset_metadata_table_without_partition(store_factory):
+    """
+    df2 doesn't have files for all partition (specificaly `A==2`).
```
NeroCorleone, line 257: - [ ] typo

kartothek/io_components/metapartition.py
```diff
@@ -183,6 +183,29 @@ class MetaPartition(Iterable):
...
+        np.dtype(int),
+        np.dtype(int),
+        np.dtype(int),
+        np.dtype(int),
+        np.dtype(int),
+        np.dtype(int),
+    )
+
+    MetadataSchema = namedtuple("MetadataSchema", _METADATA_COLUMNS)  # type: ignore
+    _METADATA_SCHEMA = MetadataSchema(**dict(zip(_METADATA_COLUMNS, _METADATA_DTYPES)))  # type: ignore
```
NeroCorleone, line 207: Idea here was to store the "schema" of the returned statistics as something immutable, preferably something like a `frozendict` mapping column names to the dtypes.
Not sure, if this constructions is the best way to solve that.

kartothek/io/dask/dataframe.py
lr4d: my bad, 

- [ ] revert this for consistency reasons

tests/io/dask/dataframe/test_stats.py
lr4d:
```suggestion
    df2 doesn't have files for all partition (specifically `A==2`).
```

kartothek/io_components/metapartition.py
lr4d: This error was removed as the logic was changed to take into account partitions that don't have files for the table queried

kartothek/io_components/metapartition.py
fjetter: I wouldn't go for frozen(anything) since these containers are not as well supported as the default containers. For instance, I've seen occasionally serialization errors when using frozensets. What would be the value of defining this as frozen? I consider the risk of a user mutating this structure as *very* low and I would prefer the reduced complexity of not using special containers

kartothek/io_components/metapartition.py
```diff
@@ -183,6 +183,29 @@ class MetaPartition(Iterable):
     about the parent dataset
     """
 
+    _METADATA_COLUMNS = (
```
fjetter, line 186: Can you please define this as one dict instead? having two tuples makes the whole thing hard to read

kartothek/io/dask/dataframe.py
NeroCorleone: Should be fixed: sampling is now done differently and rounding up to at least one metapartition is guarateed.
Renamed to `collect_metadata` and no normalization.

- [ ] Is there a test for the rounding up?

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1572,58 @@ def delete_from_store(self, dataset_uuid, store):
...
+                *[
+                    (
+                        pq_metadata.row_group(rg_id).total_byte_size,
+                        pq_metadata.row_group(rg_id).num_rows,
+                    )
+                    for rg_id in data["row_group_id"]
+                ]
+            )
+            df = pd.DataFrame(data=data, columns=self._METADATA_COLUMNS,)
+            df.astype(MetaPartition._METADATA_SCHEMA._asdict())
```
fjetter, line 1619: the `_asdict` looks private. isn't `dict(...)` the same thing?

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1572,58 @@ def delete_from_store(self, dataset_uuid, store):
...
+
+            data["row_group_byte_size"], data["number_rows_per_row_group"] = zip(
+                *[
+                    (
+                        pq_metadata.row_group(rg_id).total_byte_size,
+                        pq_metadata.row_group(rg_id).num_rows,
+                    )
+                    for rg_id in data["row_group_id"]
+                ]
+            )
```
fjetter, line 1617: tbh, I would prefer this to be expressed a bit simpler. At least for me it is very hard to read and understand what a "zip of list of tuples" actually generates. Why not

```suggestion
            data["row_group_byte_size"] = []
            data["number_rows_per_row_group"] = []

            for rg_id in data["row_group_id"]:
                data["row_group_byte_size"].append(
                    pq_metadata.row_group(rg_id).total_byte_size
                )
                data["number_rows_per_row_group"].append(
                    pq_metadata.row_group(rg_id).num_rows
                )
```

That's easier to read and probably generates a smaller diff when changing, i.e. easier to maintain

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1572,58 @@ def delete_from_store(self, dataset_uuid, store):
...
+            df = pd.DataFrame(data=data, columns=self._METADATA_COLUMNS,)
+            df.astype(MetaPartition._METADATA_SCHEMA._asdict())
+
+        else:
```
fjetter, line 1621: You can safe the extra else statement by just initializing `data={}` at the beginning

tests/io/dask/dataframe/test_stats.py
```diff
@@ -0,0 +1,291 @@
...
+from kartothek.io.dask.dataframe import collect_dataset_metadata
+from kartothek.io.eager import (
+    store_dataframes_as_dataset,
+    update_dataset_from_dataframes,
+)
+from kartothek.io_components.metapartition import MetaPartition
+from kartothek.io_components.write import store_dataset_from_partitions
+from kartothek.serialization import ParquetSerializer
+
+METADATA_COLUMNS = (
```
fjetter, line 14: Why redefine this? Why not import it?

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +310,95 @@ def update_dataset_from_ddf(
...
+        mps = mps[:cutoff_index]
+        df = dd.from_delayed(
+            [
+                dask.delayed(MetaPartition.get_parquet_metadata)(
+                    mp, store=dataset_factory.store_factory, table_name=table_name,
+                )
+                for mp in mps
+            ]
+        )
+        df = df.compute()
```
fjetter, line 393: why compute? We usually only pass on the graphs. why do it differently herer?

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +310,95 @@ def update_dataset_from_ddf(
...
+        df.sort_values(
```
fjetter, line 394: Why sorting it here? This should be done by the user, shouldn't it? Sorting on the partition_label, for instance, is not very fast since it's an object col

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +310,95 @@ def update_dataset_from_ddf(
...
+            by=["partition_label", "row_group_id"], ignore_index=True, inplace=True
+        )
+    else:
+        warnings.warn(
+            f"Can't retrieve metadata for empty dataset with uuid `{dataset_uuid}`"
```
fjetter, line 399: Don't necessarily need a warning here, do we? An empty dataset is the only way this function returns an empty dataframe so this should be enough information. No need to log/warn anything is there? I consider this ordinary behaviour.

kartothek/io/dask/dataframe.py
fjetter: I usually try to avoid to use compute internally since compute also triggers optimization and behaves differently depending on whether or not there is a client present. Also, users may want to use `Client.compute(graph)` (lazy execution instead.

kartothek/io_components/metapartition.py
NeroCorleone: No, it isn't (but I don't know how it is different).

fjetter:
FYI: There is also a `ParquetFile.metadata.to_dict()` method (maybe only in most recent versions). This already prints out a dictionary with *all* information available. This would require some reformatting and is a little too verbose for my taste but may be something you want to have a look at

kartothek/io/dask/dataframe.py
NeroCorleone: Jup, I am fine with not sorting, just "felt right" to return the data ordered by the keys

kartothek/io_components/metapartition.py
NeroCorleone: True. Still, I find keeping the else here doesn't add too much overhead and makes it more explicit that an empty dataframe is created in case the `table_name` could not be found.

kartothek/io_components/metapartition.py
NeroCorleone: Reason for constructing it this way was to make sure I can't mess with these constants at runtime.
Sure, the risk of that happening might be low and probably also in this specific use case.

I guess the reason why I did it this way is because I had a bug in the last weeks which was a bit nasty because it manipulated some class variable at runtime. And using an immutable data type felt like the right data type for "This is constant, please don't mess with it", which I don't have when I use a dict.

kartothek/io_components/metapartition.py
lr4d: > To prevent conflicts with field names, the method and attribute names start with an underscore.

Ref: https://docs.python.org/3/library/collections.html

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +307,94 @@ def update_dataset_from_ddf(
...
+          The evaluation of the predicates therefore will therefore only return an approximate result.
+
+    frac
+      Fraction of the total number of partitions to use for gathering statistics. `frac == 1.0` will use all partitions.
+    factory
+       A DatasetFactory holding the store and UUID to the source dataset.
+
+    Returns
+    -------
+    A delayed pandas.DataFrame containing the following information about dataset statistics:
```
lr4d, line 352: This is not consistent with output (currently it's either a dask.DataFrame or dask.delayed object containing a pandas.DataFrame)

kartothek/io/dask/dataframe.py
```diff
@@ -390,15 +395,9 @@ def collect_dataset_metadata(
...
-        )
     else:
-        warnings.warn(
-            f"Can't retrieve metadata for empty dataset with uuid `{dataset_uuid}`"
-        )
-        df = pd.DataFrame(columns=MetaPartition._METADATA_COLUMNS)
-        df = df.astype(MetaPartition._METADATA_SCHEMA._asdict())
+        df = pd.DataFrame(columns=_METADATA_SCHEMA.keys())
+        df = df.astype(_METADATA_SCHEMA)
+        df = dask.delayed(df)
```
fjetter: This should also be a `dask.DataFrame`, not a delayed object, see https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.from_pandas

```suggestion
        df = dd.from_pandas(
            data=df,
            npartitions=1,
        )
```

kartothek/io/dask/dataframe.py
```diff
@@ -301,3 +307,94 @@ def update_dataset_from_ddf(
...
+
+
+def collect_dataset_metadata(
+    store,
+    dataset_uuid,
+    table_name=SINGLE_TABLE,
+    predicates=None,
+    frac=1.0,
+    factory=None,
+):
```
fjetter, line 319: Can you please add type annotations to the signature (including output)

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1559,55 @@ def delete_from_store(self, dataset_uuid, store):
             store.delete(file_key)
         return self.copy(files={}, data={}, metadata={})
 
+    def get_parquet_metadata(self, store, table_name):
```
fjetter, line 1562: Please add type annotations

kartothek/io/dask/dataframe.py
marco-neumann-jdas: Why was this working then?

fjetter: Calling compute on delayed also works, of course. And the result returns a pandas dataframe. so this is _almost_ the same

lr4d: It is. See last commit

kartothek/io_components/metapartition.py
```diff
@@ -1549,6 +1559,55 @@ def delete_from_store(self, dataset_uuid, store):
...
+                "number_row_groups": pq_metadata.num_row_groups,
+                "serialized_size": pq_metadata.serialized_size,
+            }
+            data["row_group_id"] = range(data["number_row_groups"])
+
+            data["row_group_byte_size"] = []
+            data["number_rows_per_row_group"] = []
+            for rg_id in data["row_group_id"]:
+                data["row_group_byte_size"].append(
+                    pq_metadata.row_group(rg_id).num_rows,
```
lr4d, line 1601: - [ ] wrong

marco-neumann-jdas:
@NeroCorleone even the the actual change is OK, could you please also add a changelog entry?

fjetter:
good job @NeroCorleone and @lr4d

PR #308, opened 2020-07-14
Improve performance for index remove partitions

jochen-ott-by:
The current implementation could take half an hour or so for somewhat larger datasets (20k partitions, 1M entries in index, removing half the partitions). The new implementation provides a speedup in this case of about 1000x.

### DIFF
--- a/asv_bench/benchmarks/index.py
+++ b/asv_bench/benchmarks/index.py
@@ -66,6 +66,12 @@ class Index(IndexBase):
     )
     param_names = ["number_values", "number_partitions", "dtype"]
 
+    def time_remove_partitions_inplace(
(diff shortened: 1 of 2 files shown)


fjetter:
I'm fine with the changes. Code is also easier to read now, I'd probably even merge it if it was slower ;)

Apart from this code change, I would recommend to verify that the index is actually valuable enough to keep around. In my experience, indices with so many values often only offer a limited amount of information but that ultimately depends on how the data is distributed. For instance, if every value is in almost every partition, the value offers only limited value for query planing and you might be better off with reading the occasional empty file (table scan w/ miss) instead of dealing with this index.

PR #309, opened 2020-07-16
Move to GH actions

marco-neumann-by:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+
+    steps:
+      # CI setup
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Conda Bootstrap
+        uses: goanpeca/setup-miniconda@v1
+        with:
+          auto-update-conda: true
+          python-version: ${{ matrix.python }}
```
xhochy, line 41:
```suggestion
          python-version: ${{ matrix.python }}
          channels: conda-forge
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+      - name: Conda Config
+        run: |
+          conda config --add channels conda-forge
```
xhochy, line 44:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+          conda config --set channel_priority strict
+          conda config --set always_yes yes
+          conda config --set changeps1 no
+          conda config --set pip_interop_enabled True
+
+      # Installation
```
xhochy, line 50: I'm always tempeted to comment that you shouldn't install into the `base` environment but then remember that the action actually creates and activates a separate environment.
```suggestion
      # Installation into the `test` environment
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+        run: conda install -c arrow-nightlies arrow-cpp pyarrow
+        if: matrix.pyarrow == 'nightly'
+      - name: Pip Instal NumFOCUS nightly
+        shell: bash -l {0}
+        # NumFOCUS nightly wheels, contains numpy and pandas
+        run: env PRE_WHEELS="https://pypi.anaconda.org/scipy-wheels-nightly/simple" pip install --pre --upgrade --timeout=60 -f $PRE_WHEELS pandas numpy
+        if: matrix.numfocus_nightly
+      - name: Pip Install blacken-docs
+        shell: bash -l {0}
+        # currently not on conda
```
xhochy, line 72: Please: https://github.com/conda-forge/staged-recipes/pull/12155

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+        run: pip install --no-deps blacken-docs
+      - name: Pip Install kartothek
+        shell: bash -l {0}
+        run: pip install --no-deps -e .
+      - name: Conda Export
+        shell: bash -l {0}
+        run: conda list --export
+
+      # Linters
```
xhochy, line 81: What about running these using `pre-commit run -a`? I like the output of `pre-commit` for them as it shows the problems for all linters.

codecov.yml
```diff
@@ -0,0 +1,4 @@
+comment: false
```
xhochy, line 1: Can you name this `.codecov.yml` so it doesn't clutter the visible files?

.github/workflows/ci.yml
marco-neumann-jdas: This has multiple issues:

- it doesn't show up nicely in github actions
- it uses yet another venv system
- the venv setup might require retries, and our current setup is broken (the black hook also fixes the files, then fails, then gets retried and then passes 🤦 )

marco-neumann-jdas: Another change: we're now checking ALL files, not just the ones that where touched by the PR. This is esp. important for upgrades (e.g. mypy itself), config changes and changes to "unrelated" files (e.g. changes in one file might produce errors in other files importing/using them).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,151 @@
...
+      # Builds
+      - name: Build Wheel
+        run: python setup.py sdist bdist_wheel
+
+      # Artifacts
+      - name: Preserve Dist
+        uses: actions/upload-artifact@v1
+        with:
+          name: dist
+          path: dist
```
lr4d, line 133: do we want this for all jobs?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: black --check .
+      - name: blacken-docs
+        run: blacken-docs
+      - name: Isort
+        run: isort --recursive --check-only
+
+      # Tests
+      - name: Pytest
+        run: pytest --cov kartothek --cov-report xml
+      - name: ASV
```
fjetter, line 114: This is new. I'm actually not a big fan of this since it adds quite significant compute time to every job and I wouldn't consider these results very reliable anyhow.
If you want to let these run, I'd prefer to set up a dedicated job which just executes benchmarks but we shouldn't do this for every job in the build matrix

.github/workflows/ci.yml
```diff
@@ -0,0 +1,151 @@
...
+        run: |
+          asv --config ./asv_bench/asv.conf.json machine --machine travis --os unknown --arch unknown --cpu unknown --ram unknown
```
fjetter, line 116: machine would not be travis

.github/workflows/ci.yml
```diff
@@ -0,0 +1,151 @@
+name: CI
+
+
+on:
```
fjetter, line 4: Does this mean these jobs will only run for master branch, PRs and tags? What about ordinary branches?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,151 @@
...
+      - name: Pytest
+        run: pytest --cov kartothek --cov-report xml
+      - name: ASV
+        run: |
+          asv --config ./asv_bench/asv.conf.json machine --machine travis --os unknown --arch unknown --cpu unknown --ram unknown
+          asv --config ./asv_bench/asv.conf.json run --show-stderr --environment existing --quick
+      - name: Sphinx
+        run: |
+          python setup.py build_sphinx
+          touch ./docs/_build/html/.nojekyll
```
fjetter, line 121: What's this?

.github/workflows/ci.yml
marco-neumann-jdas: The UI splits this on a per-job level, so they don't overlap. But they should be virtually identical, so we could change to to "master job only" (see `IS_MASTER_JOB`)

.github/workflows/ci.yml
marco-neumann-jdas: > I'm actually not a big fan of this since it adds quite significant compute time to every job and I wouldn't consider these results very reliable anyhow.

They are not reliable (see `--quick`), but we need to check that they're not broken / runnable and IMHO this should be done for every PR. The quick mode should normally be quite fast, the issue with our benchmarks in particular is that they're slow and I think we should fix that (in a later PR, not here).

marco-neumann-jdas: I've changed that to master-job-only (see `IS_MASTER_JOB`). That's way simpler than having a whole new job type and we should rework the benchmarks later. Let me know if that is OK for you.

.github/workflows/ci.yml
marco-neumann-jdas: changed to "all branches, tags starting w/ `v` and PRs (might come from other repos)".

.github/workflows/ci.yml
marco-neumann-jdas: Copy+paste leftover from `rle-array`. **But stay with me:** there is it a marker to signal that the output is not jekyll-driven (and files with `_` are just ordinary files), so that the docs can easily by hosted on GitHub pages. Now the question is: do you want to switch from RTD to GitHub pages? This would also mean that we only need one single build pipeline and don't need to work with yet another build environment.

For reference: [the rle-array docs on GH pages](https://jdasoftwaregroup.github.io/rle-array/).

marco-neumann-jdas: @xhochy 🔝 any opinion on that?

fjetter: Well, I don't have a strong opinion there but if we change the hosting we'd need to change blog posts, etc. as well

marco-neumann-jdas: I can do that if we agree on that. It's just a quick search+replace.

.github/workflows/ci.yml
fjetter: I don't mind executing the for every PR, I agree on this. I also know they are slow and not very well written. Can you open a ticket for this?

I would prefer a dedicate job type but I do understand that this is not easily set up. I believe that's a fair compromise.
IF we wanted to invest in more heterogenous jobs, I also see the opportunity to run docs and linting in dedicated jobs for better / clearer feedback. that's obviously another PR, of course.

.github/workflows/ci.yml
lr4d: I'm just trying to figure out what the value is in keeping all these distribution files

.github/workflows/ci.yml
lr4d: +1 for running this for every PR, I'd rather detect significant performance issues early. If this is not appropriate now, then we should have it in our backlog

fjetter: @lr4d you won't detect performance issues with this. these runs are only to verify that the benchmarks are actually able to run. Setting it up that we can have an actual reporting is much more complex

.github/workflows/ci.yml
fjetter: I'm a bit confused. maybe the UI is playing tricks on me but I still see

```yaml
on:
  push:
     branches:
       - master
```

plus some links to the explanation. My point is, this will disable builds for ordinary branches, doesn't it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,152 @@
+name: CI
+
+
+on:
```
fjetter, line 4: we should also add (or similar daily runs). unfortunately they don't support @daily, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

```yaml
on:
  schedule:
    - cron: '0 0 * * *' 
```

fjetter: I can't find anything in the docs but can we enable notifications for these kind of runs?

.github/workflows/ci.yml
marco-neumann-by: Sorry, I had to revert the change. Yes. it will disable builds on ordinary branches, due to the reason illustrated in the links (you get double builds on PRs from this repo to this repo otherwise). If we wanna have release/backport branches in the future, we need to adapt this filter. I'm not entirely happy with this solution though, but having double builds on every BY PR seems worse.

.github/workflows/ci.yml
marco-neumann-by: I think if you watch the repo, you'll get notified? (ref https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/configuring-notifications#github-actions-notification-options )

.github/workflows/ci.yml
fjetter: thb, this is kind of a blocker for me. I consider letting CI run for branches essential. My preferred choice would be to simply *not create PRs from this repo* but rather use a fork instead.

marco-neumann-by: Mmh, so there is a difference between `on.branch` and `on.pull_request`: The first runs on the tip of the branch, the latter on the PR merge result (see https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout ). If you rebase before merge, both are normally identical (or at least close enough). If the branch is somewhat stale, they are NOT identical and can also fail idependently. I think we now have the following requirements:

- running CI on non-PR branches (or before filing a --- potentially WIP --- PR)
- having the PR CI as close on the merge result is possible

With these two requirements, I don't see how we can workaround the doubled job issue. Would you be OK with that?

Side-note: this is also the normal behavior if a merge-train / merge-bot like [bors](https://bors.tech/) would be employed.

fjetter: I think I'm fine with this. This just makes the "problem" of testing too much more pressing. We desperately need to drop arrow versions...

I think that's actually a pretty cool functionality to get for free. It is annoying that it cannot be easily turned off, though.

marco-neumann-by: done

fjetter:
I think this is a very good start. If something is still off, we can reiterate.

Thanks @marco-neumann-by  for taking care of this!

lr4d:
Yeah, thanks @marco-neumann-by . Great job.

Issue #269, opened 2020-03-31
Make store_dataset pipelines use corresponding update_dataset pipeline with overwrite check


lr4d:
Complexity reduction

fjetter:
@lr4d Please add a more thorough description. It's not clear for me what this issue intends to cover

PR #316, opened 2020-07-21
make `store___iter` use `update__iter`, small re-factoring

lr4d:
Small re-factoring to decrease the amount of copied code.

In particular, `store_dataframes_as_dataset__iter` now uses `update_dataset_from_dataframes__iter` under the hood.
Also added `factory` as an optional kwarg to `store_dataframes_as_dataset__iter`.

This should also make it easier to implement `store_dataset_from_ddf` in the future thanks to the function `check_dataset_creation_constraints`

### DIFF
--- a/kartothek/core/factory.py
+++ b/kartothek/core/factory.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Callable, Optional, TypeVar, cast
 
 from kartothek.core.dataset import DatasetMetadata, DatasetMetadataBase
 from kartothek.core.utils import _check_callable
+from kartothek.core.uuid import gen_uuid
 
 if TYPE_CHECKING:
     from simplekv import KeyValueStore
@@ -16,10 +17,17 @@ def _ensure_factory(
     factory: Optional["DatasetFactory"],
     load_dataset_metadata: bool,
     load_schema: bool = True,
+    gen_uuid_if_none: bool = False,
 ) -> "DatasetFactory":
+
     if store is None and dataset_uuid is None and factory is not None:
         return factory
-    elif store is not None and dataset_uuid is not None and factory is None:
+    elif store is not None and factory is None:
+        if dataset_uuid is None:
+            if gen_uuid_if_none:
+                dataset_uuid = gen_uuid()
+            else:
+                raise ValueError("Missing `dataset_uuid`.")
         return DatasetFactory(
             dataset_uuid=dataset_uuid,
             store_factory=store,
--- a/kartothek/io/dask/delayed.py
+++ b/kartothek/io/dask/delayed.py
@@ -10,9 +10,11 @@ from dask import delayed
 from kartothek.core import naming
 from kartothek.core.docs import default_docs
 from kartothek.core.factory import _ensure_factory
-from kartothek.core.naming import DEFAULT_METADATA_VERSION
+from kartothek.core.naming import (
+    DEFAULT_METADATA_STORAGE_FORMAT,
+    DEFAULT_METADATA_VERSION,
+)
 from kartothek.core.utils import _check_callable
-from kartothek.core.uuid import gen_uuid
 from kartothek.io_components.delete import (
     delete_common_metadata,
     delete_indices,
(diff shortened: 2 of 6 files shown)


lr4d:
This partially resolves https://github.com/JDASoftwareGroup/kartothek/issues/269

lr4d:
Added a commit to also make `store__delayed` use `update__delayed`

lr4d:
Looks like the CI doesn't like force-pushes and is still using my commit `c4027d243ed80c304fb2ab1c610f765065d6a278` instead of `10ebdff`

kartothek/core/factory.py
```diff
@@ -16,10 +17,17 @@ def _ensure_factory(
     factory: Optional["DatasetFactory"],
     load_dataset_metadata: bool,
     load_schema: bool = True,
+    gen_uuid_if_none: bool = False,
```
fjetter, line 20: The `DatasetFactory` still has certain design flaws and it was not intended to be used for writing paths. Primarily because it only makes sense if the dataset actually exists since it's effectively a caching layer. That's obviously not the case if we're in a pure write operation. I'm afraid that when using it for storage operations as well this leads to some weird bugs.

kartothek/io_components/write.py
```diff
@@ -246,3 +253,28 @@ def raise_if_dataset_exists(dataset_uuid, store):
...
+
+
+def check_dataset_creation_constraints(
+    factory, partition_on, secondary_indices, overwrite=False
+):
+    """Check if all required constraints for creating a dataset are met."""
+
+    if overwrite is not False:
+        warnings.warn(
+            "The keyword `overwrite` be removed in the next major release ",  # FIXME: keep or not?
```
fjetter, line 265: I think we should keep it. "Store" vs. "Update" have different semantics and for "store" I see very reasonable use cases in overwriting datasets

kartothek/io_components/write.py
```diff
@@ -246,3 +253,28 @@ def raise_if_dataset_exists(dataset_uuid, store):
...
+            stacklevel=2,
+        )
+        if DatasetMetadata.exists(
+            uuid=factory.dataset_uuid, store=factory.store_factory()
+        ):
+            # delete common metadata after partitions
+            delete_common_metadata(dataset_factory=factory)
+            delete_indices(dataset_factory=factory)
+            # Delete the top level metadata file
+            delete_top_level_metadata(dataset_factory=factory)
```
fjetter, line 276: This changes functionality quite significantly. I'm quite strongly against deleting anything explicitly in a write operation. If the write operation doesn't success for whatever reason (e.g. final commit doesn't succeed due to schema violations) the previous implementation kept the dataset intact.

At the moment there is the potential for an inconsistent dataset state in case the dataset common metadata has been written but the rest of the commit failed. I'm willing to accept this fringe edge case in favour of protection from data loss.

kartothek/io_components/update.py
```diff
@@ -28,6 +29,7 @@ def update_dataset_from_partitions(
     delete_scope,
     metadata,
     metadata_merger,
+    metadata_storage_format=DEFAULT_METADATA_STORAGE_FORMAT,
```
fjetter, line 32: This is an internal function, please remove the default value. Internally we should always call it with all values

kartothek/io/iter.py
```diff
@@ -196,6 +192,7 @@ def update_dataset_from_dataframes__iter(
     sort_partitions_by=None,
     secondary_indices=None,
     factory=None,
+    metadata_storage_format=DEFAULT_METADATA_STORAGE_FORMAT,
```
fjetter, line 195: We've been frequently discussing dropping this kwarg once and for all. I would suggest to not introduce it now for functions where it's been forgotten so far.

kartothek/io/dask/delayed.py
```diff
@@ -439,6 +433,7 @@ def update_dataset_from_delayed(
     sort_partitions_by=None,
     secondary_indices=None,
     factory=None,
+    metadata_storage_format=DEFAULT_METADATA_STORAGE_FORMAT,
```
fjetter, line 436: We've been frequently discussing dropping this kwarg once and for all. I would suggest to not introduce it now for functions where it's been forgotten so far.

kartothek/io_components/write.py
lr4d: - [ ] keep it

kartothek/io_components/write.py
lr4d: > If the write operation doesn't success for whatever reason (e.g. final commit doesn't succeed due to schema violations) the previous implementation kept the dataset intact.

Hmm, right. 

- [ ] look into this

kartothek/io_components/update.py
lr4d: - [] remove

kartothek/io/iter.py
lr4d: It's required to not break the `store_` signature

kartothek/core/factory.py
lr4d: Does it make sense to refactor this as part of a different issue?

lr4d:
Closing for now as I won't be working on this for at laest a while

PR #319, opened 2020-07-22
Add hash_dataset and make dataframe payload compression functions public

fjetter:
# Description:

This should allow for partition or group based hashing of the data and should make things easier if we need to verify data integrity of some sorts.

I also moved the pack/unpack logic to compress the payload into a public package. I think this can be quite useful outside of "standard" kartothek workflows as long as it's not part of dask, yet

### DIFF
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -12,6 +12,7 @@ from kartothek.core.common_metadata import empty_dataframe_from_schema
 from kartothek.core.docs import default_docs
 from kartothek.core.factory import DatasetFactory, _ensure_factory
 from kartothek.core.naming import DEFAULT_METADATA_VERSION
+from kartothek.io.dask.compression import pack_payload, unpack_payload_pandas
 from kartothek.io_components.metapartition import (
     _METADATA_SCHEMA,
     SINGLE_TABLE,
@@ -394,7 +395,7 @@ def collect_dataset_metadata(
         ddf = dd.from_delayed(
             [
                 dask.delayed(MetaPartition.get_parquet_metadata)(
-                    mp, store=dataset_factory.store_factory, table_name=table_name,
+                    mp, store=dataset_factory.store_factory, table_name=table_name
                 )
                 for mp in mps
             ]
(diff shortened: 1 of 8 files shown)


kartothek/io/dask/dataframe.py
```diff
@@ -405,3 +406,61 @@ def collect_dataset_metadata(
...
+    table: str = SINGLE_TABLE,
+    predicates: Optional[PredicatesType] = None,
+    factory: Optional[DatasetFactory] = None,
+):
+    """
+    Calculate a partition wise, or group wise, hash of the dataset.
+
+    .. note::
+
+        We do not guarantee the hash values to remain constant accross versions.
```
lr4d, line 435: What about same values but different data types (e.g. uint64(1) == int64(1))?

kartothek/io/dask/dataframe.py
```diff
@@ -405,3 +406,61 @@ def collect_dataset_metadata(
...
     return ddf
+
+
+def _unpack_hash(df, unpack_meta):
+    df2 = unpack_payload_pandas(df, unpack_meta)
+    return _hash_partition(df2)
+
+
+def _hash_partition(part):
+    return pd.util.hash_pandas_object(part, index=False).sum().astype("uint64")
```
lr4d, line 417: I suppose the purpose of `uint64` is to prevent overflows here
```suggestion
    return pd.util.hash_pandas_object(part, index=False).astype("uint64").sum()
```

fjetter: No this is indeed expected. Summing the uint will cause overflows but these are handled gracefully. The result is then an int64 which may be negative. to stick with the practice of having uints (as is the case for the original pandas hash) I cast them back (which works also for negative numbers)


```python
In [7]: pd.Series([-1]).astype("uint64")
Out[7]:
0    18446744073709551615
dtype: uint64
```

kartothek/io/dask/dataframe.py
fjetter: can you rephrase the question, please? Are you referring to a guarantee about type stability or what are you referring to here? This remark shall simply indicate that we may (or may not) change the implementation without notice and we do not necessarily recommend to use this implementation for long term storage or similar.

For instance, we may want to change how hashes are combined, e.g. instead of doing `hash(partition).sum()` we may want to go for another hashing of the values, e.g. `hash(*hash(partition))` and I don't want anybody to rely on this implementation detail

lr4d: Sorry for the lack of clarity. I meant to say that we might note here whether 2 datasets with the same values but different dtypes will have the same hash.

kartothek/io/dask/dataframe.py
lr4d: Makes sense. Just wondering (not suggesting), could we _prevent_ overflows here?

fjetter: only if we'd use a hash function with a different domain. the pandas hash maps to the entire uint64 space, i.e. as soon as we apply a sum we're in overflow territory. Everything else would need it's dedicated implementation mapping to a smaller domain

kartothek/io/dask/dataframe.py
fjetter: That's also quite another topic entirely. For the sake of the argument, though

TL;DR

Same value == same hash

different type family -> different value -> different hash

```python
In [1]: import pandas as pd

In [2]: pd.util.hash_pandas_object(pd.DataFrame({"int": [1]})) == pd.util.hash_pandas_object(pd.DataFrame({"int": [1]}).astype({"int": "int8"}))
Out[2]:
0    True
dtype: bool

In [3]: pd.util.hash_pandas_object(pd.DataFrame({"int": [1]})) == pd.util.hash_pandas_object(pd.DataFrame({"int": [1]}).astype({"int": "int32"}))
Out[3]:
0    True
dtype: bool

In [4]: pd.util.hash_pandas_object(pd.DataFrame({"int": [1]})) == pd.util.hash_pandas_object(pd.DataFrame({"int": [1.0]}))
Out[4]:
0    False
dtype: bool
```

kartothek/io/dask/dataframe.py
lr4d: Even if we'd put it into `np.complex128` someway (for lack of `uint128`) we could still overflow when `len(part) >= 2^64 ~= 1.8^19`

kartothek/io/dask/dataframe.py
lr4d: Thanks for the explanation.
```suggestion
        We do not guarantee the hash values to remain constant accross versions.
        It should not be expected that the hash of the same value in different data types is identical.
```

kartothek/io/dask/dataframe.py
```diff
@@ -405,3 +406,65 @@ def collect_dataset_metadata(
...
+@default_docs
+def hash_dataset(
+    store: Optional[Callable[[], KeyValueStore]] = None,
+    dataset_uuid: Optional[str] = None,
+    subset=None,
+    group_key=None,
+    table: str = SINGLE_TABLE,
+    predicates: Optional[PredicatesType] = None,
+    factory: Optional[DatasetFactory] = None,
+):
```
lr4d, line 431: Please add the return type

lr4d: And since it's a ddf, would you mind describing it in the docstring also?

PR #324, opened 2020-07-28
Fix predicate parsing for "OR" predicates

### LINKED ISSUE #295: predicate parsing with OR broken
### Problem description
See https://github.com/JDASoftwareGroup/kartothek/pull/294#issuecomment-638922148

However, after fixing the case for empty predicates on that development branch, the following bug shows up:
Reading datasets using somewhat complex predicates sometimes returns incorrect data.

```python

from storefact import get_store_from_url
from functools import partial

store_factory = partial(get_store_from_url, f"hfs://{TemporaryDirectory().name}")
dataset_uuid = "test"

import pandas as pd

df = pd.DataFrame({"A": range(10), "B": ["A", "B"] * 5, "C": range(-10, 0)})

from kartothek.io.eager import read_table, store_dataframes_as_dataset

    store=store_factory, dataset_uuid=dataset_uuid, dfs=[df], partition_on=["A", "B"]
)

df1 = read_table(
    store=store_factory,
    dataset_uuid=dataset_uuid,
    predicates=[[("A", "<", 3)], [("A", ">", 5)], [("B", "==", "non-existent")]],
)
print(df1)

df2 = read_table(
    store=store_factory,
    dataset_uuid=dataset_uuid,
    predicates=[[("A", "<", 3)], [("A", ">", 5)]],
)

pd.testing.assert_frame_equal(df1, df2)
```
Output:
```
   A  B   C
0  0  A -10
1  1  B  -9
2  2  A  -8
3  3  B  -7
4  4  A  -6
5  5  B  -5
6  6  A  -4
7  7  B  -3
8  8  A  -2
9  9  B  -1
Traceback (most recent call last):
  File "/Users/xrzq/miniconda3/envs/sc-lastcalcs/lib/python3.6/site-packages/IPython/core/interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-d6e03a9f4370>", line 31, in <module>
    pd.testing.assert_frame_equal(df1, df2)
  File "/Users/xrzq/miniconda3/envs/sc-lastcalcs/lib/python3.6/site-packages/pandas/_testing.py", line 1325, in assert_frame_equal
    obj, f"{obj} shape mismatch", f"{repr(left.shape)}", f"{repr(right.shape)}",
  File "/Users/xrzq/miniconda3/envs/sc-lastcalcs/lib/python3.6/site-packages/pandas/_testing.py", line 915, in raise_assert_detail
    raise AssertionError(msg)
AssertionError: DataFrame are different
DataFrame shape mismatch
[left]:  (10, 3)
[right]: (7, 3)
```
Note that `df1` contains values for `A` outside of the range defined by the predicates.

nefta-kanilmaz-jdas:
Predciate evaluation of "OR" connected predicates was not implemented correctly. This could lead to too many files returned by `dispatch_metapartitions`.

Closes #295

### DIFF
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -361,7 +361,7 @@ class DatasetMetadataBase(CopyMixin):
         predicates: PredicatesType = None,
     ):
         """
-        Converts the dataset indices to a pandas dataframe.
+        Converts the dataset indices to a pandas dataframe and filter relevant indices by `predicates`.
 
         For a dataset with indices on columns `column_a` and `column_b` and three partitions,
         the dataset output may look like
@@ -381,22 +381,76 @@ class DatasetMetadataBase(CopyMixin):
         elif columns == []:
             return pd.DataFrame(index=self.partitions)
 
-        dfs = []
         columns_to_scan = columns[:]
         if predicates:
             predicate_columns = columns_in_predicates(predicates)
-            # Don't use set logic to preserve order
-            for col in predicate_columns:
-                if col not in columns_to_scan and col in self.indices:
-                    columns_to_scan.append(col)
+            columns_to_scan = sorted(
+                (predicate_columns & self.indices.keys()) | set(columns)
+            )
 
-        for col in columns_to_scan:
-            if col not in self.indices:
-                if col in self.partition_keys:
-                    raise RuntimeError(
-                        "Partition indices not loaded. Please call `DatasetMetadata.load_partition_keys` first."
-                    )
-                raise ValueError("Index `{}` unknown.")
+            dfs = (
+                self._evaluate_conjunction(
+                    columns=columns_to_scan,
+                    predicates=[conjunction],
+                    date_as_object=date_as_object,
+                )
+                for conjunction in predicates
+            )
+
+            df = pd.concat(dfs)
+            index_name = df.index.name
+            df = (
+                df.loc[:, columns].reset_index().drop_duplicates().set_index(index_name)
+            )
+        else:
+            df = self._evaluate_conjunction(
+                columns=columns_to_scan, predicates=None, date_as_object=date_as_object,
+            )
+        return df
+
+    def _evaluate_conjunction(self, columns, predicates, date_as_object):
+        """
+        Evaluate all predicates related to `columns` to "AND".
+
+        Parameters
+        ----------
+        columns:
+            A list of all columns, including query and index columns.
+        predicates: list of list of tuple[str, str, Any]
+            Optional list of predicates, like [[('x', '>', 0), ...], that are used
+            to filter the resulting DataFrame, possibly using predicate pushdown,
+            if supported by the file format.
+            This parameter is not compatible with filter_query.
+
+            Predicates are expressed in disjunctive normal form (DNF). This means
+            that the innermost tuple describes a single column predicate. These
+            inner predicates are all combined with a conjunction (AND) into a
+            larger predicate. The most outer list then combines all predicates
+            with a disjunction (OR). By this, we should be able to express all
+            kinds of predicates that are possible using boolean logic.
+
+            Available operators are: `==`, `!=`, `<=`, `>=`, `<`, `>` and `in`.
+        dates_as_object: bool
+            Load pyarrow.date{32,64} columns as ``object`` columns in Pandas
+            instead of using ``np.datetime64`` to preserve their type. While
+            this improves type-safety, this comes at a performance cost.
+
+        Returns
+        -------
+        df_result:
+            A DataFrame containing all indices for which `predicates` holds true.
+        """
+        non_index_columns = set(columns) - self.indices.keys()
+        if non_index_columns:
+            if non_index_columns & set(self.partition_keys):
+                raise RuntimeError(
+                    "Partition indices not loaded. Please call `DatasetMetadata.load_partition_indices` first."
+                )
+            raise ValueError(
+                "Unknown index columns: {}".format(", ".join(sorted(non_index_columns)))
+            )
+        dfs = []
+        for col in columns:
             df = pd.DataFrame(
                 self.indices[col].as_flat_series(
                     partitions_as_index=True,
@@ -405,24 +459,16 @@ class DatasetMetadataBase(CopyMixin):
                 )
             )
             dfs.append(df)
-
-        # start joining with the small ones
-        sorted_dfs = sorted(dfs, key=lambda df: len(df))
-        result = sorted_dfs.pop(0)
-        for df in sorted_dfs:
-            result = result.merge(df, left_index=True, right_index=True, copy=False)
-
-        if predicates:
-            index_name = result.index.name
-            result = (
-                result.loc[:, columns]
-                .reset_index()
-                .drop_duplicates()
-                .set_index(index_name)
+        # dfs contains one df per index column. Each df stores indices filtered by `predicates` for each column.
+        # Performing an inner join on these dfs yields the resulting "AND" evaluation for all of these predicates.
+        # We start joining with the smallest dataframe, therefore the sorting.
+        dfs_sorted = sorted(dfs, key=len)
+        df_result = dfs_sorted.pop(0)
+        for df in dfs_sorted:
+            df_result = df_result.merge(
+                df, left_index=True, right_index=True, copy=False
             )
-            return result
-        else:
-            return result
+        return df_result
 
 
 class DatasetMetadata(DatasetMetadataBase):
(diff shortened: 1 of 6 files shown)


lr4d:
Can you add tests for `[[("A", "<", 3)], [("B", "==", "B"), ("A", ">", 2)]]`, `[[("A", "<", 3), ("B", "==", "A")], [("B", "==", "B"), ("A", ">", 2)]]`?

kartothek/core/dataset.py
```diff
@@ -381,16 +381,69 @@ def get_indices_as_dataframe(
         elif columns == []:
             return pd.DataFrame(index=self.partitions)
 
-        dfs = []
         columns_to_scan = columns[:]
         if predicates:
             predicate_columns = columns_in_predicates(predicates)
             # Don't use set logic to preserve order
+            # TODO why do I need to preserve the order here?
```
marco-neumann-by, line 388: That's a good question and doesn't make sense to me at all. If this is about deterministic behavior, a well-places `sorted` would also do the trick.

kartothek/core/dataset.py
```diff
@@ -381,16 +381,69 @@ def get_indices_as_dataframe(
...
+            dfs = [
+                self._evaluate_conjunction(
+                    columns=columns_to_scan,
+                    predicates=[conjunction],
+                    date_as_object=date_as_object,
+                )
+                for conjunction in predicates
+            ]
+
+            df = pd.concat(dfs)
```
marco-neumann-by, line 402: does this work without `ignore_index=True`?

kartothek/core/dataset.py
fjetter: I would assume this originated from a very early version. The only reason for not mixing the ordering I see here is determinism and this is only relevant for the output shape, i.e. we must preserve the original `columns` and reshape the output accordingly.

```python
columns_to_scan = sorted(set(columns) | set(columns_in_predicates))

# calc stuff

return df.loc[:, columns] # result with original columns

```

marco-neumann-by:
I'm very much in favor of adding the regression testcase from #295 as well. That said however, dispatching too many partitions is at least a performance bug.

kartothek/core/dataset.py
NeroCorleone: Replaced it.

kartothek/core/dataset.py
NeroCorleone: Wouldn't `ignore_index=True` drop the index entirely? We would lose the mapping to the partitions, wouldn't we?

tests/io/eager/test_read.py
```diff
@@ -159,3 +159,30 @@ def test_write_and_read_default_table_name(store_session):
...
+        predicates=[[("A", "<", 3)], [("A", ">", 5)], [("B", "==", "non-existent")]],
+    )
+
+    df2 = read_table(
+        store=store_session,
+        dataset_uuid=dataset_uuid,
+        predicates=[[("A", "<", 3)], [("A", ">", 5)]],
+    )
+
+    pd.testing.assert_frame_equal(df1, df2)
```
lr4d, line 188:
```suggestion
    pd.testing.assert_frame_equal(df1, df2)
    pd.testing.assert_frame_equal(df[(df["A"] < 3) | (df["A"] > 5)], df2)
```

tests/io_components/test_read.py
```diff
@@ -217,3 +217,21 @@ def test_dispatch_metapartitions_or_predicate(store_factory):
...
+        dispatch_metapartitions(dataset_uuid, store_factory, predicates=predicates)
+    )
+    assert len(mps) == 3
+
+    predicates = [[("A", "<", 3), ("B", "==", "A")], [("B", "==", "B"), ("A", ">", 2)]]
+    mps = list(
+        dispatch_metapartitions(dataset_uuid, store_factory, predicates=predicates)
+    )
+    assert len(mps) == 6
+
```
lr4d, line 232:
```suggestion
    actual = pd.concat([mp.data for mp in mps], ignore_index=True)
    expected = df[((df["A"] < 3) & (df["B"] == "A")) | ((df["B"] == "B") & (df["A"] > 2))]
    assert pd.testing.assert_frame_equal(actual, expected)

```

tests/io/eager/test_read.py
```diff
@@ -159,3 +159,30 @@ def test_write_and_read_default_table_name(store_session):
         "test_default_table_name", store_session
     )
     pd.testing.assert_frame_equal(df_write, df_read_as_dfs[0]["table"])
+
+
+def test_read_or_predicates(store_session):
+    # https://github.com/JDASoftwareGroup/kartothek/issues/295
```
NeroCorleone, line 165: This test includes the example from the original description, where the first `read_table` would raise a `Value Error: empty Predicates`. I might have fixed this behaviour accidentally, by focusing on the correct "OR" evaluation during dispatch.

I have included a test for `filter_df_from_predicates`, which works as intended (cc @fjetter), currently looking into the filtering here: https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/io_components/metapartition.py#L604
🔝 This can return an empty list as a result for `filtered_predicate`, which leads to the raise of the Value Error later.

kartothek/core/dataset.py
```diff
@@ -384,11 +384,9 @@ def get_indices_as_dataframe(
         columns_to_scan = columns[:]
         if predicates:
             predicate_columns = columns_in_predicates(predicates)
-            # Don't use set logic to preserve order
-            # TODO why do I need to preserve the order here?
-            for col in predicate_columns:
-                if col not in columns_to_scan and col in self.indices:
-                    columns_to_scan.append(col)
+            columns_to_scan = sorted(
+                predicate_columns.intersection(self.indices.keys()).union(columns)
```
lr4d, line 388: nitpick, I think this improves legibility, feel free to ignore
```suggestion
                (predicate_columns & self.indices.keys()) | set(columns)
```

kartothek/core/dataset.py
```diff
@@ -381,20 +381,71 @@ def get_indices_as_dataframe(
...
+            df = (
+                df.loc[:, columns].reset_index().drop_duplicates().set_index(index_name)
+            )
+        else:
+            df = self._evaluate_conjunction(
+                columns=columns_to_scan, predicates=None, date_as_object=date_as_object,
+            )
+        return df
+
+    def _evaluate_conjunction(self, columns, predicates, date_as_object):
```
lr4d, line 411: 👍

kartothek/core/dataset.py
```diff
@@ -381,20 +381,71 @@ def get_indices_as_dataframe(
...
+            instead of using ``np.datetime64`` to preserve their type. While
+            this improves type-safety, this comes at a performance cost.
+
+        Returns
+        -------
+        df_result:
+            A DataFrame containing all indices for which `predicates` holds true.
+        """
+        dfs = []
+        for col in columns:
```
lr4d, line 444: You could also use set logic here btw, (columns - indices) & partition_keys

kartothek/core/dataset.py
```diff
@@ -405,24 +456,16 @@ def get_indices_as_dataframe(
...
-            index_name = result.index.name
-            result = (
-                result.loc[:, columns]
-                .reset_index()
-                .drop_duplicates()
-                .set_index(index_name)
+        # dfs contains one df per index column. Each df stores indices filtered by `predicates` for each column.
+        # Performing an inner join on these dfs yields the resulting "AND" evaluation for all of these predicates.
+        # We start joining with the smallest dataframe, therefore the sorting.
+        dfs_sorted = sorted(dfs, key=lambda df: len(df))
```
lr4d, line 462: nit
```suggestion
        dfs_sorted = sorted(dfs, key=len)
```

kartothek/core/dataset.py
```diff
@@ -381,20 +381,71 @@ def get_indices_as_dataframe(
...
             predicate_columns = columns_in_predicates(predicates)
-            # Don't use set logic to preserve order
-            for col in predicate_columns:
-                if col not in columns_to_scan and col in self.indices:
-                    columns_to_scan.append(col)
+            columns_to_scan = sorted(
+                (predicate_columns & self.indices.keys()) | set(columns)
+            )
+
+            dfs = [
```
lr4d, line 391: extreme nit: this could be a generator? Not sure if `pd.concat` would handle that efficiently

marco-neumann-by:
@nefta-kanilmaz-jdas can you please rebase on latest master to silence the coverage bot?

Issue #317, opened 2020-07-22
Deprecate read support for multi tabled datasets


fjetter:
In our efforts to deprecated the in-dataset multi table functionality, the first step will be to deprecate and remove the possibility to read from a multi tabled dataset. The first step should be to detect multi tabled datasets during reading and issue a deprecation warning.

The second phase will then be to only allow reading from a single valued dataset, rendering the `table` kwarg obsolete, defaulting to whatever table exists in the dataset. In phase two we may also remove support in the write operation, s.t. it always defaults to SINGLE_TABLE unless in update mode.

Third phase (hard breaking) would be to drop the support entirely and replace the current metadata specification with a new version.

lr4d:
Do you envision a time line for each phase?

lr4d:
First phase checklist:
- [ ] issue warning when reading multiple tables (done in #337)
- [ ] at least the dict type kwargs for categoricals and columns also need to be adapted (potentially more, check this)
- [ ] there are also the functions which only work on a single table (e.g. everything in dask.dataframe and something like eager.read_table) which should issue a warning if the requested table is not the SINGLE_TABLE constant

imkumarg:
Okay in PR #349, I will address the first, third and fourth point.

PR #337, opened 2020-08-12
deprecate multi-table read support

lr4d:
# Description:

deprecate multi-table read support

- [x] Closes https://github.com/JDASoftwareGroup/kartothek/issues/317
- [x] Changelog entry

### DIFF
--- a/kartothek/core/_deprecation.py
+++ b/kartothek/core/_deprecation.py
@@ -1,5 +1,5 @@
 """
-Code was taken from pandas
+`deprecate_kwarg` code was taken from pandas
 
 https://github.com/pandas-dev/pandas/blob/c5a47118d73a53a016a28103726b57e27924bf29/pandas/util/_decorators.py#L66
 
@@ -149,3 +149,20 @@ def deprecate_kwarg(old_arg_name, new_arg_name, mapping=None, stacklevel=2):
         return wrapper
 
     return _deprecate_kwarg
+
+
+def deprecate_multi_table_read(func, *, stacklevel=2):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        tables = kwargs.get("tables")
+        if tables is None:  # sometimes `tables=None` is passed
+            tables = []
+        if len(tables) > 1:
+            warnings.warn(
+                "multi-table support is deprecated and will be removed in the next major release.",
+                DeprecationWarning,
+                stacklevel=stacklevel,
+            )
+        return func(*args, **kwargs)
+
+    return wrapper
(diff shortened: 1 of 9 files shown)


kartothek/core/_deprecation.py
```diff
@@ -149,3 +149,20 @@ def wrapper(*args, **kwargs):
         return wrapper
 
     return _deprecate_kwarg
+
+
+def deprecate_multi_table_read(func, *, stacklevel=2):
```
marco-neumann-by, line 154: While I can see the technical elegance of our function wrappers, it makes our code really badly readable / understandable.

lr4d: Generally, I would agree with you. But we are not changing any actual functionality here, so rather than calling a function which emits the warning or copy-pasting code inside of each `read_dataframes_*` function, I think this is more appropriate as we are keeping the actual function code limited to the behavior of the function instead of clobbering it with deprecation code.

fjetter:
I don't think this is all we need to do to issue appropriate and useful deprecation warnings. there are at least the dict type kwargs for categoricals and columns which also need to be adapted and potentially more.

Apart from this, there are also the functions which only work on a single table (e.g. everything in `dask.dataframe` and something like `eager.read_table`) which should issue a warning if the requested table is not the SINGLE_TABLE constant.

-----

More importantly: What does this mean for the user? How is the user expected to change their code to prepare for this change? From what I envision is that we'll have a reduced set of API functions and we should direct users to the recommended new way instead of only telling them that something will be gone (e.g. eager.read_table could be part of the next-gen API since it works only. on single tables in the first place)

Therefore, I believe we need to have a discussion about the future interface before we spam deprecation warnings. To keep the scope limited I wouldn't try to merge the single-dataset and cube interfaces in this step but rather figure out how a single dataset API should look like.

e.g.

`read_dataset_as_dataframes__delayed` could be replaced by a new, single table, slightly easier to pronounce function we can point users to in the respective deprecation warning. this way the feature drop would be a clean removal of functions instead of change of behaviour.
If we decide not to use a new function we should at least outline a migration path (could be part of the docs if it is something larger) and tell the users how they can configure the function calls such that the breakage will not happen for them. The "we'll drop this without any hints" approach really only works for minor, rarely used features

lr4d:
>at least the dict type kwargs for categoricals and columns which also need to be adapted and potentially more.

Right, we'll want to issue a warning that this format will also change, and support a format more appropriate for single-table datasets (i.e. tuple/list).

> Apart from this, there are also the functions which only work on a single table (e.g. everything in dask.dataframe and something like eager.read_table) which should issue a warning if the requested table is not the SINGLE_TABLE constant.

This also touches the UX you mention in the latter part of your post, we could create a migration function for single-table datasets to single-table datasets with `table_name=SINGLE_TABLE`.

I guess we could create some action points in #317 to outline what needs to be done for this issue and continue the discussion there.

Issue #227, opened 2020-02-17
Type error when loading categorical data in dask without specifying it explicitly


fjetter:
### Problem description

Starting with pyarrow 0.15.X we are able to store pandas categorical data directly to the parquet file. This categorical dtype is upon reading then also properly reconstructed as a categorical type w/out the explicit flag in Kartothek.
Loading the data using the `read_dataset_as_ddf` interface will then set the _incorrect_ categorical type information for the dask meta object yielding a type error when doing anything in dask due to the dask internal partition schema verification logic. 

### Example code (ideally copy-pastable)

Please provide a minimal reproducible code example to reproduce the behavior,
c.f. https://stackoverflow.com/help/minimal-reproducible-example

```python
from functools import partial

import pandas as pd
from kartothek.io.dask.dataframe import read_dataset_as_ddf
from kartothek.io.eager import store_dataframes_as_dataset
from storefact import get_store_from_url

df = pd.DataFrame({"string_col": ["A"]}).astype({"string_col": "category"})

store_fact = partial(get_store_from_url, "hfs://test")
store_dataframes_as_dataset(
    dataset_uuid="test_uuid",
    store=store_factory,
    dfs=[df]
)

ddf = read_dataset_as_ddf(
    dataset_uuid="test_uuid",
    store=store_fact,
    table="table"
)
ddf.compute()

```

raises

```
ValueError: Metadata mismatch found in `from_delayed`.

Partition type: `pandas.core.frame.DataFrame`
+------------+----------+----------+
| Column     | Found    | Expected |
+------------+----------+----------+
| string_col | category | object   |
+------------+----------+----------+
```

### Used versions

<details>

```
# Paste your output of `pip freeze` or `conda list` here
dask==2.10.1
pyarrow==0.15.1
```

</details>

PR #340, opened 2020-08-25
Drop ancient arrow versions

fjetter:
This preservers read support for effectively everything back to 0.12.X. I suggest to drop read support in the next major release.
After this change, we'll support arrow starting with `0.17.1`. How the support window looks like in the future depends on how much compat code we'll need to maintain and I'd keep this decision until later.

I suggest to cut a last release with the support of the old versions before merging this.


fjetter:
cc @xhochy @fhoehle

kartothek/serialization/_parquet.py
```diff
@@ -40,65 +37,9 @@
...
-            col = table[i]
-            if col.name in exclude:
-                continue
-            if pa.types.is_dictionary(col.type):
-                new_type = col.data.chunk(0).dictionary.type
-                new_col = col.cast(new_type)
-                table = table.remove_column(i).add_column(i, new_col)
-    return table
-
-
```
lr4d, line 101: looks like we're hitting https://github.com/JDASoftwareGroup/kartothek/issues/227 again because of this change, see https://github.com/JDASoftwareGroup/kartothek/runs/1025742983

lr4d (approved):

kartothek/serialization/_parquet.py
fjetter: Right, I forgot that we sometimes actually want to reset these... what a waste :D

kartothek/serialization/_parquet.py
```diff
@@ -40,62 +37,28 @@
...
-                        field.name,
-                        field.type.value_type,
-                        field.nullable,
-                        field.metadata,
-                    )
-                    schema = schema.remove(i).insert(i, new_field)
+    schema = table.schema
+    for i in range(len(schema)):
+        field = schema[i]
+        if field.name in exclude:
```
fjetter, line 53: In the previous code, it seems we ignored the exclude for arrow>=0.15.0.

fjetter: behaviour was the same but we worked more than necessary by converting it twice in this case. Added a test case since this has potentially huge performance implications for readers

kartothek/core/common_metadata.py
```diff
@@ -78,10 +76,7 @@ def _schema_compat(self):
 
             schema = schema.remove_metadata()
             md = {b"pandas": _dict_to_binary(pandas_metadata)}
-            if ARROW_LARGER_EQ_0150:
-                schema = schema.with_metadata(md)
-            else:
-                schema = schema.add_metadata(md)
+            schema = schema.add_metadata(md)
```
lr4d, line 79:
```suggestion
            schema = schema.with_metadata(md)
```
causes build to fail

tests/serialization/test_parquet.py
```diff
@@ -35,46 +37,6 @@ def test_timestamp_us(store):
...
-        ]
-    }"""
-    table = table.replace_schema_metadata({b"pandas": meta})
-    buf = pa.BufferOutputStream()
-    pq.write_table(table, buf)
-    store.put(key, buf.getvalue().to_pybytes())
-    pdt.assert_frame_equal(DataFrameSerializer.restore_dataframe(store, key), df)
-
-
-def test_index_metadata(store):
```
fjetter, line 60: This test was also related to the arrow 0.7.992 version (this version was only used BY internally)

lr4d (approved):

tests/serialization/test_parquet.py
marco-neumann-by: What's wrong with keeping it?

marco-neumann-by (approved):

PR #344, opened 2020-09-07
reduce memory consumption during index write

marco-neumann-by:
In a particular test case and a maxRSS background (aka usage before index creation) of
570MB this reduces maxRSS from 1.80GB to 970MB.

### DIFF
--- a/kartothek/core/index.py
+++ b/kartothek/core/index.py
@@ -9,6 +9,7 @@ import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
 from simplekv import KeyValueStore
+from toolz.itertoolz import partition_all
 
 import kartothek.core._time
 from kartothek.core import naming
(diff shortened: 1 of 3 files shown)


kartothek/core/index.py
```diff
@@ -688,9 +689,27 @@ def store(self, store: KeyValueStore, dataset_uuid: str) -> str:
                 timestamp=quote(self.creation_time.isoformat()),
             )
 
-        table = _index_dct_to_table(self.index_dct, self.column, self.dtype)
+        # The arrow representation of index_dct requires a large amount of memory because strings are duplidated and
```
lr4d, line 692:
```suggestion
        # The arrow representation of index_dct requires a large amount of memory because strings are duplicated and
```

kartothek/core/index.py
```diff
@@ -688,9 +689,27 @@ def store(self, store: KeyValueStore, dataset_uuid: str) -> str:
...
+        # flattened into the buffer. To avoid a high peak memory usage, split the index_dct into chunks and only convert
+        # one chunk a time to arrow.
+        parts_iter = partition_all(10_000, self.index_dct.items())
+
+        # Get first table explicit because its schema is required for ParquetWriter.
+        try:
+            table = _index_dct_to_table(dict(next(parts_iter)), self.column, self.dtype)
+        except StopIteration:
+            # index_dct was empty, just pass it entirely
```
lr4d, line 701: why?

lr4d (commented):
first pass, looks good. Have you measured the impact on runtime performance?

kartothek/core/index.py
fjetter: index dict empty -> index_dct.items() creates an empty iterator -> next causes StopIteration exception

We still need to call the function to get the arrow table. we do store a parquet file after all, and I wouldn't want to change that.

fjetter (approved):

fjetter (approved):

PR #345, opened 2020-09-09
Allow URLs and stores to be passed explicitly

### LINKED ISSUE #233: Allow store URLs for the `store` parameter
Currently we require callables that return simplekv stores to be passed to the `store=` parameter. When working with `kartothek` interactively this is quite an annoyance as you always just so `lambda: storefact.get_store_from_url(url)` in these settings. We could save some characters and make the enduser code more readable by also supporting URLs directly.

fjetter:
# Description:

This allows to pass either store urls, simplekv stores or callables as the store argument. the type annotations helped really out in this case. Eventually I imagine low level classes should phase out any ambiguity and simply deal with KV stores.

This obviously assumes that KV stores can be properly pickled but I don't see this as a relevant issue anymore. If this is an issue for anybody I can revert that change and only allow store URLs.

There may be a few leftovers in docs, etc. but I tried to get the essentials (e.g. in the guide/getting started)

- [x] Closes #233
- [x] Changelog entry

### DIFF
--- a/kartothek/core/utils.py
+++ b/kartothek/core/utils.py
@@ -1,9 +1,11 @@
+from functools import partial
+from typing import cast
+
+from simplekv import KeyValueStore
+from storefact import get_store_from_url
+
 from kartothek.core.naming import MAX_METADATA_VERSION, MIN_METADATA_VERSION
-
-
-def _check_callable(store_factory, obj_type="store"):
-    if not callable(store_factory):
-        raise TypeError("{} must be a factory function".format(obj_type))
+from kartothek.core.typing import STORE_FACTORY_TYPE, STORE_TYPE
 
 
 def _verify_metadata_version(metadata_version):
@@ -46,3 +48,22 @@ def ensure_string_type(obj):
         return obj.decode()
     else:
         return str(obj)
+
+
+def ensure_store(store: STORE_TYPE) -> KeyValueStore:
+    return lazy_store(store)()
+
+
+def lazy_store(store: STORE_TYPE) -> STORE_FACTORY_TYPE:
+    if callable(store):
+        return cast(STORE_FACTORY_TYPE, store)
+    elif isinstance(store, KeyValueStore):
+        return lambda: store
+    elif isinstance(store, str):
+        ret_val = partial(get_store_from_url, store)
+        ret_val = cast(STORE_FACTORY_TYPE, ret_val)  # type: ignore
+        return ret_val
+    else:
+        raise TypeError(
+            f"Provided incompatible store type. Got {type(store)} but expected {STORE_TYPE}."
+        )
--- a/kartothek/core/utils.py
+++ b/kartothek/core/utils.py
@@ -1,3 +1,4 @@
+import pickle
 from functools import partial
 from typing import cast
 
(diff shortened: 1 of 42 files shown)


marco-neumann-by:
What's the reasoning behind this change? It seems to me that the user-facing API gets more unclear now.

> This obviously assumes that KV stores can be properly pickled but I don't see this as a relevant issue anymore. If this is an issue for anybody I can revert that change and only allow store URLs.

I think pickling isn't tested in simplekv except for the azure stores. I don't know for example if this is working for the S3 implementation, redis or any of the other stores in simplekv. Also, being pickle-compatible isn't a requirement of the simplekv interface.

fjetter:
> What's the reasoning behind this change? It seems to me that the user-facing API gets more unclear now.

The reasoning is that store factories are extremely inconvenient. As you are aware yourself, it is also not straight forward to construct "a good callable" to act as a factory. On top of this, I find myself in ad-hoc situation always writing the same imports to construct something which is accepted by kartothek.
At the very least we should support store urls as proposed in #233 and construct this properly ourselves.

> I think pickling isn't tested in simplekv except for the azure stores.

That's true. As I said, if the direct `KeyValueStore` is not an option, I'm willing to remove this again but I do have a fairly strong opinion about adding a more convenient user interface. Alternatively, we could also test if the store is serializable and issue an appropriate and helpful exception.
I could also see this to be implemented in simplekv by flagging supported stores, e.g. with a mixin.

marco-neumann-by:
OK, improving the API (esp. for ad-hoc use cases) sound good. For URLs, which URL format are we referring to? I somehow fair that people in the end confuse the official azure URLs and the storefact format (same is likely true for S3).

>  I could also see this to be implemented in simplekv by flagging supported stores, e.g. with a mixin.

Yeah, sound good to me. Feel free to open a ticket on their side.

fjetter:
Opened a ticket on simplekv side to extend the interface https://github.com/mbr/simplekv/issues/107

We can either continue here by not allowing any kind of KV stores until this issue is resolved or we can test whether it is technically possible and continue. If it is not possible, we'd raise an appropriate error message. Opinions?

marco-neumann-by:
> we can test whether it is technically possible and continue.

How do you want to do that?

fjetter:
> How do you want to do that?

```python
try:
    pickle.loads(pickle.dumps(store))
except Exception:
    raise RuntimeError("Store is not serializable. Please visit <link to docs> for more information.")
```

This obviously doesn't cover semantic issues with serialization

kartothek/core/utils.py
```diff
@@ -46,3 +49,51 @@ def ensure_string_type(obj):
...
+    * Callable[[], KeyValueStore]
+    * KeyValueStore
+
+    If a KeyValueStore is provided, it is verified that the store is serializable
+    """
+    if callable(store):
+        return cast(StoreFactory, store)
+    elif isinstance(store, KeyValueStore):
+        try:
+            pickled_store = pickle.dumps(store)
```
fjetter, line 85: cc @marco-neumann-by

marco-neumann-by (approved):

PR #346, opened 2020-09-15
Allow filtering missings using predicates

### LINKED ISSUE #296: Filter out missings using predicates
We have a setting where we would like to filter out missings using predicates. As far as I understand this is currently not supported. Is this an active decision (i.e. is there a reason to explicitly not allow this)? Is there a general interest for this feature? How would this be implemented (API / technical implementation)? AFAIK partitioning on columns with missings is not supported, so it would be enough to add some code [here](https://github.com/JDASoftwareGroup/kartothek/blob/9a5c040fce316922ca3dd6053a1892652c2881da/kartothek/serialization/_generic.py#L428) (?).

mlondschien:
Closes #296.

Sorry that I did not get back to this earlier. This implements a simple solution as suggested by @lr4d to filter by missings in predicates. This will not work with `predicate_pushdown_to_io` (see test). I am using `pd.isnull` instead of `np.isnan` as the former works for object columns. Passing `[[("x", "==", np.nan)]]` for an object column will raise an error due to a dtype mismatch, not sure what the expected behaviour would be here. Also not sure what the expected behaviour would be for `<`, `<=`, `>`, `>=` operators together with missings.

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -445,10 +445,18 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
     if isinstance(val, float):
         min_value -= _epsilon(min_value)
         max_value += _epsilon(max_value)
+
+    # TODO: What to do for "<=", "<", ">", ">=" np.nan?
     if op == "==":
-        return (min_value <= val) and (max_value >= val)
+        if np.isnan(val):
+            return parquet_statistics.null_count > 0
+        else:
+            return (min_value <= val) and (max_value >= val)
     elif op == "!=":
-        return not ((min_value >= val) and (max_value <= val))
+        if np.isnan(val):
+            return parquet_statistics.null_count < row_meta.num_rows
+        else:
+            return not ((min_value >= val) and (max_value <= val))
     elif op == "<=":
         return min_value <= val
     elif op == ">=":
@@ -463,7 +471,9 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
         # We accept the predicate if there is any value in the provided array which is equal to or between
         # the parquet min and max statistics. Otherwise, it is rejected.
         for x in val:
-            if min_value <= x <= max_value:
+            if np.isnan(x) and parquet_statistics.null_count > 0:
+                return True
+            elif min_value <= x <= max_value:
                 return True
         return False
     else:
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -360,6 +360,8 @@ def _normalize_value(value, pa_type, column_name=None):
             return value.decode("utf-8")
         elif isinstance(value, str):
             return value
+        elif value is None:
+            return value
     elif pa.types.is_binary(pa_type):
         if isinstance(value, bytes):
             return value
@@ -448,12 +450,12 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
 
     # TODO: What to do for "<=", "<", ">", ">=" np.nan?
     if op == "==":
-        if np.isnan(val):
+        if pd.isnull(val):
             return parquet_statistics.null_count > 0
         else:
             return (min_value <= val) and (max_value >= val)
     elif op == "!=":
-        if np.isnan(val):
+        if pd.isnull(val):
             return parquet_statistics.null_count < row_meta.num_rows
         else:
             return not ((min_value >= val) and (max_value <= val))
@@ -471,8 +473,9 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
         # We accept the predicate if there is any value in the provided array which is equal to or between
         # the parquet min and max statistics. Otherwise, it is rejected.
         for x in val:
-            if np.isnan(x) and parquet_statistics.null_count > 0:
-                return True
+            if pd.isnull(x):
+                if parquet_statistics.null_count > 0:
+                    return True
             elif min_value <= x <= max_value:
                 return True
         return False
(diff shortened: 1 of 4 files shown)


fjetter:
Thanks for looking into this. As discussed in #296 this would only apply for the filtering here but would not account for partition pruning, and as you said for predicate pushdown.

I think we still would need to address the `predicate_pushdown_to_io` path. Otherwise this would give very weird errors, I guess. I would suggest to implement this in
https://github.com/JDASoftwareGroup/kartothek/blob/cbbed47d8154e8a2925db2063a9bc37e5f41a399/kartothek/serialization/_parquet.py#L345
and/or
https://github.com/JDASoftwareGroup/kartothek/blob/cbbed47d8154e8a2925db2063a9bc37e5f41a399/kartothek/serialization/_parquet.py#L214
The idea would be: if there is a literal including a nan which is relevant for the RG, always load the RG. This way would load the entire file but still filter what the user requested in the second filter layer.

mlondschien:
I added some logic for predicate pushdown. I am still not sure what the behaviour should be for operators `<`, `<=`, `>` and `>=` and null values. Just throw an error? Note that the code in `_generic.py` isn't especially pretty. Also note that with this implementation, using a `[[("column", "==", np.nan)]]` for an object column is disallowed.

mlondschien:
Test failures are unrelated.

fjetter:
>  I am still not sure what the behaviour should be for operators <, <=, > and >= and null values.
Raise a ValueError would be my suggestion. For nans this is simply not defined. If we were to consider `np.inf` that'd be a different story

tests/serialization/test_dataframe.py
```diff
@@ -530,6 +530,81 @@ def test_predicate_pushdown_null_col(
...
+        key,
+        predicate_pushdown_to_io=predicate_pushdown_to_io,
+        predicates=predicates,
+    )
+    pdt.assert_frame_equal(
+        result.reset_index(drop=True),
+        expected.reset_index(drop=True),
+        check_dtype=serialiser.type_stable,
+    )
+
```
fjetter, line 606: I'm positively surprised to see this changed. I was working under the assumption that the null_count is not reliable (I'm pretty sure this wasn't back in earlier pyarrow versions). Just to satisfy my paranoia, could you please add the following test to verify our assumptions about the null count statistics? (Feel free to adapt if you see improvements)

```suggestion

def test_null_count(store):
    serialiser = ParquetSerializer(chunk_size=2)
    df = pd.DataFrame(
        {
            "partial_nulls_int": [1, 2, 3, None],
            "partial_nulls_float": [1.0, 2.2, 3.3, np.nan],
            "partial_nulls_obj": [1.0, 2.2, 3.3, np.nan],
            "partial_nulls_obj_mixed": [1.0, 2.2, None, np.nan],
            "nulls_reverse_rg": [3.3, np.nan, 1.0, 2.0],
        }
    )
    key = serialiser.store(store, "prefix", df)

    reader = pa.BufferReader(store.get(key))
    parquet_file = ParquetFile(reader)

    assert parquet_file.num_row_groups == 2

    rg_0 = parquet_file.metadata.row_group(0)
    for ix in range(0, 4):
        assert rg_0.column(ix).statistics.null_count == 0
    assert rg_0.column(4).statistics.null_count == 1

    rg_1 = parquet_file.metadata.row_group(1)
    for ix in range(0, 3):
        assert rg_1.column(ix).statistics.null_count == 1
    assert rg_1.column(3).statistics.null_count == 2
    assert rg_1.column(4).statistics.null_count == 0
```

kartothek/serialization/_parquet.py
```diff
@@ -445,10 +447,18 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
     if isinstance(val, float):
         min_value -= _epsilon(min_value)
         max_value += _epsilon(max_value)
+
+    # TODO: What to do for "<=", "<", ">", ">=" np.nan?
```
fjetter, line 451: We should have a "verify valid predicate" function somewhere. I would implement this check there. Nulls are not a valid input for these operators.

fjetter: It's called `check_predicates` in `_generic.py`

fjetter (commented):
>  using a [[("column", "==", np.nan)]] for an object column is disallowed.

I think this is a valid behaviour at the momement. I don't want to open the topic of casting predicate literal values at the moment so this is fine.
We might want to document the null behaviour somewhere. Maybe as part of the argument docs https://github.com/JDASoftwareGroup/kartothek/blob/cbbed47d8154e8a2925db2063a9bc37e5f41a399/kartothek/core/docs.py#L118 ? We currently do not have a dedicated docs page describing the predicates themselves, I believe

mlondschien:
I added the test, the check in `check_predicates` and expanded the docs. Where is `check_predicates` tested?

fjetter:
> Where is check_predicates tested?

well, we have very indirect tests for this in `test_dataframe.py::test_predicate_pushdown`

fjetter:
Feel free to add a test to the existing test_predicate_pushdown. If you want to leave the world a better place, you can also refactor the tests but for this contribution I'd be ok with just adding it to the existing ones.

fjetter:
otherwise LGTM. I'll need to create another release anyhow due to #351 and we can probably include this in the next release

mlondschien:
Thanks!

> Feel free to add a test to the existing test_predicate_pushdown. 

I guess indirect test exist already within `test_predicate_parsing_null_values`, correct? What would you like to see added to the `test_predicate_pushdown` that is not covered in `test_predicate_parsing_null_values`?

>  If you want to leave the world a better place

Always!

>  you can also refactor the tests

Maybe not today.

> otherwise LGTM. I'll need to create another release anyhow due to #351 and we can probably include this in the next release

Again, thanks!

fjetter (approved):
Thanks @mlondschien !

PR #347, opened 2020-09-22
Enable shuffle in build_cube_from_dataframe

### LINKED ISSUE #213: Add store_dataset_from_ddf
We currently only have `update_dataset_from_ddf` implemented whereas other modules also have `store_` prefixed functions, e.g. https://github.com/JDASoftwareGroup/kartothek/blob/b44ccbb45c9c00ba52291a95190086e918568404/kartothek/io/dask/delayed.py#L490 It would be nice to have this as there are semantic differences between update ("append/modify a possibly existing dataset") and store ("(re)write a complete dataset).

fjetter:
# Description:

This performs the first marriage between the cube and traditional kartothek functionality by essentially hooking in all additional sanity checks performed for the cubes but calling standard kartothek in the background. To make this smooth-ish I needed to introduce a few changes

* `sort_partitions_by` now must support more than one column. That's fine.
* I introduced a new exception type `MultiTableCommitException` which indicates that a rollback is triggered. I believe this gives much better transparency about what is happening. This is not necessarily required for this change but helped me a lot in figuring out where, what broke
* To get semantics right, I introduced `kartothek.io.dask.dataframe.store_dataset_from_ddf` which has been a feature request in #213 for a while now
* The most important change is the behaviour for null values in indices now. Old kartothek simply dropped nulls and indexed it accordingly. I believe this behaviour is flawed and I adopted the "raise if null encountered" approach as is already used by the cubes. That's a breaking change in behaviour, of course, and we need to document this properly. (cc @mlondschien you've been working with nulls recently. This may interest you)

- [x] Closes #213
- [ ] Changelog entry


bjoern-meier-by (approved):
Looks good to me. The only thing I have to note is the test coverage of the introduced `store_dataset_from_ddf` which misses some parts https://codecov.io/github/JDASoftwareGroup/kartothek/commit/8d4c32f4e305391866f41be27ccb628869fd8962

kartothek/io_components/metapartition.py
```diff
@@ -1229,8 +1229,11 @@ def build_indices(self, columns):
             possible_values = set()
             col_in_partition = False
             for df in self.data.values():
+
                 if col in df:
-                    possible_values = possible_values | set(df[col].dropna().unique())
+                    if df[col].hasnans:
```
fjetter, line 1234: @xhochy you might have an opinion about this as well

xhochy: Sounds legit to disallow NULLs here. There is a possibility to encode nulls so that other tools also understand that but that turned out to be a headache for them, so raising is best.

fjetter: This is not about partitioning but indexing. Therefore, there is no encoding since ktk is the only user of these values.

partitioning raises already, the question is whether or not secondary indices should simply drop these. I see the possibility to implement this properly with `pd.NA`once this really works but until then this is a bit of no-mans land. So far, we silently dropped it

xhochy: Ah, I first defer then to @mlondschien before I make new comment.

fjetter:
@jochen-ott-by you mentioned running into this inconsistency as well. this might interest you as well

fjetter:
I will revert the index change. I don't feel confident that this will go through smoothly. It may also be worth enabling this fully in the future but we'd need more tests for this which is out of scope for this change

kartothek/io/dask/dataframe_cube.py
```diff
@@ -19,49 +37,94 @@
...
+    num_buckets: int = 1,
+    bucket_by: Optional[Iterable[str]] = None,
+) -> Delayed:
     """
     Create dask computation graph that builds a cube with the data supplied from a dask dataframe.
 
     Parameters
     ----------
-    data: Union[dask.DataFrame, Dict[str, dask.DataFrame]
+    data:
```
marco-neumann-by, line 57:
```suggestion
    data
```

Otherwise, sphinx will parse `Data that should be written...` (the next line) as a type, I think.

kartothek/io_components/cube/write.py
```diff
@@ -372,6 +401,10 @@ def multiplex_user_input(data, cube):
     return data
 
 
+class MultiTableCommitException(Exception):
```
marco-neumann-by, line 404: docstring

CHANGES.rst
```diff
@@ -5,6 +5,19 @@ Changelog
 Version 3.14.x (unreleased)
 ===========================
 
+New functionality
+^^^^^^^^^^^^^^^^^
+* Add :func:``~kartothek.io.dask.dataframe.store_dataset_from_ddf`` to offer write
```
marco-neumann-by, line 10: Thanks :)

kartothek/io_components/cube/write.py
```diff
@@ -201,18 +233,15 @@ def _check_user_df(ktk_cube_dataset_id, df, cube, existing_payload, partition_on
...
-        for col in sorted(cols):
-            if df[col].isnull().any():
-                raise ValueError(
-                    'Found NULL-values in {what} column "{col}" of dataset "{ktk_cube_dataset_id}"'.format(
-                        col=col, ktk_cube_dataset_id=ktk_cube_dataset_id, what=what
-                    )
-                )
+    # Factor this check out. All others can be performed on the dask.DataFrame.
+    # This one can only be executed on a pandas DataFame
+    if isinstance(df, pd.DataFrame):
```
marco-neumann-by, line 238: So this check is just silently skipped in dask dataframes?

marco-neumann-by (changes requested):

kartothek/io/dask/dataframe_cube.py
```diff
@@ -19,49 +37,94 @@
...
+            indices_to_build |= set(cube.dimension_columns)
+        indices_to_build -= set(partition_on_checked[table_name])
+
+        ddf = ddf.map_partitions(
+            assert_dimesion_index_cols_notnull,
+            ktk_cube_dataset_id=table_name,
             cube=cube,
+            partition_on=partition_on_checked[table_name],
+            meta=ddf._meta,
+        )
```
fjetter, line 107: @marco-neumann-by Regarding your other comment about the index null check

kartothek/io_components/cube/write.py
fjetter: see other comment

kartothek/io/dask/dataframe_cube.py
fjetter: This was the easiest way to not screw up everything else. Eventually we should consolidate the kartothek dataset behaviour by either providing full nullable support or raising there as well. Deferred this to another PR at a later time

kartothek/io_components/cube/write.py
fjetter: Wasn't that sure about the name of the exception but I felt it pretty convenient during testing to see whether something raised during execution or during a post write check since the latter triggers a rollback, which might be interesting for debugging purposes and might not be obvious to a user. Any thoughts about a good name? Will add a doc string if we keep it, of course

marco-neumann-by: `MultiTableCommitAborted` maybe? IMHO sing the word `Exception` or `Error` within the class-name is kinda pointless (except when the class is so abstract that it demands it). Also, I would subclass from `RuntimeError` to be more precise.

kartothek/io/dask/dataframe_cube.py
marco-neumann-by: > Eventually we should consolidate the kartothek dataset behaviour by either providing full nullable support or raising there as well.

I would rather raise now (aka be more restrictive) and harden + enable later (aka implement a feature) than having a half-tested inconsistent solution now and a potential really ugly breaking change later (aka the other way around). 

Some background: The actual reason why this if forbidden in the cube functionality is not because I deemed it to be an anti-feature but because I wasn't able to build a properly tested feature around it. I also wasn't concerned too much about the index data structure, but about the usage (aka within conditions) where NULL can mean different things, here for non-seed tables:

- A: index data is actual NULL
- B: (very similar) non-indexed payload data is actual NULL
- C: data absent in physical partition list (=primary index)
- D: data absent in secondary index
- E: data absent within parquet file (but both the index and/or the physical partition list hints it could be there), only happens when secondary indices are outdated
- F: data cut out due to table merge (e.g. `(C(column_in_table_a) != 20) & (C(column_in_table_b).isna())`, where both tables have data corresponding to `C(column_in_table_a) == 20`).

Now all these cases are slightly different. Most of them lead to NULL in the resulting dataframe, but the the are created at very different stages of the query processing and getting all these cases right requires some thought and additions to the query system (e.g. the way it is designed, you cannot trivially add case F with the current design).

fjetter: Raising now in kartothek single-tables is a hard breaking change which might surprise users. I'm not willing to deal with this just now. The changes as proposed by this PR would not change any behaviour.

Whether or not it is possible to provide this as a full fledged feature is something I cannot tell at the moment and I don't intend to introduce something like this lightly. Either way, this is out of scope for this change.

marco-neumann-by: This is not about single-table. The whole change here concerns cubes and there I would like to know why there is a different behavior between the bag+eager and the DDF interface.

marco-neumann-by: clarified offline

marco-neumann-by (approved):

PR #349, opened 2020-09-28
Deprecation warning for multi table support

imkumarg:
# Description:

Updated the read functions in kartothek (`read_table.*` and `read_dataset_as.*`) with a check to provide the deprecation warning for the multi table support.

Closes part of #317 
Closes #337


fjetter (changes requested):
An alternative approach would be to check  on the `ds_factory` object whether or not the dataset to be read is multi tabled.
e.g.

```
if len(ds_factory.tables) > 1:
    warning.warn("Trying to read a dataset with multiple internal tables. This functionality will be removed in the next major release. If you require a multi tabled data format, we recommend to switch to the kartothek Cube functionality <link to docs>", DeprecationWarning)
```

This warning should be issued irregardless of what the user input is. If the dataset itself is built this way, it should raise the warning.

kartothek/io/dask/dataframe.py
```diff
@@ -66,6 +66,12 @@ def read_dataset_as_ddf(
 
         For details on performance, see also `dispatch_by`
     """
+    if table is not SINGLE_TABLE or isinstance(columns, dict):
```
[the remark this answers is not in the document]
imkumarg: > An alternative approach would be to check on the `ds_factory` object whether or not the dataset to be read is multi tabled.
> e.g.
> 
> ```python
> if len(ds_factory.tables) > 1:
>     warning.warn("Trying to read a dataset with multiple internal tables. This functionality will be removed in the next major release. If you require a multi tabled data format, we recommend to switch to the kartothek Cube functionality <link to docs>", DeprecationWarning)
> ```
> 
> This warning should be issued irregardless of what the user input is. If the dataset itself is built this way, it should raise the warning.

Yes it is good to go with this check, I will incorporate this changes.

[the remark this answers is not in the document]
imkumarg: > We don't necessarily need to assert for the table name to be `SINGLE_TABLE`. Even if the table name would be `my_own_table_name` we can easily infer this during read and this is no issue.

Noted, Will remove the check.

kartothek/io/dask/bag.py
```diff
@@ -68,6 +69,12 @@ def read_dataset_as_metapartitions_bag(
     -------
     A dask.bag object containing the metapartions.
     """
+    if tables is not None or isinstance(columns, dict):
```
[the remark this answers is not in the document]
imkumarg: The check is done if user wants to read more than single table. It is made sure that the `tables` must not be more than single table.

We can change the condition to `tables is None and isinstance(columns, dict)`. For this case a warning can be raised. Please suggest over it?

kartothek/io/dask/dataframe.py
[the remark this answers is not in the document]
lr4d: @fjetter [here](https://github.com/JDASoftwareGroup/kartothek/pull/337#issuecomment-675347867) you mentioned 
>  the functions which only work on a single table (e.g. everything in dask.dataframe and something like eager.read_table) which should issue a warning if the requested table is not the SINGLE_TABLE constant.

which is why I added that as a task in the issue page [here](https://github.com/JDASoftwareGroup/kartothek/issues/317#issuecomment-686382330)

Am I mixing things up or does this no longer apply?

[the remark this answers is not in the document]
fjetter: >  here you mentioned

Yes this is what I wrote. Honestly, I can see reasons for both ways. I also mentioned that we shouldn't spam deprecation warnings without a clear migration path. Deprecations should include a suggestion on how to "remove the deprecation". What will be the replacement of a given deprecated feature? Whether or not we warn if the input is or isn't SINGLE_TABLE depends on what the replacement will be.
For instance, a possible replacement would be "as long as the dataset is single-tabled, ignore user input and just infer from dataset" That's what I suggested in #317. Another possibility could be to issue a warning telling the user to provide SINGLE_TABLE (or nothing at all, etc.)
Depending on what the replacement should be, we can formulate the deprecation warnings.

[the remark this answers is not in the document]
imkumarg: Okay @fjetter, for this PR i will not raise the warnings for the single table and will do that in the follow up PR.

kartothek/io/eager.py
```diff
@@ -131,6 +131,14 @@ def read_dataset_as_dataframes(
         load_dataset_metadata=True,
     )
 
+    if len(ds_factory.tables) > 1:
```
lr4d, line 134: This will cause the warning to appear twice as you are also raising the warning in read_dataset_as_metapartitions

kartothek/io/iter.py
```diff
@@ -159,6 +167,14 @@ def read_dataset_as_dataframes__iterator(
             {'core': pd.DataFrame, 'lookup': pd.DataFrame}
         ]
     """
+    if tables is None and isinstance(columns, dict):
```
lr4d, line 170: I think we can remove this and also use `if len(ds_factory.tables) > 1:` here, to make sure we always raise the warning if a user reads a multi-tabled dataset.
(You'll have to define `ds_factory = _ensure_factory(...)` before)

lr4d (commented):
a test for this would also be nice

kartothek/io/eager.py
imkumarg: Got it @lr4d, I will make necessary changes.

kartothek/io/iter.py
imkumarg: Okay will define the `ds_factory` and make the changes.

imkumarg:
> a test for this would also be nice

Added test case to check for deprecation warning for multi table support.

lr4d:
wrt testing I'd recommend to use `pytest.warns` (see https://docs.pytest.org/en/6.1.1/warnings.html#assertwarnings) to check for the warnings being raised in the appropriate scenarios (multi-table datasets, columns/categoricals dicts), it should suffice to put such a test in `kartothek.io.testing.read` for it to apply to all back-ends

kartothek/io/dask/bag.py
```diff
@@ -151,6 +153,14 @@ def read_dataset_as_dataframe_bag(
     dask.bag
         A dask.bag which contains the metapartitions and mapped to a function for retrieving the data.
     """
+    if isinstance(columns, dict) or isinstance(categoricals, dict):
+        warnings.warn(
+            "Trying to read a dataset with multiple internal tables. This functionality will be removed in the next "
+            "major release. If you require a multi tabled data format, we recommend to switch to the kartothek Cube "
+            "functionality https://kartothek.readthedocs.io/en/stable/index.html",
```
fjetter, line 160: You should point to the appropriate section. If we just point to the docs in general, people will get lost.

tests/io/dask/bag/test_read.py
```diff
@@ -65,3 +65,31 @@ def test_read_dataset_as_dataframes_partition_size(store_factory, metadata_versi
             dataset_uuid="partitioned_uuid", store=store_factory, partition_size=2
         )
         assert bag.npartitions == 2
+
+
+def test_read_dataset_as_dataframe_bag_multi_table_warning(store_factory, metadata_version):
```
fjetter, line 70: This test doesn't verify the warning, does it?

fjetter (changes requested):

tests/io/dask/bag/test_read.py
imkumarg: I am on it to verify it with https://docs.pytest.org/en/stable/warnings.html#warns

kartothek/io/dask/bag.py
imkumarg: Sure I will provide the appropriate section link.

imkumarg:
Added test case applicable for all backends (dask [`bag`, `dataframe`, `delayed`], eager, iter) to validate the multi table deprecation warning. Also updated the document link in the warning message.

kartothek/io/dask/bag.py
```diff
@@ -76,6 +76,15 @@ def read_dataset_as_metapartitions_bag(
         load_dataset_metadata=load_dataset_metadata,
     )
 
+    if len(ds_factory.tables) > 1:
+        warnings.warn(
+            "Trying to read a dataset with multiple internal tables. This functionality will be removed in the next "
+            "major release. If you require a multi tabled data format, we recommend to switch to the kartothek Cube "
+            "functionality. "
+            "https://kartothek.readthedocs.io/en/stable/guide/cube/cube_design_features.html?highlight=multi%20table#multi-dataset-vs-multi-table",
```
fjetter, line 84: Please just point to https://kartothek.readthedocs.io/en/stable/guide/cube/kartothek_cubes.html

imkumarg: Done.

fjetter:
for some reason the CI doesn't seem to run. Maybe because the state of your branch is too old. Can you please rebase and resolve the merge conflicts?

kartothek/io/testing/read.py
```diff
@@ -723,3 +723,29 @@ def test_extensiondtype_rountrip(store_factory, bound_load_dataframes):
...
+                "aux-table": pd.DataFrame({"id": [31], "col1": ["y"]}),
+            }
+        },
+    ]
+
+    dm = store_dataframes_as_dataset(
+        dfs=dfs, store=store_factory, dataset_uuid="dataset_uuid"
+    )
+
+    with pytest.warns(DeprecationWarning):
```
lr4d, line 748: np
```suggestion
    with pytest.warns(DeprecationWarning, match="Trying to read a dataset with multiple internal tables.*"):
```

imkumarg: Done.

lr4d:
CI is failing due to black

tests/io/eager/test_read.py
```diff
@@ -26,7 +26,8 @@ def output_type(request):
 
 def _read_table(*args, **kwargs):
     if "tables" in kwargs:
-        kwargs.pop("tables")
+        param_tables = kwargs.pop("tables")
+        kwargs["table"] = param_tables
```
lr4d, line 30: same

imkumarg: I am adding `table` here instead of `tables`. If the `table` is not provided the test case will fail for `read_table` in eager.
Also, I need to write a common test case that would be applicable to all the backends. 

https://github.com/JDASoftwareGroup/kartothek/blob/785ec8a1a54aca13f2e7cb076b970ddd0ebb275e/kartothek/io/eager.py#L300

tests/io/dask/delayed/test_read.py
```diff
@@ -23,7 +23,8 @@ def _load_dataframes(output_type, *args, **kwargs):
         func = read_dataset_as_delayed_metapartitions
     elif output_type == "table":
         if "tables" in kwargs:
-            kwargs.pop("tables")
+            param_tables = kwargs.pop("tables")
+            kwargs["table"] = param_tables
```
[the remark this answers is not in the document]
imkumarg: A quite similar scenario as explained in below comment.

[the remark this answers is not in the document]
lr4d: Just see that it's table*s*

imkumarg:
> Congrats on your first contribution ;)

Thanks, @lr4d

PR #352, opened 2020-09-30
Fix get_parquet_metadata sometimes returning empty statistics

fjetter:
# Description:

This addresses https://issues.apache.org/jira/browse/ARROW-10146 by not relying on the AttributeError. I encountered hitting this for non-empty files which would otherwise cause wrong statistics. I couldn't track down what makes those files special, therefore no test :(

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Version 3.16.1 (2020-10-??)
+===========================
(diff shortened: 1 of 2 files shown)


lr4d (approved):
Too bad we can't rely on the "nice" pyarrow method and have to construct this dictionary ourselves.

I re-triggered the CI because there was one job failing apart from codecov, apparently stuck at conda env creation.

PR #353, opened 2020-09-30
Do not call get_parquet_metadata at graph construction time

fjetter:
# Description:

Without this, the function is already executed once during graph construction to determine the schema

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


lr4d (approved):
good job. How did you catch this?

fjetter:
> good job. How did you catch this?

Watching the dashboard you'd notice a single computation pop up before you even hit compute. Of course, you'll need to know that dask "tests" a function if you don't provide the meta...

PR #357, opened 2020-10-09
Addition of repartition ratio calculation and optimize functionality to Ktk

usha-nemani-by:
# Description:

Addition of  calculation of  re-partition ratio and optimize functionality to Ktk

- [x] Changelog entry

### DIFF
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,3 +1,4 @@
+import math
 import random
 import warnings
 from typing import (
@@ -13,7 +14,6 @@ from typing import (
 
 import dask
 import dask.dataframe as dd
-import numpy as np
 import pandas as pd
 from simplekv import KeyValueStore
 
@@ -332,9 +332,7 @@ def _write_dataframe_partitions(
     bucket_by: List[str],
 ) -> dd.Series:
     if repartition_ratio and ddf is not None:
-        ddf = ddf.repartition(
-            npartitions=int(np.ceil(ddf.npartitions / repartition_ratio))
-        )
+        ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
 
     if ddf is None:
         mps = dd.from_pandas(
@@ -606,3 +604,67 @@ def hash_dataset(
             .apply(_unpack_hash, unpack_meta=ddf._meta, subset=subset, meta="uint64")
             .astype("uint64")
         )
+
+
+def calculate_repartition_ratio(
+    dataset_factory: DatasetFactory,
+    predicates: Optional[PredicatesType] = None,
+    *,
+    target_nrows_per_partition: int,
+    dataset_metadata_frac: float,
+) -> dd.core.Scalar:
+    """
+    Calculate a `repartition_ratio` according to the dataset input and `target_nrows_per_partition` (the desired
+    number of rows of each dataframe partition).
+
+
+    Parameters
+    ----------
+    dataset_factory
+        dataset factory
+    predicates
+        Optional , but predicates to be applied during calculatin repartition ratio
+    target_nrows_per_partition
+        target num of rows per partition
+    dataset_metadata_frac
+        used during collection of dataset metadata
+
+    Returns
+    -------
+    repartition_ratio
+        calculated repartition_ratio
+    """
+    dataset_metadata = collect_dataset_metadata(
+        factory=dataset_factory, predicates=predicates, frac=dataset_metadata_frac,
+    )
+    expected_nrows_per_partition = dataset_metadata.groupby("partition_label")[
+        "number_rows_per_row_group"
+    ].sum()
+
+    mean_nrows_per_partition = expected_nrows_per_partition.mean()
+    repartition_ratio = target_nrows_per_partition / mean_nrows_per_partition
+    return repartition_ratio
+
+
+def repartition_ddf(
+    ddf: dd.DataFrame, repartition_ratio: SupportsFloat
+) -> dd.DataFrame:
+    """
+    Repartition a dask DataFrame according, dividing the current number of partitions by `repartition_ratio`.
+
+    Parameters
+    ----------
+    ddf
+        dask.Dataframe
+    repartition_ratio
+        calculated repartition ratio
+
+    Returns
+    -------
+    dask.DataFrame
+    """
+    with dask.config.set({"optimization.fuse.ave-width": math.ceil(repartition_ratio)}):
+        # Minimum amount of partitions is 1.
+        target_npartitions = max(1, int(math.ceil(ddf.npartitions / repartition_ratio)))
+        ddf2 = ddf.repartition(npartitions=target_npartitions)
+        return dask.optimize(ddf2, optimizations=[dask.dataframe.optimize])[0]
--- a/tests/io/dask/dataframe/test_repartition.py
+++ b/tests/io/dask/dataframe/test_repartition.py
@@ -1,0 +1,106 @@
+import dask
+import pandas as pd
+import pytest
+
+from kartothek.core.factory import DatasetFactory  # noqa: E402
+from kartothek.io.dask.dataframe import (
+    calculate_repartition_ratio,
+    read_dataset_as_ddf,
+    repartition_ddf,
+)
+from kartothek.io.iter import store_dataframes_as_dataset__iter
+
+
+@pytest.mark.parametrize("frac", (1.0, 0.5))
+@pytest.mark.parametrize(
+    "df,target_nrows_per_partition",
+    (
+        (
+            pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}),
+            100,
+        ),  # rep_ratio = 4
+        (
+            pd.DataFrame({"A": range(2 ** 7), "part": ["A", "B", "C", "D"] * 2 ** 5}),
+            8,
+        ),  # rep_ratio = 0.25
+    ),
+)
+def test_calculate_repartition_ratio_exact_match(
+    df, target_nrows_per_partition, frac, store_factory
+):
+    """
+    Because we carefully choose the number of original partitions and the value of
+    `target_nrows_per_partition`, we get exact matches for `target_nrows_per_partition`.
+    This is not expected in a real-world scenario where the conditions are very unlikely.
+    """
+    dfs = [df]
+    store_dataframes_as_dataset__iter(
+        dfs, store=store_factory, dataset_uuid="test", partition_on=["part"],
+    )
+
+    repartition_ratio = calculate_repartition_ratio(
+        dataset_factory=DatasetFactory(
+            dataset_uuid="test", store_factory=store_factory
+        ),
+        dataset_metadata_frac=frac,
+        target_nrows_per_partition=target_nrows_per_partition,
+    ).compute()
+
+    ddf = read_dataset_as_ddf(store=store_factory, dataset_uuid="test")
+    ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+
+    assert (ddf.map_partitions(len).compute() == target_nrows_per_partition).all()
+
+
+@pytest.mark.parametrize(
+    "df,target_nrows_per_partition",
+    (
+        (
+            pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}),
+            97,
+        ),  # rep_ratio = 3.88
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 1),  # rep_ratio = 1.0
+    ),
+)
+def test_calculate_repartition_ratio_approximate_match(
+    df, target_nrows_per_partition, store_factory
+):
+    dfs = [df]
+    store_dataframes_as_dataset__iter(
+        dfs, store=store_factory, dataset_uuid="test", partition_on=["part"],
+    )
+
+    repartition_ratio = calculate_repartition_ratio(
+        dataset_factory=DatasetFactory(
+            dataset_uuid="test", store_factory=store_factory
+        ),
+        dataset_metadata_frac=0.3,
+        target_nrows_per_partition=target_nrows_per_partition,
+    ).compute()
+
+    ddf = read_dataset_as_ddf(store=store_factory, dataset_uuid="test")
+    ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+    # this is somewhat arbitrary as original partition size may be non-constant
+    margin_of_error = 0.5  # proportion
+    partition_lengths = ddf.map_partitions(len).compute()
+    assert (
+        ((1 - margin_of_error) * target_nrows_per_partition <= partition_lengths)
+        & (partition_lengths <= (1 + margin_of_error) * target_nrows_per_partition)
+    ).all()
+
+
+@pytest.mark.parametrize("repartition_ratio", (3.88, 0.25, 7, 1.0))
+def test_optimize_with_diff_repartition_ratios(repartition_ratio, store_factory):
+    dfs = [pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25})]
+    store_dataframes_as_dataset__iter(
+        dfs, store=store_factory, dataset_uuid="test", partition_on=["part"],
+    )
+    ddf = read_dataset_as_ddf(store=store_factory, dataset_uuid="test")
+    optimized_ddf = dask.optimize(ddf, optimizations=[dask.dataframe.optimize])[0]
+    repartitioned_ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+    if repartition_ratio >= 1:
+        assert len(repartitioned_ddf.dask) <= len(
+            optimized_ddf.dask
+        )  # assert the number of tasks in dask graph
+    else:
+        assert len(repartitioned_ddf.dask) > len(optimized_ddf.dask)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@
 Changelog
 =========
 
+Version 3.17.0 (2020-10-22)
+===========================
+
+* Kartothek now support functionality to calculate repartition_ratio and repartition dask Dataframe using this ratio.
+    :func:`~kartothek.io.dask.dataframe.calculate_repartition_ratio`
+    :func:`~kartothek.io.dask.dataframe.repartition_ddf`
+
 Version 3.16.1 (2020-10-??)
 ===========================
 


tests/conftest.py
```diff
@@ -553,3 +553,8 @@ def evaluation_dataset(meta_partitions_evaluation_files_only, store_session):
             dataset_uuid="evaluation_uuid",
             store=store_session,
         )
+
+
+@pytest.fixture
+def dataset_factory_function_scope(store_factory):  # Note, this is empty
```
fjetter, line 559: What is this used for if it is empty?

kartothek/utils/repartition.py
```diff
@@ -0,0 +1,54 @@
...
+    table_name: str,
+    predicates: Optional[PredicatesType] = None,
+    *,
+    target_nrows_per_partition: int,
+    dataset_metadata_frac: float,
+) -> dd.core.Scalar:
+    """
+    Calculate a `repartition_ratio` according to the dataset input and `target_nrows_per_partition` (the desired
+    number of rows of each dataframe partition).
+    """
```
fjetter, line 24: Can you please add a more thorough description here, i.e. parameter docs.

kartothek/utils/optimize.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def optimize(graph: Any) -> Any:
+    """Optimize a dask graph using the dask.dataframe optimizer.
+
+    Kartothek may return only a delayed object which has no default optimizers
+    registered with dask. Therefore, we'll call the optimizer with the
+    dask.dataframe optimizer ourselves.
+    """
+    return dask.optimize(graph, optimizations=[dask.dataframe.optimize])[0]
```
fjetter, line 13: Not sure if this should be part of the public ktk API. If so, the function should have a more specific name. I'm rather thinking about calling this internally on our objects before we return, optionally toggle-able. This would then remove the need to expose this

fjetter (changes requested):
These modules should somehow live in a `dask` submodule, not necessarily top-level utils

fjetter:
cc @xhochy 
we briefly had a discussion about a similar request of yours. This logic should be rather easily extendable using the deserialized size stored in the rowgroup metadata to achieve something like `ddf.repartition("100MB")`

tests/conftest.py
usha-nemani-by: it is not empty .it contains test dataset. removed the comment.

kartothek/utils/repartition.py
usha-nemani-by: Updated. Please verify.

kartothek/utils/optimize.py
usha-nemani-by: Removed this file from Ktk. and updated the Return description.. That we can apply `dask.optimize ` if required.

usha-nemani-by:
As per @xhochy  discussion, I have updated the signature of `repartition_ddf()` to take `repartition_ratio` and `parquet_size` as inputs. 
Need to write new testcases to verify the inputs.

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+        factory=dataset_factory,
+        predicates=predicates,
+        table_name=table_name,
+        frac=dataset_metadata_frac,
+    )
+    # Once the kartothek issue is resolved, one should use the rowgroup wise
+    # metadata. This would set up future improvements once kartothek is able to
+    # filter RG predicates while collecting statistics (assuming this will ever
+    # be possible this way)
+    # See https://github.com/JDASoftwareGroup/kartothek/pull/352
```
lr4d, line 51: The PR linked here is merged. Can we make the change suggested in the comment?

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+    # and https://issues.apache.org/jira/browse/ARROW-10146
+    expected_nrows_per_partition = dataset_metadata[
+        ["partition_label", "number_rows_total"]
+    ].drop_duplicates()
+    mean_nrows_per_partition = expected_nrows_per_partition.number_rows_total.mean()
+    repartition_ratio = target_nrows_per_partition / mean_nrows_per_partition
+    return repartition_ratio
+
+
+def repartition_ddf(ddf: dd.DataFrame, **kwargs) -> dd.DataFrame:
```
lr4d, line 61: why `**kwargs` instead of specifying the arguments directly?

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+        calculated repartition ratio
+    parquet_size
+        deserialized  size of the parquet file.
+    Returns
+    -------
+    ddf
+       dask dataframe after applying repartition on it.
+       We can trigger a dask optimization to fuse the repartitioned tasks properly.
+    """
+    if kwargs["repartition_ratio"]:
```
lr4d, line 79: Please check that no repartition ratio and parquet_size can be provided at the same time, causing an error to the user.

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+            {"optimization.fuse.ave-width": math.ceil(repartition_ratio)}
+        ):
+            # Minimum amount of partitions is 1.
+            target_npartitions = max(
+                1, int(math.ceil(ddf.npartitions / repartition_ratio))
+            )
+            ddf2 = ddf.repartition(npartitions=target_npartitions)
+            return ddf2
+    elif kwargs["parquet_size"]:
+        ddf2 = ddf.repartition(partition_size=kwargs["parquet_size"])
```
lr4d, line 91: I get that we factor the `repartition_ratio` logic out into a separate function, but for `parquet_size` we are just passing the function input directly to `ddf.repartition(partiiton_size=`, I don't see any value this indirection brings atm

lr4d (changes requested):

kartothek/io/dask/repartition.py
usha-nemani-by: we can either send `repartition_ratio` or `parquet_size` to partition the ddf.  since the input is optional, I am sending it in kwargs .

kartothek/io/dask/repartition.py
fjetter: Not only is this indirection misleading but this also would trigger an entire dataset load. My earlier comment was rather we can achieve this repartitioning by size by *not* calling dask repartition but rather do this ourselves. Please remove this from the current PR since this is a non-trivial extension to the logic and we should do this in a dedicated PR

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+        dask dataframe
+    repartition_ratio
+        calculated repartition ratio
+    parquet_size
+        deserialized  size of the parquet file.
+    Returns
+    -------
+    ddf
+       dask dataframe after applying repartition on it.
+       We can trigger a dask optimization to fuse the repartitioned tasks properly.
```
fjetter, line 77: We not only _can_ trigger but we actually do. I would clarify this in the description above already.

tests/conftest.py
```diff
@@ -553,3 +553,8 @@ def evaluation_dataset(meta_partitions_evaluation_files_only, store_session):
             dataset_uuid="evaluation_uuid",
             store=store_session,
         )
+
+
+@pytest.fixture
+def dataset_factory_function_scope(store_factory):
```
fjetter, line 559: Can you please define this in the test_repartition.py? For now this doesn't make sense globally. In particular, there doesn't exist a dataset with UUID "test" but this is created in the test cases explicilty. A fixture should make sense even outside of the context of a particular test case but this is not the case here

fjetter (changes requested):

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    if kwargs["repartition_ratio"]:
+        repartition_ratio = kwargs["repartition_ratio"]
+        with dask.config.set(
+            {"optimization.fuse.ave-width": math.ceil(repartition_ratio)}
```
fjetter, line 82: We should also add tests verifiying that the block fusing here works as expected

kartothek/io/dask/repartition.py
usha-nemani-by: @lr4d ,
What is the RG attribute I should use here . 
We used `number_rows_total` till now. 
Should I use `number_row_groups`* `number_rows_per_row_group` to get total number of rows?

kartothek/io/dask/repartition.py
usha-nemani-by: Removed parquet_size related code. from here. Added arguments explicitly.

kartothek/io/dask/repartition.py
usha-nemani-by: Removed parquet_size related code. from here. Added repartition_ratio only.

kartothek/io/dask/repartition.py
usha-nemani-by: Removed parquet_size related code from here. Added repartition_ratio only.

kartothek/io/dask/repartition.py
usha-nemani-by: done. Please verify.

tests/conftest.py
usha-nemani-by: done. please verify.

kartothek/utils/optimize.py
usha-nemani-by: I directly added `dask.optimize(ddf2, optimizations=[dask.dataframe.optimize])[0]` to be returned , after `repartitioning the ddf` in `repartition_ddf()` function.  Please verify if it is fine.

kartothek/io/dask/repartition.py
lr4d: From the phrasing it sounds like just `number_rows_per_row_group` but I don't see how that applies here as we are loading each Parquet file as a partition, not each row-group. @fjetter mind elaborating a bit about what you meant here?

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,81 @@
...
+    Parameters
+    ----------
+    ddf
+        dask dataframe
+    repartition_ratio
+        calculated repartition ratio
+    Returns
+    -------
+    Optimized Dask Dataframe
+       a dask optimized Dataframe to fuse the repartitioned tasks properly.
```
lr4d, line 75: I don't think this phrasing is very clear. How about you just add to the function docstring that we call `dask.optimize` on the ddf before we return?
```suggestion
    dask.DataFrame
```

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,81 @@
...
+    expected_nrows_per_partition = dataset_metadata[
+        ["partition_label", "number_rows_total"]
+    ].drop_duplicates()
+
+    mean_nrows_per_partition = expected_nrows_per_partition.number_rows_total.mean()
+    repartition_ratio = target_nrows_per_partition / mean_nrows_per_partition
+    return repartition_ratio
+
+
+def repartition_ddf(ddf: dd.DataFrame, repartition_ratio: float) -> Any:
```
lr4d, line 62: Not sure why this wasn't set before if logic was the same
```suggestion
def repartition_ddf(ddf: dd.DataFrame, repartition_ratio: float) -> dd.DataFrame:
```

usha-nemani-by: we have to return `Any`  . I am returning the `dask.otpmize()` result. 
I added Optimization logic here, to the `repartition_ddf ()` .

lr4d: Is the return value not a `dd.DataFrame`?

kartothek/io/dask/repartition.py
fjetter: The original implementation was

```
expected_nrows_per_partition = dataset_metadata.groupby("partition_label")[
        "number_rows_per_row_group"
    ].sum()
```

If the collect metadata would implement predicate pushdown, this would give more accurate metrics. For now it isn't really relevant

kartothek/io/dask/repartition.py
fjetter: The return type is the same as the input type. It might be the case that `dask.optimize` does not expose the types correctly but the output is definitely `dd.DataFrame`

fjetter (changes requested):
Still missing a test about the optimization

kartothek/io/dask/repartition.py
lr4d: > If the collect metadata would implement predicate pushdown

hm, why didn't we support this originally? This would be very useful for repartitioning after reading part of a dataset, perhaps we should pick it up again

kartothek/io/dask/repartition.py
usha-nemani-by: done.please verify

kartothek/io/dask/repartition.py
usha-nemani-by: I added few assertions, Would you please clarify if I am in correct direction ?
But 3 testcases seems to fail. I need some help here, to resolve them.

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,114 @@
...
+    # call optimize here and find the ddf length
+    ddf_before_repartition = dask.optimize(
+        ddf, optimizations=[dask.dataframe.optimize]
+    )[0]
+    # import IPython
+    # IPython.embed()
+    ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+    # find length after repartition
+    # length of ddf after repartition should be less than before.
+    assert len(ddf.dask.keys()) <= len(ddf_before_repartition.dask.keys())
```
fjetter, line 106: Can you please add this as a dedicated test? This is quite an important feature of this repartitioning and should not be squeezed in as a simple assert. Writing a dedicated test with explicit parameters for the use case we're about to test is much easier to debug. Every unit test should test one feature not multiple

fjetter (changes requested):

tests/io/dask/dataframe/test_repartition.py
fjetter: This assert can only be true if the repartition ratio is indeed `>=1` and the fusing works. That the repartition ratio is in the area we expect it to be is not really clear to me at this stage of the test

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,114 @@
...
+    repartition_ratio = calculate_repartition_ratio(
+        dataset_factory=dataset_factory_function_scope,
+        table_name=SINGLE_TABLE,
+        dataset_metadata_frac=frac,
+        target_nrows_per_partition=target_nrows_per_partition,
+    ).compute()
+
+    ddf = read_dataset_as_ddf(factory=dataset_factory_function_scope)
+    # call optimize here and find the ddf length
+    ddf_before_repartition = dask.optimize(
```
fjetter, line 98: FYI: This optimize call is not sufficient to merge the repartition blocks. This is why we have the additional configuration in the `reparititon_ddf` function

kartothek/io/dask/repartition.py
usha-nemani-by: Test cases are fixed.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: Have written dedicated test case, with different repartition ratio values-  greater,  less than, and equal to 1.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: Have reverted the  changes in old testcases . Have added new testcase to test repartition ratio.

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns
+    --------
+    dask.DataFrame
+    """
+    return dask.optimize(ddf, optimizations=[dask.dataframe.optimize])[0]
+
+
+def calculate_repartition_ratio(
+    dataset_factory: DatasetFactory,
+    table_name: str,
```
fjetter, line 28: I suggest to drop this. We don't want to support multi tabled datasets anymore. Therefore, I suggest to not offer the possibility to select a table for any new function.

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,94 @@
...
+        dataset factory
+    table_name
+        Dataset Table name
+    predicates
+        Optional , but predicates to be applied during calculatin repartition ratio
+    target_nrows_per_partition
+        target num of rows per partition
+    dataset_metadata_frac
+        used during collection of dataset metadata
+    Returns
```
fjetter, line 50: The section requires a newline to separate it from the above section. I believe otherwise the docs are not rendered properly, are they?

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Repartition a dask DataFrame according, dividing the current number of partitions by `repartition_ratio`.
+
+    Parameters
+    ----------
+    ddf
+        dask dataframe
+    repartition_ratio
+        calculated repartition ratio
+    Returns
```
fjetter, line 86: Newline

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+    optimize,
+    repartition_ddf,
+)
+from kartothek.io.iter import store_dataframes_as_dataset__iter
+from kartothek.io_components.metapartition import SINGLE_TABLE
+
+
+@pytest.fixture
+def dataset_factory_function_scope(store_factory):
+    return DatasetFactory(dataset_uuid="test", store_factory=store_factory)
```
fjetter, line 17: I would prefer defining these in the test functions instead of using a fixture. The fixture itself is not functional since the factory only works for reading a dataset but the fixture doesn't set up the dataset. Therefore this factory is non-functional and the value of defining it in the fixture is marginal.

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@pytest.mark.parametrize("frac", (1.0, 0.5, 0.1))
```
fjetter, line 20: Use either 0.5 or 0.1. There is not much need for both since both are partial fractions. With the additional parametrization below we should keep combinatorics in check.

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+@pytest.mark.parametrize(
+    "df,target_nrows_per_partition",
+    (  # repartition by decreasing partition size, increasing number of partitions
+        (pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}), 100),
+        (pd.DataFrame({"A": range(2 ** 7), "part": ["A", "B", "C", "D"] * 2 ** 5}), 8),
```
fjetter, line 25: apart from using a large datafame, is this test case any different than the one above? Wouldn't it suffice to keep just one of it?

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+        # repartition by increasing partition size, decreasing number of partitions
+        (pd.DataFrame({"A": range(50), "part": range(50)}), 10),
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 10),
```
fjetter, line 28: Same here. is this any different to the case above?

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+        target_nrows_per_partition=target_nrows_per_partition,
+    ).compute()
+
+    ddf = read_dataset_as_ddf(factory=dataset_factory_function_scope)
+    ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+
+    assert (ddf.map_partitions(len).compute() == target_nrows_per_partition).all()
+
+
+@pytest.mark.parametrize("frac", (0.3, 0.1))
```
fjetter, line 60: I think we can hard code the frac for this test. This is not really impacting the logic we want to test, does it?

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+@pytest.mark.parametrize(
+    "df,target_nrows_per_partition",
+    (
+        # repartition by decreasing partition size, increasing number of partitions
+        (pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}), 97),
+        (pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}), 31),
+        # repartition by increasing partition size, decreasing number of partitions
+        (pd.DataFrame({"A": range(2 ** 6), "part": list("0123456789abcdef") * 4}), 11),
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 7),
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 1),
```
fjetter, line 70: These test cases also feel a bit redundant. Can we reduce this to a minimal set of cases we want to ensure work?

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,131 @@
...
+    "df,repartition_ratio",
+    (  # repartition by decreasing partition size, increasing number of partitions
+        (pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}), 3.88),
+        (
+            pd.DataFrame({"A": range(2 ** 7), "part": ["A", "B", "C", "D"] * 2 ** 5}),
+            0.25,
+        ),
+        # repartition by increasing partition size, decreasing number of partitions
+        (pd.DataFrame({"A": range(50), "part": range(50)}), 7),
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 1.0),
```
fjetter, line 112: I think we can use a fixed dataset for these tests and test a repartitioning of `<1, ==1, >1`. That would reduce test complexity

fjetter (changes requested):
An important place where this connects to the existing code base is https://github.com/JDASoftwareGroup/kartothek/blob/fc69b7c0f1f5893ad66989ab851acc5cf8d1f9f4/kartothek/io/dask/dataframe.py#L335
Can you call the new repartitioning there, please?

fjetter:
I'm still not entirely sold on this API. The repartition ratio itself was more or less a workaround I introduced back in the day to justify the repartitioning in `update_dataset_from_ddf`. Taking a step back, we'd be introducing two functions here which would always be called together

```python
ratio_graph = calculate_repartition_ratio(..., target_nrows_per_partition, frac)
ratio = ratio_graph.compute()
new_ddf = repartition_ddf(ddf, repartition_ratio=ratio)
```

I'm wondering if we should not instead introduce *one* function

```python
def repartition_ddf(ddf, partition_size: int):
    # We would block here
    npartitions = _calc_npartitions(...).compute()
    return ddf.repartition(npartitions=npartitions)

```
The implementation would be more or less the same but it feels to me as if the API was more user friendly. The only reason why two separate calls would be justified if we'd wanted to stick to the lazy execution.

kartothek/io/dask/repartition.py
usha-nemani-by: Done.

kartothek/io/dask/repartition.py
usha-nemani-by: Agreed. Updated the doc section accordingly.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: Agreed. Removed this fixture and its usage. Using store_factory wherever required. Please check and confirm.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: using 0.5 . Removed 0.1

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: removed 2 testcases from here. Have taken care that minimal combinations are used.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: done.

kartothek/io/dask/repartition.py
usha-nemani-by: Done. Please check and confirm.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: taken care of using minimal combinations.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: Done. Please check and confirm.

kartothek/io/dask/repartition.py
fjetter: > hm, why didn't we support this originally? This would be very useful for repartitioning after reading part of a dataset, perhaps we should pick it up again

Start small. Implement the easy feature first. Improvements and performance optimizations later. Feel free to implement the predicate pushdown properly for the stats collection but this was not included in the first iteration since it was not necessary to enable the feature. Always start with small increments...

kartothek/io/dask/repartition.py
```diff
@@ -0,0 +1,93 @@
...
+    )
+    # Once the kartothek issue is resolved, one should use the rowgroup wise
+    # metadata. This would set up future improvements once kartothek is able to
+    # filter RG predicates while collecting statistics (assuming this will ever
+    # be possible this way)
+    # See https://github.com/JDASoftwareGroup/kartothek/pull/352
+    # and https://issues.apache.org/jira/browse/ARROW-10146
+    expected_nrows_per_partition = dataset_metadata[
+        ["partition_label", "number_rows_total"]
+    ].drop_duplicates()
```
fjetter, line 67:
```suggestion
    expected_nrows_per_partition = dataset_metadata.groupby("partition_label")[
        "number_rows_per_row_group"
    ].sum()
```

fjetter: @usha-nemani-by let's do it future proof. While the groupby is technically more expensive, it shoundn't matter much. The drop dups is not cheap either. If we choose this implementation, future improvements in the statistics will automatically propagate to this function

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: Done. Please check and confirm.

usha-nemani-by:
> An important place where this connects to the existing code base is
> 
> https://github.com/JDASoftwareGroup/kartothek/blob/fc69b7c0f1f5893ad66989ab851acc5cf8d1f9f4/kartothek/io/dask/dataframe.py#L335
> 
> 
> Can you call the new repartitioning there, please?

I tried calling the new repartitioning, but it throws cyclic dependency error.
`
tests/io/dask/dataframe/test_repartition.py:5: in <module>
    from kartothek.io.dask.dataframe import read_dataset_as_ddf
kartothek/io/dask/dataframe.py:18: in <module>
    from kartothek.io.dask.repartition import  repartition_ddf
kartothek/io/dask/repartition.py:7: in <module>
    import kartothek.io.dask.dataframe as ktk_ddf
E   AttributeError: module 'kartothek.io.dask' has no attribute 'dataframe'
`

fjetter:
> I tried calling the new repartitioning, but it throws cyclic dependency error.

Put the stuff of repartition in the dataframe submodule. I don't think there is a dedicated reason to separate these two

usha-nemani-by:
> An important place where this connects to the existing code base is
> 
> https://github.com/JDASoftwareGroup/kartothek/blob/fc69b7c0f1f5893ad66989ab851acc5cf8d1f9f4/kartothek/io/dask/dataframe.py#L335
> 
> 
> Can you call the new repartitioning there, please?

Done. called `repartition_ddf()` using the input `repartition_ratio`.
Is it correct? or do I need to calculate the ratio as well ?

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +604,83 @@ def hash_dataset(
...
+        calculated repartition_ratio
+    """
+    dataset_metadata = collect_dataset_metadata(
+        factory=dataset_factory, predicates=predicates, frac=dataset_metadata_frac,
+    )
+    expected_nrows_per_partition = dataset_metadata.groupby("partition_label")[
+        "number_rows_per_row_group"
+    ].sum()
+
+    mean_nrows_per_partition = expected_nrows_per_partition.mean()
```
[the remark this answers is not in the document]
lr4d: this code hasn't been changed, has it. Why do you think it may be incorrect?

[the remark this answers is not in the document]
usha-nemani-by: It is changed now.  We are using `number_rows_per_row_group `  and directly taking the mean .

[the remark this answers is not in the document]
usha-nemani-by: As discussed with @lr4d , this changes looks fine.

lr4d (approved):

fjetter:
So we didn't go with my proposal? Any reason?

usha-nemani-by:
We are still using `repartition_ddf()` and `calculate_repartition_ratio()` functions independently. 
Once `repartition_ratio` gets deprecated, we can merge the above two functions and make call `repartition_ddf()`which internally calculates the ratio.
This was our conclusion.
CC: @lr4d

fjetter:
1. I'd appreciate the transparency of posting these conclusions before merging anything
2. We didn't discuss any kind of deprecation, yet. 
3. What has the deprecation of an old feature to do with the introduction of a new function?

lr4d:
>We didn't discuss any kind of deprecation, yet.
>What has the deprecation of an old feature to do with the introduction of a new function?

This is indeed another topic. We can have this discussion in a separate issue.

> We are still using repartition_ddf() and calculate_repartition_ratio() functions independently.

This is the rationale. Specifically, in `update_dataset_from_ddf` we are using `repartition_ddf` based on user-provided `repartition_ratio`. To not introduce any breaking changes, the reasoning was to not automatically calculate the repartition ratio in that function for now, and instead rely on the logic of `repartition_ddf`.

lr4d:
> I'd appreciate the transparency of posting these conclusions before merging anything

My bad. We had an offline discussion and forgot to include our conclusions here

lr4d:
@usha-nemani-by Can you open another PR so we can settle the discussion?

PR #358, opened 2020-10-09
Improve filter performance for long object lists for 'in' literals

fjetter:
# Description:

For long lists of strings in the `in` clause this should help speed up the filtering significantly. I've seen this dominating query planing and IO. While this is not necessarily an encouraged practice, I've seen users doing this rather often.

With this change we can see several orders of magnitude better performance for very large lists (>>100). In particular, the new implementation essentially scales with the size of the array but is approximately constant in the length of the value. This is probably only true as long as `len(value) << len(array)` but this should be true for most applications.

For native types the performance of native numpy is not beatable and therefore I build in the switch based on types. While the new implementation seems to be slightly slower for small samples even for string types I consider this overhead to be a price worth paying to save on complexity. Hitting this threshold properly would otherwise require a lot of measurements. The optimal cutoff is probably a function of both array and value length. This might also be dependent on the kind of strings of the input dataset since the comparison of strings is also not constant time. Therefore I'm fine for this being slower.

W.r.t. the below measurements, I was meassuring them by enable/disabling the new code path explicitly. Therefore the code I was benchmarking used an `if enabled` instead of the `if value.dtype.kind in ("U", "S", "O")` but this should not impact the benchmark.

cc @NeroCorleone you have been looking into this as well

| | | array_size / enabled | | | | | |
| --------- | ------------- | ------------- | --------------- | --------------- | ---------------- | ---------------- | ----------------- |
|   column  |  filter_size  |  10000 / True |  10000 / False  |  100000 / True  |  100000 / False  |  1000000 / True  |  1000000 / False
|   int64   |       10      |  3.82±0.06ms  |     216±10μs    |    15.8±0.3ms   |     660±20μs     |     246±10ms     |     7.26±0.1ms |
|   int64   |      100      |   3.65±0.2ms  |     907±30μs    |    15.8±0.6ms   |   10.4±0.04ms    |     237±7ms      |      154±1ms |
|   int64   |      1000     |   3.90±0.2ms  |     978±60μs    |    16.3±0.4ms   |    11.5±0.2ms    |     244±6ms      |      168±1ms |
|  unicode  |       10      |   5.48±0.2ms  |    2.32±0.1ms   |    51.9±0.9ms   |     41.6±1ms     |     980±10ms     |     1.10±0.02s |
|  unicode  |      100      |   5.38±0.2ms  |     21.2±1ms    |     52.5±1ms    |     475±30ms     |     987±30ms     |     10.2±0.2s |
|  unicode  |      1000     |   5.97±0.3ms  |     208±6ms     |     54.6±2ms    |    4.05±0.07s    |     973±9ms      |       timeout |


kartothek/serialization/_generic.py
```diff
@@ -486,14 +486,26 @@ def filter_array_like(
             np.logical_and(array_like > value, mask, out=out)
         elif op == "in":
             value = np.asarray(value)
+            nullmask = pd.isnull(value)
+            if value.dtype.kind in ("U", "S", "O"):
+                unique_vals = np.unique(value[~nullmask])
+                value_ser = pd.Series(unique_vals, name="value")
+                arr_ser = pd.Series(array_like, name="array").to_frame()
+                matching_idx = (
```
lr4d, line 494: Please add a comment here to explain what you're doing

fjetter: what do you have in mind? "pandas join is faster than numpy.isin"?

lr4d: More like: left join user-input with row-group arrays, do something with nulls...

lr4d: And wrt the rationale, I think it'd be great if you can summarize this PR description and add it as a comment, so that anybody looking at the code can understand why for these dtypes we use different logic

lr4d:
Interesting benchmarks, @fjetter . Thanks for the nice description.
Did you look into potentially using categoricals for this?

kartothek/serialization/_generic.py
fjetter: > More like: left join user-input with row-group arrays, do something with nulls...

My problem with stuff like this is that it essentially spells out what the code says. Array left join values is literally what's written there

lr4d (approved):

fjetter:
> Did you look into potentially using categoricals for this?

No, not yet. Performance should only be marginally better for cases where the size of the to-be-filtered array is very large. The only way how we'd safe time/memory here is that the join would probably be a bit cheaper. otherwise I'd expect categoricals to be slower since they require one additional pass over the data and require more complicated internals. Overall I figured the complexity wouldn't be worth it

PR #360, opened 2020-10-14
Fix get_indices_as_dataframe for empty columns and indices

mlondschien:
On master, the following code block raises a `IndexError: pop from empty list`:

```python
import pandas as pd
import storefact
from functools import partial
from kartothek.io.eager import read_table, store_dataframes_as_dataset

path = "/tmp"
uuid = "dataset"
df = pd.DataFrame({"x": [1, 2], "y": ["a", "b"]})
store = partial(storefact.get_store_from_url, f"hfs://{path}?create_if_missing=False")
    dfs=[df],
    dataset_uuid=uuid,
    store=store,
    partition_on=["x"],
    overwrite=True,
)
dm.get_indices_as_dataframes()
```

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-2-75c96ab15718> in <module>
     15     overwrite=True,
     16 )
---> 17 dm.get_indices_as_dataframe()

~/code/kartothek/kartothek/core/dataset.py in get_indices_as_dataframe(self, columns, date_as_object, predicates)
    409         # start joining with the small ones
    410         sorted_dfs = sorted(dfs, key=lambda df: len(df))
--> 411         result = sorted_dfs.pop(0)
    412         for df in sorted_dfs:
    413             result = result.merge(df, left_index=True, right_index=True, copy=False)

IndexError: pop from empty list
```

To help my understanding: Why is here `dm.indices` not equal to `["x"]`? I guess `tests/io_components/test_dataset.py` would be the appropriate place to add a test. How can I replicate this without writing a dataset (as is done in the other tests)?

- [ ] Closes #xxxx
- [ ] Changelog entry


fjetter:
Thanks for the report. Really glad to see you chime in here!

> Why is here dm.indices not equal to ["x"]?

Indices are not "loaded" by default. For partition indices we need to parse the indices first and secondary indices need to be loaded from store. Therefore, the initialised `DatasetMetadata` does not hold these datastructures in memory unless they are required. Usually we opt for an approach where we load them automatically once required or, if that's not possible, indicate this with an explicit error message. This is definitely wrong.

A suitable fix would be to load the partition indices if they are required, something like

```python
# Pseudo!
def get_indices_as_dataframe(self,..., columns):
    if set(columns) & set(self.partition_keys):
        self.load_partition_indices()
    # Do the stuff we're here for...

```

> How can I replicate this without writing a dataset

That's tricky. In terms of testability the architecture of kartothek is not very inviting I am afraid. Putting enough internal knowledge into it, creates the following reproducing example

```python
import pandas as pd
from kartothek.core.common_metadata import make_meta
from kartothek.core.dataset import DatasetMetadataBuilder
from kartothek.core.partition import Partition

builder = DatasetMetadataBuilder(
    uuid='test_uuid',
    partition_keys=["partition"],
    table_meta={
        "table": make_meta(pd.DataFrame({"partition": [1], "payload": ["A"]}), origin='test')
    },
)
part = Partition(
    label="partition=1/12345",
    files={"table": "test_uuid/table/partition=1/12345.parquet"},
)
builder.add_partition(part.label, part)
builder.to_dataset().get_indices_as_dataframe()
```
This is a little bit fickle because once we change anything about the storage format this breaks immediately. Writing something a layer above this is usually a bit more robust which is why most tests create an actual dataset.

Note, that this uses a few classes I want to get rid of eventually. The `Partition` class outlived its usefulness years ago and this is literally the only way it is internally used. The MetadatBuilder also feels redundant. For now, this is, however, the only possibility to create this without writing actual files.

That all being said, I'd be fine either way.

Once we redesign these classes we should keep the testability in mind. Most of our tests could be implemented w/out writing any file but we're unfortunately not there, yet.

mlondschien:
Sure. We're big fans of the repo, thanks for open-sourcing maintaining it!

I can add the suggested code block. However, I think that we still need to take care about the case where `columns=None` and `dm.indices={}`. I would add a test that writes a temporary dataset. Is there ever a case where columns which are not indices would be passed to `columns`?

kartothek/core/dataset.py
```diff
@@ -408,6 +415,7 @@ def get_indices_as_dataframe(
 
         # start joining with the small ones
         sorted_dfs = sorted(dfs, key=lambda df: len(df))
+        # TODO: Why not do a pd.concat here?
```
fjetter, line 418: probably faster and easier to read, you are right. Not sure why we haven't used this, probably because the one who implemented it (which I believe was me) didn't think of it at the time.

Unfortunately we don't have any ready-to-use benchmarks for this function yet. That would be a nice thing to benchmark. If you feel like it, you can add a case to `asv_bench/benchmarks` and give it a shot.

cc @NeroCorleone I believe this is the place where the memory spike you're investigating occurs. You might want to check out the concat instead of the merge as well for the high mem scenarios

I guess, this would be a good change for a followup PR, though, to not mix up too much stuff in one go.

kartothek/core/dataset.py
```diff
@@ -376,29 +376,36 @@ def get_indices_as_dataframe(
         Parameters
         ----------
         """
+        if not self.primary_indices_loaded:
+            # self.load_partition_indices is not inplace
```
fjetter, line 380: I guess this is why we implemented this differently so far. That's fine for me, though.

kartothek/core/dataset.py
```diff
@@ -376,29 +376,36 @@ def get_indices_as_dataframe(
...
-        elif columns == []:
-            return pd.DataFrame(index=self.partitions)
+            columns = sorted(dm.indices.keys())
+
+        if columns == []:
+            return pd.DataFrame(index=dm.partitions)
+
+        if set(columns).difference(dm.indices):
+            # TODO: What to do here? Why not just subset?
+            raise ValueError("Columns not in partition_keys.")
```
fjetter, line 393: I think this is a matter of API design. I consider it a bit more user friendly to raise if the input is unexpected. If I compare this to pandas API, I usually receive a KeyError if I'm trying to perform an operation on something which isn't there. Why should we choose a subset here?

fjetter: Should probably be something like

`Requested indices for columns '{set(columns).difference(dm.indices)}' which were not indexed.`

since this does not only affect partition columns

tests/io_components/test_read.py
```diff
@@ -57,10 +57,10 @@ def test_dispatch_metapartitions_without_dataset_metadata(dataset, store_session
     assert len(partitions) == 2
 
     mp = partitions[0]
-    assert mp.dataset_metadata == {}
+    assert list(mp.dataset_metadata.keys()) == ["creation_time"]
```
fjetter, line 60: weird, this is a regression. We changed the dispatch to no longer include the metadata with the MPs since this redundant data duplication was causing issues. I can't understand which change affected this behaviour.

fjetter (changes requested):
Apart from a few minor details (see comments) this looks good

fjetter:
Can you rebase, please? There is not only a merge conflict but something seems weird with the build jobs. maybe the configuration on your branch is outdated

mlondschien:
> Can you rebase, please? There is not only a merge conflict but something seems weird with the build jobs. maybe the configuration on your branch is outdated

My mistake. I forgot to pull from upstream before committing. It appears the `get_indices_as_dataframes` method has substantially changed in between. I'll have a look and merge the changes later.

mlondschien:
@fjetter could you give me a hint how to fix the tests?
```
=========================== short test summary info ============================
FAILED tests/io/cube/test_query.py::test_delayed_index_build_partition_by[dask_bag_bs1]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_partition_by[dask_bag_bs3]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_partition_by[dask_dataframe]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_partition_by[eager]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_correction_restriction[dask_bag_bs1]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_correction_restriction[dask_bag_bs3]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_correction_restriction[dask_dataframe]
FAILED tests/io/cube/test_query.py::test_delayed_index_build_correction_restriction[eager]
FAILED tests/io/dask/dataframe/test_stats.py::test_collect_dataset_metadata_empty_dataset
```

lr4d:
@mlondschien Can you post a traceback of the test failure?

lr4d:
Nvm. Got them here

<details>
  <summary>test_delayed_index_build_correction_restriction[dask_bag_bs3] </summary>
  
  ```
________ _________

driver = <function wrap_bag_read.<locals>._driver at 0x7f1c43403a60>
function_store = functools.partial(<function _get_store at 0x7f1c93c6ad30>, '/tmp/pytest-of-runner/pytest-0/test_delayed_index_build_corre1/function_store')

    def test_delayed_index_build_correction_restriction(driver, function_store):
        """
        Ensure that adding extra indices for dimension columns does not mark other datasets as restrictive.
        """
        df_seed = pd.DataFrame({"x": [0, 1, 2, 3, 4, 5], "p": [0, 0, 1, 1, 2, 2]})
        df_extend = pd.DataFrame({"x": [0, 1, 2], "p": [0, 0, 1], "v": [0, 1, 2]})
        cube = Cube(
            dimension_columns=["x"],
            partition_columns=["p"],
            uuid_prefix="delayed_index_cube",
            index_columns=[],
        )
        build_cube(
            data={"seed": df_seed, "extend": df_extend}, store=function_store, cube=cube
        )
    
        build_dataset_indices(
            store=function_store,
            dataset_uuid=cube.ktk_dataset_uuid("extend"),
            columns=["x"],
        )
    
>       results = driver(cube=cube, store=function_store, conditions=C("x") >= 0)

kartothek/io/testing/query_cube.py:1142: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/io/cube/utils.py:50: in _driver
    b = f(blocksize=blocksize, **kwargs)
kartothek/io/dask/bag_cube.py:180: in query_cube_bag
    _empty, b = query_cube_bag_internal(
kartothek/io/dask/common_cube.py:262: in query_cube_bag_internal
    intention, empty, groups = plan_query(
kartothek/io_components/cube/query/__init__.py:299: in plan_query
    datasets = discover_datasets(
kartothek/api/discover.py:186: in discover_datasets
    check_datasets(result, cube)
kartothek/api/consistency.py:324: in check_datasets
    _check_indices(datasets, cube)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

datasets = {'extend': DatasetMetadata(uuid=delayed_index_cube++extend, tables=['table'], partition_keys=['p'], metadata_version=4..._cube++seed, tables=['table'], partition_keys=['p'], metadata_version=4, indices=['x', 'p'], explicit_partitions=True)}
cube = Cube(dimension_columns=('x',), partition_columns=('p',), uuid_prefix='delayed_index_cube', seed_dataset='seed', index_columns=frozenset())

    def _check_indices(datasets: Dict[str, DatasetMetadata], cube: Cube) -> None:
        """
        Check if required indices are present in given datasets.
    
        For all datasets the primary indices must be equal to ``ds.partition_keys``. For the seed dataset secondary
        indices for all dimension columns are expected.
    
        Parameters
        ----------
        datasets
            Datasets.
        cube
            Cube specification.
    
        Raises
        ------
        ValueError: In case indices are broken.
        """
        for ktk_cube_dataset_id in sorted(datasets.keys()):
            ds = datasets[ktk_cube_dataset_id]
            primary_indices = ds.partition_keys
            columns = get_dataset_columns(ds)
            secondary_indices = set()
            any_indices = set(cube.index_columns) & columns
    
            if ktk_cube_dataset_id == cube.seed_dataset:
                secondary_indices |= set(cube.dimension_columns)
    
            for types_untyped, elements in (
                ((PartitionIndex,), primary_indices),
                ((ExplicitSecondaryIndex,), secondary_indices),
                ((ExplicitSecondaryIndex, PartitionIndex), any_indices),
            ):
                types = cast(Tuple[type, ...], types_untyped)
    
                tname = " or ".join(t.__name__ for t in types)
    
                # it seems that partition indices are not always present (e.g. for empty datasets), so add partition keys to
                # the set
                indices = cast(Dict[str, Union[IndexBase, str]], copy(ds.indices))
                if PartitionIndex in types:
                    for pk in ds.partition_keys:
                        if pk not in indices:
                            indices[pk] = "dummy"
    
                for e in sorted(elements):
                    if e not in indices:
                        raise ValueError(
                            '{tname} "{e}" is missing in dataset "{ktk_cube_dataset_id}".'.format(
                                tname=tname, e=e, ktk_cube_dataset_id=ktk_cube_dataset_id
                            )
                        )
    
                    idx = indices[e]
                    t2 = type(idx)
                    tname2 = t2.__name__
                    if (idx != "dummy") and (not isinstance(idx, types)):
>                       raise ValueError(
                            '"{e}" in dataset "{ktk_cube_dataset_id}" is of type {tname2} but should be {tname}.'.format(
                                tname=tname,
                                tname2=tname2,
                                e=e,
                                ktk_cube_dataset_id=ktk_cube_dataset_id,
                            )
                        )
E                       ValueError: "p" in dataset "extend" is of type ExplicitSecondaryIndex but should be PartitionIndex.

kartothek/api/consistency.py:253: ValueError

```
  
</details>

<details>
  <summary> test_collect_dataset_metadata_empty_dataset</summary>
  

  ```
_________________ __________________

store_factory = functools.partial(<function get_store_from_url at 0x7f1c9b3a5af0>, 'hfs:///tmp/pytest-of-runner/pytest-0/test_collect_dataset_metadata_3/store')

    def test_collect_dataset_metadata_empty_dataset(store_factory):
        df = pd.DataFrame(columns=["A", "b"], index=pd.RangeIndex(start=0, stop=0))
        store_dataframes_as_dataset(
            store=store_factory, dataset_uuid="dataset_uuid", dfs=[df], partition_on=["A"]
        )
>       df_stats = collect_dataset_metadata(
            store=store_factory, dataset_uuid="dataset_uuid", table_name="table",
        ).compute()

tests/io/dask/dataframe/test_stats.py:213: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<decorator-gen-23>:2: in collect_dataset_metadata
    ???
kartothek/io_components/utils.py:286: in normalize_args
    return _wrapper(*args, **kwargs)
kartothek/io_components/utils.py:284: in _wrapper
    return function(*args, **kwargs)
kartothek/io/dask/dataframe.py:511: in collect_dataset_metadata
    mps = list(
kartothek/io_components/read.py:98: in dispatch_metapartitions_from_factory
    base_df = dataset_factory.get_indices_as_dataframe(
kartothek/core/dataset.py:383: in get_indices_as_dataframe
    dm = self.load_partition_indices()
kartothek/core/factory.py:158: in load_partition_indices
    self._cache_metadata = self.dataset_metadata.load_partition_indices()
kartothek/core/dataset.py:348: in load_partition_indices
    indices = _construct_dynamic_index_from_partitions(
kartothek/core/dataset.py:725: in _construct_dynamic_index_from_partitions
    return _empty_partition_indices(partition_keys, table_meta, default_dtype)
kartothek/core/dataset.py:714: in _empty_partition_indices
    indices[col] = PartitionIndex(column=col, index_dct={}, dtype=arrow_type)
kartothek/core/index.py:567: in __init__
    super(PartitionIndex, self).__init__(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = PartitionIndex(), column = 'A', index_dct = {}, dtype = DataType(null)
normalize_dtype = True

    def __init__(
        self,
        column: str,
        index_dct: Optional[IndexDictType] = None,
        dtype: Optional[pa.DataType] = None,
        normalize_dtype: bool = True,
    ):
        """
        Initialize an IndexBase.
    
        Parameters
        ----------
        column:
            Name of the column this index is for.
        index_dct:
            Mapping from index values to partition labels
        dtype:
            Type of index. If left out and ``index_dct`` is present, this will be inferred.
        normalize_dtype:
            Normalize type information and values within ``index_dct``. The user may
            disable this when it the index was already normalized, e.g. when the
            index python objects gets copied, or when the index data is restored
            from a parquet file that was written by a trusted write path.
        """
        if column == _PARTITION_COLUMN_NAME:
            raise ValueError(
                "Cannot create an index for column {} due to an internal implementation conflict. "
                "Please contact a kartothek maintainer if you receive this error message".format(
                    column
                )
            )
        if (dtype is None) and index_dct:
            # do dtype given but index_dct is present => auto-derive dtype
            table = _index_dct_to_table(index_dct, column, None)
            schema = table.schema
            dtype = schema[0].type
    
        if dtype is not None:
            if pa.types.is_nested(dtype):
                raise NotImplementedError("Indices w/ nested types are not supported")
            if pa.types.is_null(dtype):
>               raise NotImplementedError("Indices w/ null/NA type are not supported")
E               NotImplementedError: Indices w/ null/NA type are not supported

kartothek/core/index.py:79: NotImplementedError
  ```
  
</details>

File: kartothek/core/dataset.py [hunk 5]
```diff
@@ -378,20 +378,26 @@ def get_indices_as_dataframe(
...
+            dm = self
+
         if columns is None:
-            columns = sorted(self.indices.keys())
-        elif columns == []:
-            return pd.DataFrame(index=self.partitions)
+            columns = sorted(dm.indices.keys())
+
+        if columns == []:
+            return pd.DataFrame(index=dm.partitions)
```
lr4d, line 391: Can you try putting these `if` blocks above the `         if not self.primary_indices_loaded:` block?

mlondschien: I don't think I understand. The calls `dm.indices.keys()` and `dm.partitions` would not work if indices are not loaded. This is why I added the `dm = self.load_partition_indices` in the first place.

lr4d: hm, right. The issue with the `test_collect_dataset_metadata_empty_dataset` failure is related to calling this function with `columns=[]`. If you place another
```
 if columns == []:
            return pd.DataFrame(index=dm.partitions)
```
at the beginning, this will pass, but it's not the most elegant solution

lr4d:
@marco-neumann-by can you maybe chime in on the failures related to the cube tests? See the traceback I posted for `test_delayed_index_build_correction_restriction[dask_bag_bs3] ` 2 posts above

kartothek/core/dataset.py
mlondschien: I added a commit on top to only load indices if `columns!=[]`.

tests/io_components/test_dataset.py
```diff
@@ -117,3 +109,22 @@ def test_dataset_get_indices_as_dataframe_predicates():
         columns=["l_external_code"], index=pd.Index([], name="partition")
     )
     pdt.assert_frame_equal(result, expected)
+
+
+def test_dataset_get_indices_as_dataframe_no_index(dataset):
+    assert not dataset.primary_indices_loaded
+    df = dataset.get_indices_as_dataframe()
+    assert not dataset.primary_indices_loaded
```
lr4d, line 117: I don't know if we should have this assert in here (and the test below). Imo, ideally we would load the indices in-place but that's another discussion
```suggestion
```

tests/io_components/test_dataset.py
```diff
@@ -117,3 +109,22 @@ def test_dataset_get_indices_as_dataframe_predicates():
...
+    pdt.assert_frame_equal(df, pd.DataFrame(index=["cluster_1", "cluster_2"]))
+
+
+def test_dataset_get_indices_as_dataframe_with_index(dataset_with_index, store_session):
+    assert not dataset_with_index.primary_indices_loaded
+    df = dataset_with_index.get_indices_as_dataframe()
+    assert not dataset_with_index.primary_indices_loaded
```
lr4d, line 124:
```suggestion
```

lr4d (commented):
Looks good @mlondschien . Thanks for the effort. Some nitpicks. Can you add a changelog entry also?

lr4d (approved):

fjetter (approved):
Looks great, thanks @mlondschien

PR #365, opened 2020-10-22
Fix cube remove_partitions with conditions that match no partition

### LINKED ISSUE #364: remove_partitions removes all partition in case conditions do not match any partition
### Problem description

The remove_partitions function for cubes removes all partitions of a dataset in case a condition is provided that matches no partition at all.

jochen-ott-by:
- [ ] Closes #364
- [x] Changelog entry

### DIFF
--- a/kartothek/io_components/cube/remove.py
+++ b/kartothek/io_components/cube/remove.py
@@ -61,30 +61,26 @@ def prepare_metapartitions_for_removal_action(
     for ktk_cube_dataset_id in ktk_cube_dataset_ids:
         ds = existing_datasets[ktk_cube_dataset_id]
         ds = ds.load_partition_indices()
-
-        df_partitions = get_partition_dataframe(dataset=ds, cube=cube)
-        df_partitions = df_partitions.drop_duplicates()
-        local_condition = reduce(
-            lambda a, b: a & b,
-            (
-                cond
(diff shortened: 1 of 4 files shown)


kartothek/io_components/cube/remove.py
```diff
@@ -61,30 +61,26 @@ def prepare_metapartitions_for_removal_action(
...
-
         mp = _prepare_mp_empty(ds)
 
-        if df_partitions.empty:
-            # might be empty due to:
-            #  - empty dataset
-            #  - empty partition_keys
-            #
-            # in both cases, use a "delete all"-scope
+        if not ds.partition_keys:
```
fjetter, line 66: Not sure if this is a safe default. If there are no partition keys, this would delete the entire dataset, no matter what the input is, wouldn't it?

jochen-ott-by: Correct. I believe this is sane behaviour. But as it may be unexpected, I added the according documentation for it at the `remove_partitions` function, please see there whether it makes sense. If not, we have to modify current behaviour in a non-trivial way, which I'd rather do separately to this fix.

fjetter: Yes, just wanted to understand what's going on

fjetter (approved):
Thanks @jochen-ott-by

PR #367, opened 2020-10-27
Addition of  partition_ddf to Ktk

usha-nemani-by:
This PR is continuation to PR#357 , to add more changes based on the discussion.

Addition of  partition_ddf func to Ktk

- [ ] Addition of  partition_ddf  func to Ktk
- [ ] New test cases addition. 
- [ ] Changelog entry

### DIFF
--- a/kartothek/io/dask/dataframe.py
+++ b/kartothek/io/dask/dataframe.py
@@ -1,3 +1,4 @@
+import math
 import random
 import warnings
 from typing import (
@@ -606,3 +607,82 @@ def hash_dataset(
             .apply(_unpack_hash, unpack_meta=ddf._meta, subset=subset, meta="uint64")
             .astype("uint64")
         )
+
+
+def _calc_target_npartitions(
+    ddf: dd.DataFrame, dataset_metadata: dd.DataFrame, target_nrows_per_partition: int
+) -> float:
+    """
+        This is an internal function to calculate target_npartitions using dataset metadata information and target_nrows_per_partition.
+
+        Parameters
+        ----------
+        ddf
+            dask.DataFrame
+        dataset_metadata
+            dask.DataFrame . This is the dataset metadata information of the input ddf
+        target_nrows_per_partition
+            Target num of rows per partition provided by customer
+
+
+        Returns
+        -------
+        target_npartitions
+            int
+        """
+    expected_nrows_per_partition = dataset_metadata.groupby("partition_label")[
+        "number_rows_per_row_group"
+    ].sum()
+
+    mean_nrows_per_partition = expected_nrows_per_partition.mean()
+    repartition_ratio = target_nrows_per_partition / mean_nrows_per_partition
+    repartition_ratio = repartition_ratio.compute()
+    with dask.config.set({"optimization.fuse.ave-width": math.ceil(repartition_ratio)}):
+        # Minimum amount of partitions is 1.
+        target_npartitions = max(1, int(math.ceil(ddf.npartitions / repartition_ratio)))
+        return target_npartitions
+
+
+def repartition_ddf(
+    ddf: dd.DataFrame,
+    dataset_factory: DatasetFactory,
+    target_nrows_per_partition: int,
+    dataset_metadata_frac: float,
+    predicates: Optional[PredicatesType] = None,
+) -> dd.DataFrame:
+    """
+    This is an IO  function used to repartition the input ddf using the target_nrows_per_partition input param
+    The metadata of the ddf is collected using Ktk Functions and expected_nrows_per_partition is extracted`.
+    An intermediate parameter repartition_ratio (per_partition) which is target_rows/expected_rows is computed.
+    Final target number of partitions (target_npartitions) are calculated  on the ddf using the above ratio.
+    The input ddf is repartitioned using target_npartitions value.
+
+    Parameters
+    ----------
+    ddf
+        dask.DataFrame
+    dataset_factory
+        dataset factory
+    predicates
+        Optional , but predicates can be applied while collecting_dataset_metadata.
+        Note that predicates will only be applied during query planning, predicate pushdown is not (yet) supported.
+        Thus, this is only useful if providing predicates involving indexed columns.
+    target_nrows_per_partition
+        target num of rows per partition
+    dataset_metadata_frac
+        This fraction is helps in collecting metadata on a fraction(sample) of the ddf, instead of the whole ddf.
+
+
+    Returns
+    -------
+    dask.DataFrame
+    """
+
+    dataset_metadata = collect_dataset_metadata(
+        factory=dataset_factory, predicates=predicates, frac=dataset_metadata_frac,
+    )
+    target_npartitions = _calc_target_npartitions(
+        ddf, dataset_metadata, target_nrows_per_partition
+    )
+    ddf2 = ddf.repartition(npartitions=target_npartitions)
+    return dask.optimize(ddf2, optimizations=[dask.dataframe.optimize])[0]
--- a/tests/io/dask/dataframe/test_repartition.py
+++ b/tests/io/dask/dataframe/test_repartition.py
@@ -1,0 +1,82 @@
+import pandas as pd
+import pytest
+
+from kartothek.core.factory import DatasetFactory  # noqa: E402
+from kartothek.io.dask.dataframe import read_dataset_as_ddf, repartition_ddf
+from kartothek.io.iter import store_dataframes_as_dataset__iter
+
+
+@pytest.mark.parametrize("frac", (1.0, 0.5))
+@pytest.mark.parametrize(
+    "df,target_nrows_per_partition",
+    (
+        (
+            pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}),
+            100,
+        ),  # rep_ratio = 4
+        (
+            pd.DataFrame({"A": range(2 ** 7), "part": ["A", "B", "C", "D"] * 2 ** 5}),
+            8,
+        ),  # rep_ratio = 0.25
+    ),
+)
+def test_repartition_ddf_exact_match(
+    df, target_nrows_per_partition, frac, store_factory
+):
+    """
+    Because we carefully choose the number of original partitions and the value of
+    `target_nrows_per_partition`, we get exact matches for `target_nrows_per_partition`.
+    This is not expected in a real-world scenario where the conditions are very unlikely.
+    """
+    dfs = [df]
+    store_dataframes_as_dataset__iter(
+        dfs, store=store_factory, dataset_uuid="test", partition_on=["part"],
+    )
+
+    ddf = read_dataset_as_ddf(store=store_factory, dataset_uuid="test")
+    ddf = repartition_ddf(
+        ddf,
+        dataset_factory=DatasetFactory(
+            dataset_uuid="test", store_factory=store_factory
+        ),
+        target_nrows_per_partition=target_nrows_per_partition,
+        dataset_metadata_frac=frac,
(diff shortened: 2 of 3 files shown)


usha-nemani-by:
@fjetter Comment :

I'm still not entirely sold on this API. The repartition ratio itself was more or less a workaround I introduced back in the day to justify the repartitioning in update_dataset_from_ddf. Taking a step back, we'd be introducing two functions here which would always be called together

```
ratio_graph = calculate_repartition_ratio(..., target_nrows_per_partition, frac)
ratio = ratio_graph.compute()
new_ddf = repartition_ddf(ddf, repartition_ratio=ratio)

```
I'm wondering if we should not instead introduce one function

```
def repartition_ddf(ddf, partition_size: int):
    # We would block here
    npartitions = _calc_npartitions(...).compute()
    return ddf.repartition(npartitions=npartitions)
```

The implementation would be more or less the same but it feels to me as if the API was more user friendly. The only reason why two separate calls would be justified if we'd wanted to stick to the lazy execution.

lr4d:
I had already approved the (previous) PR. @fjetter Can you share your thoughts?

fjetter:
My thoughts are effectively written above. the repartition ratio is not a good api I believe, since it's rather obscure without delivering actual value to the user in terms of functionality.
I introduced the parameter back in the day on application level and figured that repartitioning is essential for shuffling but I didn't have a proper tool to describe it better and just moved the application level config to the library. However, the repartition ratio has little value compared to the plain `npartitions` of task. The value we want to offer in this PR is that we may calculate rapidly how many partitions are requested by the user. In effect we're offering a faster alternative to the existing api of dask but dask doesn't have a repartition ratio. 
Therefore, I argue that that we should rather use quantities already familiar. 
After all, the task/story we are working on at the moment is about this exact problem. Repartiton ratio is a fragile abstraction. Number of rows (or size in MB) is much more tangible and robust

lr4d:
>  Therefore, I argue that that we should rather use quantities already familiar.

I agree

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +604,67 @@ def hash_dataset(
...
+    predicates: Optional[PredicatesType] = None,
+    *,
+    target_nrows_per_partition: int,
+    dataset_metadata_frac: float,
+) -> dd.core.Scalar:
+    """
+    Calculate a `repartition_ratio` according to the dataset input and `target_nrows_per_partition` (the desired
+    number of rows of each dataframe partition).
+
+
```
marco-neumann-by, line 620: 1. please add a note that this function performs IO.
2. "calculate" could be read by many users as "compute on client". I would prefer something like `compute`.

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +604,67 @@ def hash_dataset(
...
+) -> dd.DataFrame:
+    """
+    Repartition a dask DataFrame according, dividing the current number of partitions by `repartition_ratio`.
+
+    Parameters
+    ----------
+    ddf
+        dask.Dataframe
+    repartition_ratio
+        calculated repartition ratio
```
marco-neumann-by, line 660: Please linke `calculate_repartition_ratio`

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +604,67 @@ def hash_dataset(
...
+    mean_nrows_per_partition = expected_nrows_per_partition.mean()
+    repartition_ratio = target_nrows_per_partition / mean_nrows_per_partition
+    return repartition_ratio
+
+
+def repartition_ddf(
+    ddf: dd.DataFrame, repartition_ratio: SupportsFloat
+) -> dd.DataFrame:
+    """
+    Repartition a dask DataFrame according, dividing the current number of partitions by `repartition_ratio`.
```
marco-neumann-by, line 653: What exactly is a "repartition ratio"? Neither this nor the other function explain this term.

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,106 @@
...
+    # this is somewhat arbitrary as original partition size may be non-constant
+    margin_of_error = 0.5  # proportion
+    partition_lengths = ddf.map_partitions(len).compute()
+    assert (
+        ((1 - margin_of_error) * target_nrows_per_partition <= partition_lengths)
+        & (partition_lengths <= (1 + margin_of_error) * target_nrows_per_partition)
+    ).all()
+
+
+@pytest.mark.parametrize("repartition_ratio", (3.88, 0.25, 7, 1.0))
```
marco-neumann-by, line 92: where do these numbers come from?

marco-neumann-by (changes requested):

kartothek/io/dask/dataframe.py
marco-neumann-by: I should have read the PR comments before, sorry. I see you already more or less agreed that this term is somewhat fragile / non-user-friendly.

usha-nemani-by:
>Therefore, I argue that that we should rather use quantities already familiar.
> After all, the task/story we are working on at the moment is about this exact problem. Repartiton ratio is a fragile 
abstraction. Number of rows (or size in MB) is much more tangible and robust


Based on above discussion, I modified the functionality.
I have removed the `calculate_repartition_ratio()` function.
Modified the `repartition_ddf()` functionality  as `compute_repartition_ddf()`to compute  the repartition_ratio and compute `target_npartitions` and finally repartition the ddf.
I felt that `compute_repartition_ddf()` itself is a small function with 4 steps:
1. Collect dataset_metadata
2. compute repartition_ratio using `target_nrows_per_partition`
3. calculate `target_npartitions` using above ratio.
4. repartition the ddf and optimize it.


If I divide the function into 2 smaller funcs again, I need to send all the parameters from `compute_repartition_ddf()` to the `_calc_target_npartitions()` as well.

```
def compute_repartition_ddf(ddf, dataset_factory, target_nrows_per_partition, frac, predicates , ):
      code block here
     _calc_target_npartitions(ddf, dataset_factory, target_nrows_per_partition, frac, predicates)
  
```

I have updated the description for the same.
Please confirm if the single function looks fine.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: They are random numbers - >1,  <1 , ==1

kartothek/io/dask/dataframe.py
usha-nemani-by: We are removing this functionality. Please review the latest code changes.

kartothek/io/dask/dataframe.py
usha-nemani-by: `calculate_repartion_ratio()` function is removed.

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,98 @@
...
+    "df,target_nrows_per_partition",
+    (
+        (
+            pd.DataFrame({"A": range(100), "part": ["A", "B", "C", "D"] * 25}),
+            97,
+        ),  # rep_ratio = 3.88
+        (pd.DataFrame({"A": range(10), "part": range(10)}), 1),  # rep_ratio = 1.0
+    ),
+)
+def test_calculate_repartition_ratio_approximate_match(
```
lr4d, line 57:
```suggestion
def test_repartition_ddf_approximate_match(
```

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +607,53 @@ def hash_dataset(
...
+    dataset_factory
+        dataset factory
+    predicates
+        Optional , but predicates can be applied while collecting_dataset_metadata
+    target_nrows_per_partition
+        target num of rows per partition
+    dataset_metadata_frac
+        This fraction is helps in collecting metadata on a fraction(sample) of the ddf, instead of the whole ddf.
+
+
```
lr4d, line 639: I was thinking if we should add `chunksize`as dask allows i.e. the uncompressed size per partition. That is probably more valuable than `target_nrows_per_partition`

kartothek/io/dask/dataframe.py
```diff
@@ -606,3 +607,53 @@ def hash_dataset(
...
+    The input ddf is repartitioned using target_npartitions value.
+
+    Parameters
+    ----------
+    ddf
+        dask.DataFrame
+    dataset_factory
+        dataset factory
+    predicates
+        Optional , but predicates can be applied while collecting_dataset_metadata
```
lr4d, line 633:
```suggestion
    predicates
        Optional , but predicates can be applied while collecting_dataset_metadata.
        Note that predicates will only be applied during query planning, predicate pushdown is not (yet) supported.
        Thus, this is only useful if providing predicates involving indexed columns.
```

tests/io/dask/dataframe/test_repartition.py
```diff
@@ -0,0 +1,98 @@
...
+#     )
+#     ddf = read_dataset_as_ddf(store=store_factory, dataset_uuid="test")
+#     optimized_ddf = dask.optimize(ddf, optimizations=[dask.dataframe.optimize])[0]
+#     repartitioned_ddf = repartition_ddf(ddf, repartition_ratio=repartition_ratio)
+#     if repartition_ratio >= 1:
+#         assert len(repartitioned_ddf.dask) <= len(
+#             optimized_ddf.dask
+#         )  # assert the number of tasks in dask graph
+#     else:
+#         assert len(repartitioned_ddf.dask) > len(optimized_ddf.dask)
```
lr4d, line 98: We'll want to replace this with something analogous for repartition_ddf (target_nrows or chunksize)

lr4d (changes requested):

lr4d:
> If I divide the function into 2 smaller funcs again, I need to send all the parameters from repartition_ddf() to the _calc_target_npartitions() as well.

I think it makes sense to isolate the "calculate" logic from the repartitioning logic though.

tests/io/dask/dataframe/test_repartition.py
usha-nemani-by: The first two testcase parameters are ddf and  target_nrows . So I was thinking, is this testcase required at all?

kartothek/io/dask/dataframe.py
usha-nemani-by: We should fix about this Parameter as this would be the customer facing one. 
@fjetter , any thoughts here?

usha-nemani-by:
> > If I divide the function into 2 smaller funcs again, I need to send all the parameters from repartition_ddf() to the _calc_target_npartitions() as well.
> 
> I think it makes sense to isolate the "calculate" logic from the repartitioning logic though.

I have isolated the calculation logic. Please verify and confirm.

kartothek/io/dask/dataframe.py
usha-nemani-by: The logic is updated completely. Please review new logic and give suggestions.

usha-nemani-by:
@fjetter / @lr4d , 
In the previous PR  (#357 ), there was a suggestion to call   `repartition_ddf()` in   [https://github.com/JDASoftwareGroup/kartothek/blob/9e4909102c0b5d95b36d25136a6479fb71c644ab/kartothek/io/dask/dataframe.py#L336](url)  

If that is the case, do I  need to  remove the `repartition_ratio` from input params, and change the signature of that function?

[https://github.com/JDASoftwareGroup/kartothek/blob/9e4909102c0b5d95b36d25136a6479fb71c644ab/kartothek/io/dask/dataframe.py#L431](url)

lr4d:
After discussion involving Usha, Florian and I, it was decided to close this until we agree on an API interface for this functionality

PR #369, opened 2020-11-05
Correctly handle dispatch_by=[]

### LINKED ISSUE #368: Argument dispatch_by=[] is ignored when reading dataset
### Problem description

Currently passing `dispatch_by = []` to `read_...` behaves as passing `dispatch_by = None`.

```python
In [2]: import pandas as pd
   ...: import storefact
   ...: from functools import partial
   ...: from kartothek.io.eager import store_dataframes_as_dataset
   ...: from kartothek.io.dask.dataframe import read_dataset_as_ddf
   ...: 
   ...: path = "/tmp"
   ...: uuid = "dataset"
   ...: df = pd.DataFrame({"x": [1, 2], "y": ["a", "b"]})
   ...: store = partial(storefact.get_store_from_url, f"hfs://{path}?create_if_missing=False")
   ...: dm = store_dataframes_as_dataset(
   ...:     dfs=[df],
   ...:     dataset_uuid=uuid,
   ...:     store=store,
   ...:     partition_on=["x"],
   ...:     overwrite=True,
   ...: )
   ...: 
   ...: ddf = read_dataset_as_ddf(dataset_uuid=uuid, store=store, dispatch_by=[])
   ...: ddf.npartitions
   ...: 
Out[2]: 2

```

### Used versions

This was run on master.

mlondschien:
Closes #368 

Not sure why `mypy` wasn't complaining before. Tests and Changelog are still open.

### DIFF
--- a/kartothek/io_components/read.py
+++ b/kartothek/io_components/read.py
@@ -61,7 +61,7 @@ def dispatch_metapartitions_from_factory(
             DeprecationWarning,
         )
 
-    if dispatch_by and concat_partitions_on_primary_index:
+    if dispatch_by is not None and concat_partitions_on_primary_index:
         raise ValueError(
             "Both `dispatch_by` and `concat_partitions_on_primary_index` are provided, "
             "`concat_partitions_on_primary_index` is deprecated and will be removed in the next major release. "
@@ -74,7 +74,7 @@ def dispatch_metapartitions_from_factory(
         )
         dispatch_by = dataset_factory.partition_keys
 
-    if dispatch_by and not set(dispatch_by).issubset(
+    if dispatch_by is not None and not set(dispatch_by).issubset(
         set(dataset_factory.index_columns)
     ):
         raise RuntimeError(
--- a/kartothek/io_components/read.py
+++ b/kartothek/io_components/read.py
@@ -51,7 +51,6 @@ def dispatch_metapartitions_from_factory(
     dispatch_by: Optional[List[str]] = None,
     dispatch_metadata: bool = False,
 ) -> Union[Iterator[MetaPartition], Iterator[List[MetaPartition]]]:
-
     if dispatch_metadata:
 
         warnings.warn(
--- a/kartothek/io_components/read.py
+++ b/kartothek/io_components/read.py
@@ -108,16 +108,20 @@ def dispatch_metapartitions_from_factory(
         if isinstance(ix, ExplicitSecondaryIndex)
     }
 
-    if dispatch_by:
+    if dispatch_by is not None:
         base_df = cast(pd.DataFrame, base_df)
 
-        # Group the resulting MetaParitions by partition keys or a subset of those keys
-        merged_partitions = base_df.groupby(
-            by=list(dispatch_by), sort=True, as_index=False
-        )
+        if dispatch_by == []:
+            merged_partitions = [((""), base_df)]
+        else:
+            # Group the resulting MetaParitions by partition keys or a subset of those keys
+            merged_partitions = base_df.groupby(
+                by=list(dispatch_by), sort=True, as_index=False
+            )
+
         for group_name, group in merged_partitions:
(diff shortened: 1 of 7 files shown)


mlondschien:
It seems that the `normalize_args` method automatically replaces `None` with `[]`. Is this the expected behaviour here? I know that loading a dataset by a single worker is not a typical use-case.

fjetter:
Yes, this is expected behaviour. I want to get rid of this function since it's too much magic for the complexity but the idea is that we can rely on certain kwargs to always have a given type. We do not account for sentinels in the normalization.
There are some tests for this https://github.com/JDASoftwareGroup/kartothek/blob/462a7b060902af8101b54ebca304045b2336ac39/tests/io_components/test_utils.py#L82

kartothek/io_components/read.py
```diff
@@ -108,16 +107,20 @@ def dispatch_metapartitions_from_factory(
...
 
-    if dispatch_by:
+    if dispatch_by is not None:
         base_df = cast(pd.DataFrame, base_df)
 
-        # Group the resulting MetaParitions by partition keys or a subset of those keys
-        merged_partitions = base_df.groupby(
-            by=list(dispatch_by), sort=True, as_index=False
-        )
+        if dispatch_by == []:
```
fjetter, line 113: I would prefer `if dispatch_by` or `if len(dispatch_by) == 0`. I don't trust the comparison to a mutable object, even though it is empty. The `if dispatch_by` is probably considered the most "pythonic" way

kartothek/io_components/utils.py
```diff
@@ -258,6 +257,13 @@ def _make_list(_args):
             return old_value
         else:
             return _make_list(old_value)
+    elif arg_name == "dispatch_by":
+        if old_value is None:
+            return []
+        elif isinstance(old_value, list):
+            return old_value
+        else:
+            return _make_list(old_value)
```
fjetter, line 266: Isn't this the same logic as the `_make_list`? What's the reason for this special case?

fjetter (changes requested):

kartothek/io_components/read.py
mlondschien: Here I wanted to actively differentiate from the `dispatch_by=None` case (for readability). However `if len(dispatch_by) == 0` is much more elegant.

kartothek/io_components/utils.py
mlondschien: My mistake. This should have been
```python
        if old_value is None:
            return old_value
```

mlondschien:
Still missing tests. Any suggestions where to put those @fjetter? Is it enough to test `dispatch_metapartitions_from_factory` or do we want integration like tests for the dependent functions `read_dataset_as_metapartitions__iterator`, `read_dataset_as_delayed_metapartitions`, `read_dataset_as_metapartitions_bag` or even `read_dataset_as_dataframes__iterator`, `read_dataset_as_ddf`, `read_dataset_as_dataframe_bag`, `read_dataset_as_delayed`?

mlondschien:
This is still missings tests, but to my understanding good to go otherwise. Any suggestion where to put tests?

lr4d:
Sorry for the delay @mlondschien.

The testing setup of kartothek is a bit messy (for historical reasons, we'll want to refactor that eventually). 

My suggestion would be to put a test in https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/io/testing/read.py for this, similar to https://github.com/JDASoftwareGroup/kartothek/blob/18370b3c8f48ac794cf111e71248800c5bf720c8/kartothek/io/testing/read.py#L428 .
That would then be (automatically) tested for all io back-ends

mlondschien:
Not quite sure why codecov is complaining. To my understanding the new test is running. @l4rd, could you have another look?

fjetter (approved):
Thanks @mlondschien and sorry for the long wait

PR #376, opened 2020-11-24
Fix conversion of 'store' argument

### LINKED ISSUE #375: Cannot use store decorator as store argument
Cannot use store decorator, because of this isinstance check:
https://github.com/JDASoftwareGroup/kartothek/blame/master/kartothek/core/utils.py#L83

However, simplekv adopts a duck-typing approach, see e.g. https://github.com/mbr/simplekv/blob/master/simplekv/decorator.py#L6

This blocks migration from kartothek 3.14 for our code base.

jochen-ott-by:
Resolves #375

### DIFF
--- a/kartothek/core/utils.py
+++ b/kartothek/core/utils.py
@@ -1,6 +1,6 @@
 import pickle
 from functools import partial
-from typing import cast
+from typing import Any, cast
 
 from simplekv import KeyValueStore
 from storefact import get_store_from_url
@@ -51,22 +51,33 @@ def ensure_string_type(obj):
         return str(obj)
 
 
+def _is_simplekv_key_value_store(obj: Any) -> bool:
+    """
+    Check whether ``obj`` is the ``simplekv.KeyValueStore``-like class.
+
+    simplekv uses duck-typing, e.g. for decorators. Therefore,
+    avoid `isinstance(store, KeyValueStore)`, as it would be unreliable. Instead,
+    only roughly verify that `store` looks like a KeyValueStore.
+    """
+    return hasattr(obj, "iter_prefixes")
+
+
 def ensure_store(store: StoreInput) -> KeyValueStore:
     """
-    Ensure that the input is a valid KeyValueStore.
+    Convert the ``store`` argument to a ``KeyValueStore``, without pickle test.
     """
-    # This function is often used in an eager context where we may allow non-serializable stores, i.e. we do not need to check for serializability
-    if isinstance(store, KeyValueStore):
+    # This function is often used in an eager context where we may allow
+    # non-serializable stores, so skip the pickle test.
+    if _is_simplekv_key_value_store(store):
         return store
     return lazy_store(store)()
 
 
-def _factory_from_store(pickled_store):
+def _identity(store: KeyValueStore) -> KeyValueStore:
     """
-    Ensure that we're using internally only deserialized stores / proper
-    factories to avoid funny business with unix sockets in forked subprocesses.
+    Helper function for `lazy_store`.
     """
-    return pickle.loads(pickled_store)
+    return store
 
 
 def lazy_store(store: StoreInput) -> StoreFactory:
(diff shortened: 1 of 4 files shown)


kartothek/core/utils.py
```diff
@@ -51,22 +51,33 @@ def ensure_string_type(obj):
...
 
+def _is_simplekv_key_value_store(obj: Any) -> bool:
+    """
+    Check whether ``obj`` is the ``simplekv.KeyValueStore``-like class.
+
+    simplekv uses duck-typing, e.g. for decorators. Therefore,
+    avoid `isinstance(store, KeyValueStore)`, as it would be unreliable. Instead,
+    only roughly verify that `store` looks like a KeyValueStore.
+    """
+    return hasattr(obj, "iter_prefixes")
```
fjetter, line 62: I'm fine with duck typing and I guess the iter_prefixes is a good classifier for a simplekv store since I doubt it is used anywhere else. However, what we mostly rely upon is `get`, `put`, `open` and `__contains__` and I'm wondering if this check should be more tailored to what we need.
the iter_prefixes is "only" used in the disovery api

fjetter: otoh, we'll easily miss something and we shouldn't overdo it here. Just asking for an opinion

fjetter (approved):
I think this proposal is a good enough stretch between correctness/safety and practicality. I'd be fine with merging it but welcome to other opinions on the matter
cc @marco-neumann-by

kartothek/core/utils.py
lr4d: I agree that this could be "nicer" but I'd say it's fine for now

marco-neumann-by: Would `isinstance(obj, (KeyValueStore, StoreDecorator))` work?

jochen-ott-by: > Would `isinstance(obj, (KeyValueStore, StoreDecorator))` work?

No.

marco-neumann-by: Sure ducktyping is nice but IMHO it's nonsense when you wanna dispatch actions based on types.

jochen-ott-by: > Sure ducktyping is nice but IMHO it's nonsense when you wanna dispatch actions based on types.

Why? You can dispatch on the "duck-type", as done here. There is a chance that the type is not well-defined, but this is the same problem with "classical" types that allow multiple inheritance. I think this boils down to a API problem (I think the API should allow less types here), but this is hard to change now in a backward-compatible way.

marco-neumann-by: Well, you can dispatch on the duck-type if you can sufficiently check the contract the duck-type has. This then brings us back to the comment Florian had about which methods to check for (and is checking for existence even enough?). I see that you deliberatly chose a method that seems to be used only by simplekv (which btw [is not true](https://github.com/search?l=Python&q=iter_prefixes&type=Code)). I would say: instead of risking weird bugs by doing some semi-documented checks, I would rather check for the two base types that simplekv provides (and which to my knownledge are also used by all relevant code paths).

jochen-ott-by: > Well, you can dispatch on the duck-type if you can sufficiently check the contract the duck-type has.

This does not have to be accurate here. In fact, I think this is ONLY convenience to raise an exception early. I think we could skip the check entirely and it would still be "correct".

> This then brings us back to the comment Florian had about which methods to check for (and is checking for existence even enough?). I see that you deliberatly chose a method that seems to be used only by simplekv (which btw [is not true](https://github.com/search?l=Python&q=iter_prefixes&type=Code)).

Citing some third-party packages is beside the point; this does not have to be accurate.

> I would say: instead of risking weird bugs by doing some semi-documented checks, I would rather check for the two base types that simplekv provides (and which to my knownledge are also used by all relevant code paths).

I'd rather leave as it is. But if this is a blocker for this PR and the consensus of the kartothek owners is to do as you suggest, I can change it.

marco-neumann-by: Well, since @fjetter agreed to this change, you are still convinced this is a good idea (which is totally fine in a technical discussion), and this change isn't impactful enough for me to pull out some veto, I would say: go ahead.

PR #386, opened 2020-12-22
CI: test pandas nightly

jorisvandenbossche:
I was checking if kartothek tested against pandas nightly / rc (to otherwise do a test PR to check pandas 1.2.0rc), and noticed that the current spelling of the pip install didn't actually work to install pandas from the nightly wheels.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
         include:
           - numfocus_nightly: true
             os: "ubuntu-latest"
-            pyarrow: "0.17.1"
+            pyarrow: "2.0.0"
             python: "3.8"
           - numfocus_nightly: false
             os: "macos-latest"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,7 @@ jobs:
         run: conda install --file=conda-test-requirements.txt
       - name: Pip Instal NumFOCUS nightly
         # NumFOCUS nightly wheels, contains numpy and pandas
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -94,16 +94,16 @@ jobs:
...
       - name: Conda Export
         run: conda list --export
 
       # Linters
       - name: Flake8
         run: flake8
-      - name: Mypy
-        run: mypy .
+      # - name: Mypy
+      #   run: mypy .
```
lr4d, line 106: ^

jorisvandenbossche: I commented this out temporarily, because mypy is actually failing on the `numfocus_nightly: true` build. And it then doesn't run tests if it fails on that step, and so I wanted to see if the tests are actually passing with the nightly builds, and thus needed to comment this.

jorisvandenbossche:
The tests seem to be passing, so that's good!

As mentioned in the inline comment, mypy is failing on that build, though. So I can uncomment that again, but that will mean that the tests are still not run on the numfocus_nightly build until mypy is fixed. I can also skip mypy just for that build.

.github/workflows/ci.yml
```diff
@@ -104,6 +104,7 @@ jobs:
         run: flake8
       - name: Mypy
         run: mypy .
+        if: !matrix.numfocus_nightly
```
lr4d, line 107: Github actions is complaining about the syntax
Maybe this is worth trying?
```suggestion
        if: ${{ !matrix.numfocus_nightly }}
```

lr4d (commented):
Thanks for doing this :)

lr4d (approved):

jorisvandenbossche:
Thanks for the fixup!

lr4d:
My pleasure. Congratulations on your first kartothek commit ;)

PR #390, opened 2021-01-12
cubes: add option to suppress auto index creation for dimension columns

jochen-ott-by:
For large datasets, the automatic index creation can be a performance problem. Also, at some point of dataset size, there is the risk that the 2GB limit pyarrow has is hit. Therefore, it is desirable to switch off thye current default behaviour of creating a secondary index for all dimension columns.

This change would keep the current default of creating an index for all dimension columns. It adds  `cube.suppress_index_on`, which can be set to the subset of dimension columns to suppress the index creation for.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,14 @@
 Changelog
 =========
 
-Version 3.17.3 (2010-12-04)
+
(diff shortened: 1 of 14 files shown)


marco-neumann-by (approved):
Thanks!

PR #391, opened 2021-01-12
fix master

marco-neumann-by:
(hopefully)


fjetter (approved):
Thanks for looking into this!

PR #392, opened 2021-01-14
Setup docs test

### LINKED ISSUE #395: Isolate CI workflows/jobs into lint/test/release
### Problem description
If CI fails at the moment, it could be because linting has gone wrong, test have failed or the release has failed.

IMO, it makes more sense to have a separate workflow/job for linting, and for release, isolated from testing. This way, it is easier for a contributor to spot what part of the job has failed.

Isolating linting from testing would also speed up the CI as we would no longer need to lint for every combination of the CI matrix.

Separating the workflows/jobs into its separate components would also simplify the CI code, which is becoming quite spaghetti-like

fjetter:
Doc builds fail on readthedocs. This has been an issue a few times. This will setup a workflow which tests the configuration used there.

closes #395

### DIFF
--- a/.github/workflows/ci-pre-commit.yml
+++ b/.github/workflows/ci-pre-commit.yml
@@ -1,0 +1,11 @@
+name: pre-commit
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,7 @@ jobs:
           python -c "import kartothek.api"
           python -c "import kartothek.cli"
           python -c "import kartothek.core"
+          python -c "import not_existing"
           python -c "import kartothek.io"
           python -c "import kartothek.io_components"
           python -c "import kartothek.serialization"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,12 +109,6 @@ jobs:
       - name: Pytest
         run: pytest --cov kartothek --cov-report xml
 
-      - name: Sphinx
-        run: |
-          python setup.py build_sphinx
-          touch ./docs/_build/html/.nojekyll
-        if: env.IS_MASTER_BUILD == 'true'
-
       # Builds
       - name: Build Wheel
         run: python setup.py sdist bdist_wheel
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,13 +90,14 @@ jobs:
           python -c "import kartothek.api"
           python -c "import kartothek.cli"
           python -c "import kartothek.core"
-          python -c "import not_existing"
           python -c "import kartothek.io"
           python -c "import kartothek.io_components"
           python -c "import kartothek.serialization"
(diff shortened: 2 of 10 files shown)


fjetter:
There are apparently some actual failure in the content / doctests. Will need to investigate further but will probably take some time since this is currently not high prio. If somebody else wants to have a look, be my guest

marco-neumann-by:
One question here would also be if we wanna move from RTD to github pages for doc hosting so we have more control over the build environment and a unified setup + access management + notification system.

fjetter:
I don't have a strong opinion about readthedocs

lr4d:
>One question here would also be if we wanna move from RTD to github pages for doc hosting so we have more control over the build environment and a unified setup + access management + notification system.

I'd be curious on your opinion on this @marco-neumann-by . (I don't have a strong opinion either)

marco-neumann-by:
> > One question here would also be if we wanna move from RTD to github pages for doc hosting so we have more control over the build environment and a unified setup + access management + notification system.
> 
> I'd be curious on your opinion on this @marco-neumann-by . (I don't have a strong opinion either)

My opinion is that these are all desirable attributes to have for a build service. However there is one drawback: RTD also archives old doc versions. The question is if that is something our users really use.

fjetter:
I'm sold on GH pages. readthedocs only supports pip environments and pip (or rather wheels) are just too problematic. I ended up needing to pin numpy because the docs execute stuff triggering the broken ABI :(

xhochy:
@fjetter It does support `conda` https://docs.readthedocs.io/en/stable/guides/conda.html and using the [`CONDA_USES_MAMBA`](https://docs.readthedocs.io/en/latest/guides/feature-flags.html#available-flags ) flag should also bring this down to very installations

fjetter:
Alright, then I'll give it another shot. Thanks for the pointers @xhochy . Any suggestions for the numpy pinning in general? I would like to avoid it here in kartothek but then again, there are probably few wheels out there which would be compatible

xhochy:
> Alright, then I'll give it another shot. Thanks for the pointers @xhochy . Any suggestions for the numpy pinning in general? I would like to avoid it here in kartothek but then again, there are probably few wheels out there which would be compatible

I have patched all conda package but that kind of patching is not supported for wheels / packages on PyPI. Thus only the new `pyarrow=3` wheels are compatible with `numpy>=1.20`.

fjetter:
Ok so the original issue is/was that the docs didn't have all requirements installed, in particular there was a bug regarding the existance of distributed which I addressed here.

This unfortunately wasn't sufficient and I ended up running into a conda issue where `click` was uninstalled when installing the test requirements. I reduced the number of requirements we put on top, factored out the linting into a dedicated job and froze the "production" environment before we install the tests. That seemed to do the trick

I also noticed that the ASV job is kind of useless since asv returns a zero exit code even if the benchmarks fail. I believe we require a different solution to keep those tested.


Setting up conda for readthedocs was rahter simple, actually, but I didn't enable mamba since it's not self service and I don't want to play ping pong with their support in case we ever want to disable it again

fjetter:
closes #395

fjetter:
oh, this click thing is again failing :/ for some reason it is removed / corrupted by the second (or Nth) install. I'll see if I can clean the script up a bit further

fjetter:
ok, it's not clicks fault. apparently our CLI starts an ipython kernel but it is not listed as a runtime requirement. I removed it to the docs since I figured we just used it for docs generation :/

PR #401, opened 2021-02-04
add retries for ParquetSerializer.restore_dataframe

lr4d:
Retries in this setup go from 10ms to 160 ms (specifically: `0.01 0.02 0.04 0.08 0.16` seconds).

This supersedes the first commit from https://github.com/JDASoftwareGroup/kartothek/pull/397.

- [ ] Closes #xxxx
- [x] Changelog entry

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -213,7 +213,8 @@ class ParquetSerializer(DataFrameSerializer):
                     predicates=predicates,
                     date_as_object=date_as_object,
                 )
-            except (AssertionError, OSError):
+            except (AssertionError, OSError) as err:
+                raised_error = err
                 _logger.warning(
                     msg=(
                         f"Failed to restore dataframe, attempt {nb_retry + 1} of {MAX_NB_RETRIES} with parameters "
@@ -232,7 +233,7 @@ class ParquetSerializer(DataFrameSerializer):
             f"key: {key}, filter_query: {filter_query}, columns: {columns}, "
             f"predicate_pushdown_to_io: {predicate_pushdown_to_io}, categories: {categories}, "
             f"date_as_object: {date_as_object}, predicates: {predicates}."
-        )
+        ) from raised_error
 
     def store(self, store, key_prefix, df):
         key = "{}.parquet".format(key_prefix)
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -140,7 +140,7 @@ class ParquetSerializer(DataFrameSerializer):
             "Failed to restore dataframe after {MAX_NB_RETRIES} attempts. Parameters: "
             f"key: {key}, filter_query: {filter_query}, columns: {columns}, "
             f"predicate_pushdown_to_io: {predicate_pushdown_to_io}, categories: {categories}, "
-            f"predicates: {predicates}, date_as_object: {date_as_object}."
+            f"date_as_object: {date_as_object}, predicates: {predicates}."
         )
 
     def store(self, store, key_prefix, df):
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -6,6 +6,8 @@ This module contains functionality for persisting/serialising DataFrames.
 
 
 import datetime
+import logging
+import time
 from typing import Iterable, Optional
 
 import numpy as np
@@ -33,8 +35,12 @@ try:
 except ImportError:
     HAVE_BOTO = False
 
+_logger = logging.getLogger(__name__)
+
 
 EPOCH_ORDINAL = datetime.date(1970, 1, 1).toordinal()
+MAX_NB_RETRIES = 6  # longest retry backoff = BACKOFF_TIME * 2**(MAX_NB_RETRIES - 2)
+BACKOFF_TIME = 0.01  # 10 ms
 
 
 def _empty_table_from_schema(parquet_file):
@@ -63,6 +69,14 @@ def _reset_dictionary_columns(table, exclude=None):
 
     table = table.cast(schema)
     return table
+
+
+class ParquetReadError(IOError):
+    """
+    Internal kartothek error while attempting to read Parquet file
+    """
+
+    pass
 
 
 class ParquetSerializer(DataFrameSerializer):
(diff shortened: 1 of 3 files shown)


kartothek/serialization/_parquet.py
```diff
@@ -95,85 +109,39 @@ def restore_dataframe(
...
+                # we don't sleep when we're done with the last attempt
+                if nb_retry < (MAX_NB_RETRIES - 1):
+                    time.sleep(BACKOFF_TIME * 2 ** nb_retry)
+
+        raise ParquetReadError(
+            "Failed to restore dataframe after {MAX_NB_RETRIES} attempts. Parameters: "
+            f"key: {key}, filter_query: {filter_query}, columns: {columns}, "
+            f"predicate_pushdown_to_io: {predicate_pushdown_to_io}, categories: {categories}, "
+            f"predicates: {predicates}, date_as_object: {date_as_object}."
+        )
```
fjetter, line 144: please use 

```
raise ParquetReaderError(...) from original_excetion
```

to not loose that context information

----

I suggest to put the predicates last since they are the longest and may get lost if the exception message is truncated for whatever reason (e.g. distributed truncates at 1000chars)

kartothek/serialization/_parquet.py
```diff
@@ -95,85 +109,39 @@ def restore_dataframe(
...
+                        f"predicates: {predicates}, date_as_object: {date_as_object}."
+                    ),
+                    exc_info=True,
+                )
+                # we don't sleep when we're done with the last attempt
+                if nb_retry < (MAX_NB_RETRIES - 1):
+                    time.sleep(BACKOFF_TIME * 2 ** nb_retry)
+
+        raise ParquetReadError(
+            "Failed to restore dataframe after {MAX_NB_RETRIES} attempts. Parameters: "
```
fjetter, line 140: an `f` is missing. This would render without formatting but I would expect the intention was to have MAX_NB_RETRIES to be formatted

kartothek/serialization/_parquet.py
```diff
@@ -195,6 +163,94 @@ def store(self, store, key_prefix, df):
         return key
 
 
+def _restore_dataframe(
```
fjetter, line 166: just a note regarding git diffs and git blame:

git has a much easier time figuring the history out if you actually do not change too much. Instead, you're moving this entire code section, i.e. we're effectively lossing easier git blame.
This is mostly a style note but I would prefer to define this as a static_method as well to keep diffs small

kartothek/serialization/_parquet.py
```diff
@@ -95,85 +109,39 @@ def restore_dataframe(
...
+                        f"predicate_pushdown_to_io: {predicate_pushdown_to_io}, categories: {categories}, "
+                        f"predicates: {predicates}, date_as_object: {date_as_object}."
+                    ),
+                    exc_info=True,
+                )
+                # we don't sleep when we're done with the last attempt
+                if nb_retry < (MAX_NB_RETRIES - 1):
+                    time.sleep(BACKOFF_TIME * 2 ** nb_retry)
+
+        raise ParquetReadError(
```
fjetter, line 139: What's your reasoning behind introducing a dedicated class of exceptions? The information it contains is identical to the log message and the type of exception is not reused anywhere

fjetter (changes requested):

kartothek/serialization/_parquet.py
lr4d:
```suggestion
            f"Failed to restore dataframe after {MAX_NB_RETRIES} attempts. Parameters: "
```

kartothek/serialization/_parquet.py
```diff
@@ -95,85 +109,39 @@ def restore_dataframe(
...
-        else:
-            df = filter_df(df, filter_query)
-        if columns is not None:
-            return df.reindex(columns=columns, copy=False)
-        else:
-            return df
+            except (AssertionError, OSError):
+                _logger.warning(
+                    msg=(
+                        f"Failed to restore dataframe, attempt {nb_retry} of {MAX_NB_RETRIES} with parameters "
```
lr4d, line 128:
```suggestion
                        f"Failed to restore dataframe, attempt {nb_retry + 1} of {MAX_NB_RETRIES} with parameters "
```

kartothek/serialization/_parquet.py
lr4d: the problem is that the exception is out of scope at this point

kartothek/serialization/_parquet.py
```diff
@@ -95,85 +109,39 @@ def restore_dataframe(
...
+                )
+                # we don't sleep when we're done with the last attempt
+                if nb_retry < (MAX_NB_RETRIES - 1):
+                    time.sleep(BACKOFF_TIME * 2 ** nb_retry)
+
+        raise ParquetReadError(
+            "Failed to restore dataframe after {MAX_NB_RETRIES} attempts. Parameters: "
+            f"key: {key}, filter_query: {filter_query}, columns: {columns}, "
+            f"predicate_pushdown_to_io: {predicate_pushdown_to_io}, categories: {categories}, "
+            f"predicates: {predicates}, date_as_object: {date_as_object}."
```
lr4d, line 143:
```suggestion
            f"date_as_object: {date_as_object}, predicates: {predicates}."
```

kartothek/serialization/_parquet.py
lr4d: I guess I can work around it

kartothek/serialization/_parquet.py
lr4d: Made `_restore_dataframe` into static method, `restore_dataframe` into class method

kartothek/serialization/_parquet.py
lr4d: The exception type describes the problem and it being a subtype of IOError is appropriate imo. I'd rather be too specific here than too broad

kartothek/serialization/_parquet.py
```diff
@@ -175,6 +189,52 @@ def restore_dataframe(
...
 
+    @classmethod
+    def restore_dataframe(
+        cls,
+        store: KeyValueStore,
+        key: str,
+        filter_query: Optional[str] = None,
+        columns: Optional[Iterable[str]] = None,
+        predicate_pushdown_to_io: bool = True,
+        categories: Optional[Iterable[str]] = None,
```
jochen-ott-by, line 200: I think you need stronger guaranteed here than Iterable; after all, the first try to consume those iterators, leaving wrong arguments on the second try.

kartothek/serialization/_parquet.py
```diff
@@ -175,6 +189,52 @@ def restore_dataframe(
...
+        predicates: Optional[PredicatesType] = None,
+        date_as_object: bool = False,
+    ) -> pd.DataFrame:
```
jochen-ott-by, line 203: There should be some explanation here as to why there is a re-try, especially why there is a re-try on AssertionErrors, which is typically considered not a best practice and dangereous: AssertionErrors usually point to logic errors and logic errors can lead to invalid state of the program which is present even for the re-try. Therefore, a usual argument is to simply abort on assertion errors. So it;'s worth explaining why deviating from this here.

kartothek/serialization/_parquet.py
fjetter: Probably `Sequence` is then a better candidate. It is very likely that this is not 100% properly done in many places.

Imho, it's not entirely clear, though. afaik, the difference between `Sequence` and `Iterable` is that a sequence must define `__len__` and `__getitem__` but that would exclude sets.

From your comment I would judge we need something like "iterable but not generator" but I do not know which type that should be

jochen-ott-by: If you really need to support sets and lists, I think the right type is `Collection[str]`. OTOH, why not simply use something concrete here that reflects how it's used?

For instance, `categories` needs to be a list already, as it's passed to an external function expecting a list.
`columns` can;t be a set today because some functions call `columns[:]` to copy it, which does not work for sets ...

fjetter: If it is that explicit, I would argue it is even mandatory.
Using a sequence/iterable instead of explicit list is a recommendation of mypy since, e.g. sequence is covariant in its variables where a list isn't. It also makes it very explicit that the object is not mutated and allows for more compatible typing. 

Some documentation about this can be found here https://mypy.readthedocs.io/en/latest/generics.html#variance-of-generic-types 

That all said, I'm fine either way. If mypy doesn't complain, it's good enough for me

fjetter: For my own curiosity

```python
from collections import Iterable, Sequence
In [4]: isinstance([1, 2], Sequence)
Out[4]: True

In [5]: isinstance((1, 2), Sequence)
Out[5]: True

In [6]: isinstance((1, 2), Iterable)
Out[6]: True

In [7]: isinstance({1, 2}, Iterable)
Out[7]: True

In [8]: isinstance({1, 2}, Sequence)
Out[8]: False

In [9]: isinstance((x for x in [1, 2]), Iterable)
Out[9]: True

In [10]: isinstance((x for x in [1, 2]), Sequence)
Out[10]: False
```

generators are not sequences

kartothek/serialization/_parquet.py
lr4d: I will just change the underlying error to IOError since that's more appropriate than AssertionError

kartothek/serialization/_parquet.py
lr4d: In my personal opinion, this should be `set[str]`, the logical data type but this would break the tests and is not really the "paradigm" we're using here anyways. I'd also rather have the type be explicit than duck typing.

I just copied the original function signature, we can probably have this debate somewhere else

kartothek/serialization/_parquet.py
lr4d: Added a comment and changed the BlockBuffer error to an error type inheriting from IOError

kartothek/serialization/_parquet.py
jochen-ott-by: As I said, set is wrong.
Why not just change it to list?

fjetter:
I'm fine with the changes here. Still, implementing a retry on this abstraction level really doesn't sit well with me. Can we at least create one github issue with the parquet AssertionError we've seen? afaik, we do not have any public breadcrumbs at all and I believe this helps with a justification for this retry if the issue is then linked in the code

kartothek/serialization/_parquet.py
```diff
@@ -175,6 +189,57 @@ def restore_dataframe(
...
+        # XXX: We have been seeing weired `IOError` while reading Parquet files from Azure Blob Store,
+        # thus, we implement retries at this point. This code should not live forever, it should be removed
+        # once the underlying cause has been resolved
+        for nb_retry in range(MAX_NB_RETRIES):
```
stephan-hesselmann-by, line 207: Is there a reason you implemented the retry logic itself? I would suggest to use an established library like [tenacity](https://github.com/jd/tenacity). This also has nice features such as being able to overwrite the wait function in tests.

Also I think you should add tests for the retry logic.

fjetter: I am very strongly against exposing this as some kind of configuration. Current investigations hint towards a bug in our azure-storage interface and/or the iobuffer interface.

Considering our simple requirement, I am in favour of this approach to not introduce tenacity as a core dependency

stephan-hesselmann-by: Fair enough. But especially with the self baked retries I think tests are important because the retry now contains untested code logic.

fjetter: Fair point about the testing. I would suggest to test this by patching the `_restore_dataframe` method to raise once but succeeds on the second try. While this test would be very close to the implementation, I believe this is justified for this case

cc @NeroCorleone

stephan-hesselmann-by: We expect the following behavior:
* Retry on `OSError` `MAX_NB_RETRIES` times or less if success. Log on retry
* Don't retry on other exceptions
* Raise `ParquetReadError` after `MAX_NB_RETRIES`
* Retry times use exponential backoff

I think all of these could be individual tests.

NeroCorleone:
@fjetter fyi https://github.com/JDASoftwareGroup/kartothek/issues/407
I'll reference it in the code.

PR #404, opened 2021-02-08
Add optional Parquet serializer to cube interface

stephan-hesselmann-by:
This allows to specify the compression algorithm and a chunk size to
generate row groups.

The Parquet serializer parameter is added the following cube operations:
  * `build_cube`
  * `append_cube`
  * `extend_cube`
  * `update_cube`

Specifying different Parquet serializers is back- and forwards compatible.
Row groups are kept for partitions that are not touched during cube operations.
If partitions are replaced or newly created, they are rewritten by the
specified Parquet serializer.

# Description:

- [ ] Closes #xxxx
- [ ] Changelog entry

### DIFF
--- a/kartothek/io/dask/bag_cube.py
+++ b/kartothek/io/dask/bag_cube.py
@@ -46,6 +46,7 @@ def build_cube_from_bag(
     metadata=None,
     overwrite=False,
     partition_on=None,
+    df_serializer=None,
 ):
     """
     Create dask computation graph that builds a cube with the data supplied from a dask bag.
@@ -68,6 +69,8 @@ def build_cube_from_bag(
     partition_on: Optional[Dict[str, Iterable[str]]]
         Optional parition-on attributes for datasets (dictionary mapping :term:`Dataset ID` -> columns).
         See :ref:`Dimensionality and Partitioning Details` for details.
+    df_serializer: Optional[ParquetSerializer]
+        Optional Dataframe to Parquet serializer
 
     Returns
     -------
@@ -83,6 +86,7 @@ def build_cube_from_bag(
         metadata=metadata,
         overwrite=overwrite,
         partition_on=partition_on,
+        df_serializer=df_serializer,
     )
 
 
@@ -94,6 +98,7 @@ def extend_cube_from_bag(
     metadata=None,
     overwrite=False,
     partition_on=None,
+    df_serializer=None,
 ):
     """
     Create dask computation graph that extends a cube by the data supplied from a dask bag.
@@ -117,6 +122,8 @@ def extend_cube_from_bag(
     partition_on: Optional[Dict[str, Iterable[str]]]
         Optional parition-on attributes for datasets (dictionary mapping :term:`Dataset ID` -> columns).
         See :ref:`Dimensionality and Partitioning Details` for details.
+    df_serializer: Optional[ParquetSerializer]
+        Optional Dataframe to Parquet serializer
 
     Returns
     -------
@@ -132,6 +139,7 @@ def extend_cube_from_bag(
         metadata=metadata,
         overwrite=overwrite,
         partition_on=partition_on,
+        df_serializer=df_serializer,
     )
 
 
@@ -365,7 +373,9 @@ def cleanup_cube_bag(cube, store, blocksize=100):
     )
 
 
-def append_to_cube_from_bag(data, cube, store, ktk_cube_dataset_ids, metadata=None):
+def append_to_cube_from_bag(
+    data, cube, store, ktk_cube_dataset_ids, metadata=None, df_serializer=None
+):
     """
     Append data to existing cube.
 
@@ -394,6 +404,8 @@ def append_to_cube_from_bag(data, cube, store, ktk_cube_dataset_ids, metadata=No
     metadata: Optional[Dict[str, Dict[str, Any]]]
         Metadata for every dataset, optional. For every dataset, only given keys are updated/replaced. Deletion of
         metadata keys is not possible.
+    df_serializer: Optional[ParquetSerializer]
+        Optional Dataframe to Parquet serializer
 
     Returns
     -------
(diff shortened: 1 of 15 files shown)


kartothek/io/testing/build_cube.py
```diff
@@ -269,6 +273,52 @@ def test_parquet(driver, function_store):
         pdt.assert_frame_equal(df_actual.reset_index(drop=True), df_expected)
 
 
+@pytest.mark.filterwarnings("ignore::UnicodeWarning")
+@pytest.mark.parametrize("chunk_size", [None, 1, 2, 3])
```
fjetter, line 277: `[None, >0]` should be sufficient. No need to have multiple integers in here

kartothek/io/testing/build_cube.py
```diff
@@ -269,6 +273,52 @@ def test_parquet(driver, function_store):
         pdt.assert_frame_equal(df_actual.reset_index(drop=True), df_expected)
 
 
+@pytest.mark.filterwarnings("ignore::UnicodeWarning")
```
fjetter, line 276: Where is this coming from?

kartothek/io/testing/extend_cube.py
```diff
@@ -79,6 +84,86 @@ def test_simple(driver, function_store, existing_cube):
     assert set(ds.table_meta) == {SINGLE_TABLE}
 
 
+@pytest.mark.parametrize("chunk_size", [None, 1, 2, 3])
```
fjetter, line 87: same here, please reduce the multiplicity

kartothek/io/testing/extend_cube.py
```diff
@@ -79,6 +84,86 @@ def test_simple(driver, function_store, existing_cube):
...
+
+    part_num_rows = {0: 1, 1: 3}
+    part_chunk_size = {0: None, 1: None}
+
+    assert len(dataset.partitions) == 2
+    assert_num_row_groups(function_store(), dataset, part_num_rows, part_chunk_size)
+
+
+@pytest.mark.parametrize(
+    "compression_build", ["NONE", "SNAPPY", "GZIP", "BROTLI", "LZ4", "ZSTD"]
```
fjetter: same here, please reduce the multiplicity

kartothek/serialization/_parquet.py
```diff
@@ -69,8 +69,22 @@ class ParquetSerializer(DataFrameSerializer):
     _PARQUET_VERSION = "2.0"
     type_stable = True
 
-    def __init__(self, compression="SNAPPY", chunk_size=None):
+    def __init__(
+        self, compression: str = "SNAPPY", chunk_size: Optional[int] = None
+    ) -> None:
+        if compression not in ["NONE", "SNAPPY", "GZIP", "BROTLI", "LZ4", "ZSTD"]:
```
fjetter, line 75: I would assume this list is exported by arrow. Can you please check and reuse an arrow interface? Which algo is supported is 100% dictated by arrow and depends even on how it was built, therefore, I would prefer a more dynamic approach, if any

fjetter (changes requested):

kartothek/io/testing/append_cube.py
```diff
@@ -98,6 +103,117 @@ def test_append_partitions(driver, function_store, existing_cube):
...
+
+    part_num_rows = {0: 2, 1: 2, 2: 1, 3: 3}
+    part_chunk_size = {0: None, 1: None, 2: None, 3: None}
+
+    assert len(dataset.partitions) == 4
+    assert_num_row_groups(function_store(), dataset, part_num_rows, part_chunk_size)
+
+
+@pytest.mark.parametrize(
+    "compression_build", ["NONE", "SNAPPY", "GZIP", "BROTLI", "LZ4", "ZSTD"]
```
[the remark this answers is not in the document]
stephan-hesselmann-by: Would you agree that the test itself makes sense if the parameters are removed/hardcoded to e.g. SNAPPY and GZIP?

kartothek/io/testing/build_cube.py
stephan-hesselmann-by: Note that these do test different edge cases (e.g. whether number of rows is divisible by chunk size). But I agree this is probably the wrong abstraction level to test this.

kartothek/io/testing/append_cube.py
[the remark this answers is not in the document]
fjetter: from my POV what you are trying to test is "what happens if I create + update with differently configured serialiser instances." From that angle, I see value and it is good to have this explicit.
however, this is technically already done with the test on different chunk sizes above so there is some redundancy in this test, already.
If we keep one combination hard coded, with a comment next to it explaining why it is not parametrized, I'd be fine with keeping one compression test if you feel it adds value

kartothek/serialization/_parquet.py
stephan-hesselmann-by: I didn't find an exported list by arrow on first glance, perhaps because it is indeed based on the arrow build. I will just remove this check

kartothek/io/testing/build_cube.py
fjetter: indeed, the test makes sense but not on cube level. the only thing we actually want to test on cube level is "did I actually use the provided serializer or not".

You could even take this a level further by implementing a mock serializer and we merely verify if it was called. This way we'd not do any IO in the test. After all, we rely on the serializer to be implemented correctly so we don't need to test all peculiarities of it. We might finally have fast tests this way :D

Note: Completely fine if you don't go down that route but eventually I would hope we can test kartothek on this level instead of always performing full end-to-end tests

fjetter (approved):
LGTM. Thanks @stephan-hesselmann-by!

kartothek/io/dask/bag_cube.py
```diff
@@ -39,35 +44,38 @@
...
-    cube,
-    store,
-    ktk_cube_dataset_ids=None,
-    metadata=None,
-    overwrite=False,
-    partition_on=None,
-):
+    data: db.Bag,
+    cube: Cube,
+    store: Callable[[], KeyValueStore],
```
fjetter, line 49: FYI We have a type alias `StoreFactory` for this in `kartothek.core.typing`

stephan-hesselmann-by: Thanks for pointing this out. The type hints are overall quite confusing because sometimes store is a callable and sometimes not. I have simply followed the convention from the previous doc strings.

fjetter: Indeed, this is very confusing. With https://github.com/JDASoftwareGroup/kartothek/pull/345 I tried to change the behaviour such that user facing code just doesn't care but this was not done for the cubes and everything is overall pretty messy

stephan-hesselmann-by:
Rebased after https://github.com/JDASoftwareGroup/kartothek/pull/398 was merged

PR #408, opened 2021-02-10
Add retries to restore dataframe

NeroCorleone:
# Description:

Same as in https://github.com/JDASoftwareGroup/kartothek/pull/401
I don't have write access to lr4d's fork, so I can't push on his branch.

Baseline: Add retries for restoring dataframes. We have seen IOErrors on long running ktk + dask tasks with no clear idea what the root cause is. Therefore retry the serialization to gain more stability, until the root cause is fixed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,9 @@ Version 3.19.0 (2021-02-XY)
   state after the update
 * Expose compression type and row group chunk size in Cube interface via optional
   parameter of type :class:`~kartothek.serialization.ParquetSerializer`.
-* Add retries to  :func:`~kartothek.serialization.parquet.ParquetSerializer.restore_dataframe`
+* Add retries to  :func:`~kartothek.serialization._parquet.ParquetSerializer.restore_dataframe`
+  IOErrors on long running ktk + dask tasks have been observed. Until the root cause is fixed,
+  the serialization is retried to gain more stability.
 
 Version 3.18.0 (2021-01-25)
 ===========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,7 @@ Version 3.19.0 (2021-02-XY)
   state after the update
 * Expose compression type and row group chunk size in Cube interface via optional
   parameter of type :class:`~kartothek.serialization.ParquetSerializer`.
+* Add retries to  :func:`~kartothek.serialization.parquet.ParquetSerializer.restore_dataframe`
 
 Version 3.18.0 (2021-01-25)
 ===========================
--- a/tests/serialization/test_parquet.py
+++ b/tests/serialization/test_parquet.py
@@ -470,6 +470,37 @@ def test_retry_on_IOError(monkeypatch, caplog, store):
     this test and the workaround can be removed.
     """
 
+    df = pd.DataFrame({"A": [0, 1, 2, 3]})
+
+    retry_count = 0
+
+    def patched__restore_dataframe(**kwargs):
+        nonlocal retry_count
+        retry_count += 1
+
+        if not retry_count > 1:
+            # fail for the first try
+            raise IOError()
+        elif retry_count > 1:
+            # simulate a successful retry
+            return df
+
+    monkeypatch.setattr(
+        ParquetSerializer, "_restore_dataframe", patched__restore_dataframe
+    )
+    serializer = ParquetSerializer()
+    key = serializer.store(store, "key", df)
+    df_result = serializer.restore_dataframe(store=store, key=key)
+    pdt.assert_frame_equal(df, df_result)
+
+
+def test_retries_on_IOError_logs(monkeypatch, caplog, store):
+    """
+    See https://github.com/JDASoftwareGroup/kartothek/issues/407 :
+    We are testing a retry-workaround for the above issue here. Once the issue is resolved,
+    this test and the workaround can be removed.
+    """
+
     def patched__restore_dataframe(**kwargs):
         # This kind of exception should be captured by the retry mechanism.
         raise IOError()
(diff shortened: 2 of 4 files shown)


tests/serialization/test_parquet.py
```diff
@@ -459,3 +461,55 @@ def test_reset_dict_cols(store):
...
+
+
+def test_retry_on_IOError(monkeypatch, caplog, store):
+    """
+    See https://github.com/JDASoftwareGroup/kartothek/issues/407 :
+    We are testing a retry-workaround for the above issue here. Once the issue is resolved,
+    this test and the workaround can be removed.
+    """
+
+    def patched__restore_dataframe(**kwargs):
```
fjetter, line 473: This test shows that a `IOError` will be rewritten to a `ParquetReadError` and that we'll get three logs, not that we actually retry.

```python
count = 0
def patched__restore_dataframe(**kwargs):
    nonlocal count
    count += 1
    if count > 1:
        return
    else:
        raise IOError
```

stephan-hesselmann-by (commented):
Can you add another test for the exponential backoff? Could be implemented by patching the sleep function and asserting the arguments it was called with. Also in the other tests you could patch the sleep function as no-op so that there is no unwanted wait time during tests.

fjetter:
imho, I do not think we'll need to test the backoff. At some point we'll need to draw a line between pragmatism and thoroughness and I have the feeling that testing this blows this change out of proportion. After all, this is not the core functionality of kartothek but merely a patch. Whether or not this is backed off should not even matter

fjetter:
doc builds fail (yay, so glad I put this into GH actions; not yay for you :) )

CHANGES.rst
```diff
@@ -10,6 +10,7 @@ Version 3.19.0 (2021-02-XY)
   state after the update
 * Expose compression type and row group chunk size in Cube interface via optional
   parameter of type :class:`~kartothek.serialization.ParquetSerializer`.
+* Add retries to  :func:`~kartothek.serialization.parquet.ParquetSerializer.restore_dataframe`
```
fjetter, line 13:
```suggestion
* Add retries to  :func:`~kartothek.serialization.ParquetSerializer.restore_dataframe`
```

Please be a bit more specific about what this retry does

stephan-hesselmann-by:
> imho, I do not think we'll need to test the backoff. At some point we'll need to draw a line between pragmatism and thoroughness and I have the feeling that testing this blows this change out of proportion. After all, this is not the core functionality of kartothek but merely a patch. Whether or not this is backed off should not even matter

Perhaps a matter of opinion but I think if the backoff is implemented it should also be tested. I think that's what you have to pay by not using an already tested library that implements this functionality for you. I also think that a bug in the retry logic could have quite severe consequences...

NeroCorleone:
> Can you add another test for the exponential backoff?

I left that out on purpose, because I am not sure what we gain by testing that the waiting times are exponential -- to me, this sounds like an implementation detail and not the actual functionality.

tests/serialization/test_parquet.py
NeroCorleone: Technically true, I thought that asserting for `len(caplog.records) == MAX_NB_RETRIES`  **and** the corresponding log message would be a strong enough condition for the current setup...

stephan-hesselmann-by:
> > Can you add another test for the exponential backoff?
> 
> I left that out on purpose, because I am not sure what we gain by testing that the waiting times are exponential -- to me, this sounds like an implementation detail and not the actual functionality.

Well, it is a feature of Kartothek now because it has been implemented this way. I also don't get your reasoning: Should details of the implementation not be tested? Even if they are implemented from scratch? I do agree that such detailed retry functionality should not be part of Kartothek, hence my suggestion of using an established library for that. 

Imo what we gain is security against a bug in this "implementation detail". Let's say somehow a bug is introduced that causes the backoff to be on the scale of hours instead of ms. I think a test to prevent such regressions is beneficial.

NeroCorleone:
> Should details of the implementation not be tested?

No, I think they shouldn't, for several reasons: I want to test the functionality (gain more stability on certain kind of errors by retrying) and not how this functionality is implemented (a hand-made exponential retry).
However, I see your point about potential bugs by accidentally increasing the back-off time and also like the idea of using an already existing library, so I don't have to retry manually.

I am currently seeing two ways forward in this discussion:
1. Replace the handmade retry with something that is proven to work (would require a tiny little bit more work time)
2. Mitigate the risk for an accidental increase of the back-off time by adding an explicit comment (would be a weak safe guard).

I am fine either way, with a slight preference to the pragmatic approach.

CHANGES.rst
NeroCorleone: Done.

stephan-hesselmann-by:
We could also just remove the backoff and use a fixed time to reduce the complexity of the untested code part.

fjetter:
Whether one decides to test implementation details or not is a tough line to walk. Usually you do not want to be to close to the implementation and rather test on a high level to ensure the behaviour of your library is what you would expect. In my opinion, the backoff here is an example of a details we should skip

fjetter:
> We could also just remove the backoff and use a fixed time to reduce the complexity of the untested code part.

If you sleep for a fixed time, this is still a backoff and we'd still have the same problem.

fjetter:
> Let's say somehow a bug is introduced that causes the backoff to be on the scale of hours instead

That would fail the CI due to a timeout ;)

stephan-hesselmann-by (approved):
Well I think all arguments have been put forth, I'll leave it up to you. In any case this should not block us from merging this important fix.

tests/serialization/test_parquet.py
fjetter: I just want to ensure this is not an "all or nothing"

NeroCorleone: Done.

fjetter (approved):

fjetter:
Looks like this is caused by https://github.com/Azure/azure-sdk-for-python/issues/16723

PR #414, opened 2021-02-16
Fix sphinx refs

fjetter:
# Description:

This enables intersphinx for our most used dependencies and fixes a lot of wrong links which were causing wrongfully rendered docs. The workaround with the target replacing is unfortunate but I couldn't find a nicer way to address the issue. For some reason the type annotations are not properly rendered. For example

`ParquetSerializer` would become `kartothek.serialization._parquet.ParquetSerializer` but the only available sphinx target is the public import without `_parquet`. Same is for the import redirects of dask, pandas, etc. but I believe this is fine

Partially, I also converted our doc style type annotations to proper annotations in some places. Covering the entire code base is out of scope for this PR


kartothek/serialization/_parquet.py
```diff
@@ -99,18 +101,6 @@ def __init__(
...
-            isinstance(other, ParquetSerializer)
-            and (self.compression == other.compression)
-            and (self.chunk_size == other.chunk_size)
-        )
-
-    def __repr__(self):
-        return "ParquetSerializer(compression={compression!r}, chunk_size={chunk_size!r})".format(
-            compression=self.compression, chunk_size=self.chunk_size
-        )
-
```
fjetter, line 113: - [ ] oops

kartothek/io/dask/common_cube.py
```diff
@@ -101,7 +101,7 @@ def build_cube_from_bag_internal(
 
     Parameters
     ----------
-    data:
+    data: dask.bag.Bag
```
stephan-hesselmann-by, line 104: Why did you add the type here? Shouldn't it just be `data` (without colon) to be consistent with other changes (and numpy docstring style)?

fjetter: yes, indeed. Sometimes my "regex search and replace" hammer was a bit too forceful, I assume. In particular with the cube API I had issues since the API is using a lot of type annotations in docs w/out having proper type annotations in general. Sometimes I fixed that, sometimes I didn't. TBH, I consider this PR one of the "it's not 100% but it still an improvement" things :)

If it is just about this line in particular, I can address this but I am affraid this or similar mistakes were made and I would prefer not to go over the code base again. This can be _eventually_ cleaned up since having the type documented twice should generally not be harmful

stephan-hesselmann-by: Alright no problem, I just wanted to know if this was an intentional change.

stephan-hesselmann-by (approved):

PR #415, opened 2021-02-17
Clarify type compatibiltiy

fjetter:
# Description:

Builds on top of #414 and adds more documentation about type compatibilities, known issues and a section about indexing in general


docs/spec/indexing.rst
```diff
@@ -0,0 +1,144 @@
...
+
+All indices implement :class:`~kartothek.core.index.IndexBase` which allows the user to interact with the indices in some useful ways.
+
+.. ipython:: python
+
+    from kartothek.core.index import IndexBase
+
+    index = IndexBase(column="FieldName", index_dct=index_dct)
+
+    index.dtype
```
jochen-ott-by, line 30: Either the output, a one-sentence explanation, or a link to the docs of the respective function would be helpful here.

docs/spec/indexing.rst
```diff
@@ -0,0 +1,144 @@
...
+Typing
+------
+
+Every index has a well defined arrow data type which is usually inferred automatically and ensured to be consistent with the overall dataset schema.
+
+.. ipython:: python
+
+    part_index.dtype
+
+Kartothek generally does not impose any restrictions on the data types to be used as long as they are compatible and in alignment with the :doc:`pyarrow pandas integration<python/pandas>`.
```
jochen-ott-by, line 115: It is true for *payload data* that kartothek essentially supports what arrow does, but for indices, it's a bit more complex, as there is string-to-value conversion involved, so there is actually a limited set supported by kartothek:
https://github.com/JDASoftwareGroup/kartothek/blob/5b239bd906b8e01fe1691a26745330b018814925/kartothek/core/index.py#L144
... which ends with
https://github.com/JDASoftwareGroup/kartothek/blob/5b239bd906b8e01fe1691a26745330b018814925/kartothek/core/index.py#L201

Therefore, for indices, I think this statement is misleading.

BTW: does the documentation  already contain the statement about what data types are supported for payload data? I could not find it ...

docs/spec/type_system.rst
```diff
@@ -244,6 +246,35 @@ We cannot treat timestamps for different timezones as the same time because the
 semantic meaning. We also cannot treat timestamps with different unit types as same since they all have very different
 ranges. So, no normalization is implemented for timestamps.
 
+.. note::
+
+  For compatibility reasons, `kartothek` coerces timestamps to `us` accuracy, effectively truncating the timestamp. If the timestamp actually has a higher accuracy, arrow raises an exception, rejecting it
```
jochen-ott-by, line 251: Compatibility with what? hive?

Also, a period or colon is missing at the end of this line.

docs/spec/indexing.rst
fjetter: these are a few examples and the index class is linked above with the full docs. I think this is as verbose as it needs to be

docs/spec/type_system.rst
fjetter: not hive. to be 100% honest, I do not know anymore. I believe this is connected to parquet itself and I merely know that it will/is possible with arrow 3.0
iirc, different java implementations are struggling with this.

I simply cannot put more information into this

docs/spec/indexing.rst
fjetter: > BTW: does the documentation already contain the statement about what data types are supported for payload data? I could not find it ...


The documentation includes a very long description about what types are used and how they are supported https://kartothek.readthedocs.io/en/stable/spec/type_system.html

If there is something missing and/or wrong, I'd rather deal with it by fixing it over there instead of documenting it again. Otherwise I moved this statement and put it into https://kartothek.readthedocs.io/en/stable/spec/serialization.html with a link to the type system page

jochen-ott-by (approved):

PR #416, opened 2021-02-17
Add reference parquet files for pyarrow>=1

fjetter:
# Description:

Follow up to #411 . I added a test which should alarm us if we forget about the reference file


stephan-hesselmann-by:
Are the reference files supposed to be empty or this a UI bug?

fjetter:
they are not supposed to be empty :) I'll check again (I'd expect the tests to detect this if they were)

fjetter:
I guess this is a UI thing
image

stephan-hesselmann-by (approved):

Issue #413, opened 2021-02-16
Incorrect indices error when updating cube


lr4d:
### Problem description
It looks like https://github.com/JDASoftwareGroup/kartothek/pull/398 has broken backwards compatibility for `kartothek.io.**_cube` users

Unit test stacktrace:
```
/tmp/venv/raq/raqbundle/core_data_snapshotter/build_datasets.py:686: in update_cube
    metadata=ktk_cube_metadata,
/tmp/venv/lib/python3.6/site-packages/kartothek/io/dask/bag_cube.py:489: in update_cube_from_bag
    df_serializer=df_serializer,
/tmp/venv/lib/python3.6/site-packages/kartothek/io/dask/common_cube.py:418: in append_to_cube_from_bag_internal
    cube = ensure_valid_cube_indices(existing_datasets, cube)
/tmp/venv/lib/python3.6/site-packages/kartothek/io/dask/common_cube.py:75: in ensure_valid_cube_indices
    compatible_indices = _ensure_compatible_indices(ds, table_indices)
/tmp/venv/lib/python3.6/site-packages/kartothek/io_components/utils.py:125: in _ensure_compatible_indices
    f"Incorrect indices provided for dataset.\n"
E   ValueError: Incorrect indices provided for dataset.
E   Expected: []
E   But got: {'XXX', 'YYY'}

```

fjetter:
Do you have a minimal example? this is otherwise impossible to reproduce

lr4d:
Of course, I'll post a minimal example today. I just wanted to create the issue already to link to it

stephan-hesselmann-by:
The issue occurs when updating a cube that has been created with multiple datasets. The problem is with `ensure_valid_cube_indices` - it mutates the required indices while depending on the unmutated value when looping over the datasets. It should be relatively straight forward to fix. I'll check whether there were more issues on Monday and provide a fix for this one.

PR #418, opened 2021-02-22
Fix: Cube index validation (#413)

stephan-hesselmann-by:
The issue occured when updating a cube with multiple datasets, which also have different dimension columns.

This fixes issue #413.

The bug is caused by access of a mutated variable - namely `required_indices` via `table_indices` - in the loop. I rewrote the loop to circumvent this problem and added a unit test which verifies that the index validation is working as expected.


fjetter (approved):
Thanks @stephan-hesselmann-by

Issue #325, opened 2020-07-28
Re-write predicate involving `in` operator to used disjunction of `==`


lr4d:
### Problem description

We use the `in` operator internally in predicate parsing, but we can just re-write the predicates to use a disjunction of `==` terms.
e.g.
`  [[('A', 'in', [1, 4, 9, 13])]] -> [[('A', '==', 1)], [('A', '==', 4)], [('A', '==', 9)], [('A', '==', 13)]]`

We could implement this re-write when a user passes predicates involving `in`, before the predicates are evaluated. This seems to be as fast as or faster than our current evaluation of predicates in micro-benchmarks (see below).

### Example code (ideally copy-pastable)
```python
import pyarrow as pa
import numpy as np
from tempfile import TemporaryDirectory
from storefact import get_store_from_url
from functools import partial
store_factory = partial(get_store_from_url, f"hfs://{TemporaryDirectory().name}")
dataset_uuid = "test"
import pandas as pd
df = pd.DataFrame({"A": range(10), "B": ["A", "B"] * 5, "C": [np.nan, *range(-10, -1)]})
from kartothek.io.eager import read_table, store_dataframes_as_dataset
dm = store_dataframes_as_dataset(
    store=store_factory, dataset_uuid=dataset_uuid, dfs=[df]*100, # partition_on=["A", "B"]
)

store_dataframes_as_dataset(
    store=store_factory, dataset_uuid="part", dfs=[df]*100, partition_on=["A", "B"]
)
from kartothek.io.eager import read_dataset_as_metapartitions
from kartothek.io_components.read import dispatch_metapartitions_from_factory

target = [1, 4, 9, 13]
predicates_in = [[("A", "in", target)]]
predicates_normal = [[("A", "==", n)] for n in target]

from kartothek.core.factory import DatasetFactory
f = DatasetFactory(dataset_uuid=dataset_uuid, store_factory=store_factory)
f_part = DatasetFactory(dataset_uuid="part", store_factory=store_factory)

%timeit dispatch_metapartitions_from_factory(f, predicates=predicates_in)
# 61 µs ± 9.64 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit dispatch_metapartitions_from_factory(f, predicates=predicates_normal)
# 50.7 µs ± 2.45 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

[...]

PR #423, opened 2021-02-24
normalize "in" predicate as disjunction of "=="

lr4d:
# Description:

The following two statements are equivalent:
- `x in [1,2,3]`
- `(x == 1) or (x == 2) or (x == 3)`

This approach simplifies the function by just calling itself instead of iterating over a for loop and essentially running the code block under `if op == "==":` for each value of an "in" predicate

- [ ] Closes #xxxx
- [ ] Changelog entry

### DIFF
--- a/kartothek/serialization/_parquet.py
+++ b/kartothek/serialization/_parquet.py
@@ -555,17 +555,16 @@ def _predicate_accepts(predicate, row_meta, arrow_schema, parquet_reader):
     elif op == ">":
         return max_value > val
     elif op == "in":
-        # This implementation is chosen for performance reasons. See
-        # https://github.com/JDASoftwareGroup/kartothek/pull/130 for more information/benchmarks.
-        # We accept the predicate if there is any value in the provided array which is equal to or between
-        # the parquet min and max statistics. Otherwise, it is rejected.
-        for x in val:
-            if pd.isnull(x):
-                if parquet_statistics.null_count > 0:
-                    return True
-            elif min_value <= x <= max_value:
-                return True
-        return False
+        # We don't duplicate the logic here, as "in" is just a disjunction of "=="'s
+        return any(
+            _predicate_accepts(
+                predicate=[col, "==", value],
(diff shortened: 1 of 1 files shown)


fjetter:
Does the `any` statement abort evaluation as soon as it finds the first value which evaluates to True or does it continue the evaluation? aborting early could vastly improve best case scenarios but I'm not sure about how the stdlib implementation deals with this (https://docs.python.org/3/library/functions.html#any doesn't tell us anything about it)

fjetter:
eventually we might want to consider passing the filters directly to pyarrow since they implemented this by now as well. I would expect them to deal with these things much faster than we are in python.

fjetter:
do we have any kind of benchmarks? What I'm a bit concerned about is that we interact with the parquet reader now for every element in the list. I don't know how expensive this is

lr4d:
> What I'm a bit concerned about is that we interact with the parquet reader now for every element in the list. I don't know how expensive this is

Very valid concern. We can isolate the operation evaluation logic in a separate function and call that directly so that we don't need to bother about this

fjetter:
cc @mlondschien this might interest you

xref https://github.com/JDASoftwareGroup/kartothek/issues/325

fjetter:
> Very valid concern. We can isolate the operation evaluation logic in a separate function and call that directly so that we don't need to bother about this

All depends on how complex the end state will be since your intention is to simplify things. I think this would be ok, though

mlondschien:
Thanks @fjetter for pinging me. This does not seem to explode the complexity exponentially as we're normalizing for a single column only. So I don't think the concerns mentioned in #325 would apply here.

lr4d:
> eventually we might want to consider passing the filters directly to pyarrow since they implemented this by now as well. I would expect them to deal with these things much faster than we are in python.

Good point. Do you know how "mature" this logic for pyarrow is atm? It might make more sense to invest in using their functionality directly rather than this kind of work

PR #424, opened 2021-02-25
Fix read_dataset_as_dataframes(_iterator) to correctly return categoricals for misaligned categories

### LINKED ISSUE #420: read_dataset_as_dataframes(_iterator) does not return categoricals as requested
See below:
```
In [1]: import pandas as pd
   ...: import storefact
   ...: from functools import partial
   ...: from kartothek.io.eager import store_dataframes_as_dataset, read_dataset_as_dataframes                                                                                                                                                                                               ...:
   ...: path = "/tmp"
   ...: uuid = "dataset"
   ...: df = pd.DataFrame({"a": [0, 0, 1, 1], "b": ["a", "a", "a", "b"], "c": [0, 0, 0, 0]})
   ...: store = partial(storefact.get_store_from_url, f"hfs://{path}?create_if_missing=False")
   ...: dm = store_dataframes_as_dataset(
   ...:     dfs=[df],
   ...:     dataset_uuid=uuid,
   ...:     store=store,
   ...:     partition_on=["a", "c"],
   ...:     overwrite=True,
   ...: )
   ...:
   ...: df = read_dataset_as_dataframes(
   ...:     dataset_uuid=uuid, store=store, dispatch_by=["c"], categoricals={"table": ["b"]}
   ...: )
   ...: df[0]["table"].dtypes
Out[1]:
a     int64
c     int64
b    object
dtype: object
```

To my understanding, this is due to a missing `align_categories` here: 
https://github.com/JDASoftwareGroup/kartothek/blob/4008de4436dde947faa3979f3ccf035f99c955c0/kartothek/io_components/metapartition.py#L1564

<details>

```
import pandas as pd
import storefact
from functools import partial
from kartothek.io.eager import store_dataframes_as_dataset, read_dataset_as_dataframes

path = "/tmp"
uuid = "dataset"
df = pd.DataFrame({"a": [0, 0, 1, 1], "b": ["a", "a", "a", "b"], "c": [0, 0, 0, 0]})
store = partial(storefact.get_store_from_url, f"hfs://{path}?create_if_missing=False")
    dfs=[df],
    dataset_uuid=uuid,
    store=store,
    partition_on=["a", "c"],
    overwrite=True,
)

df = read_dataset_as_dataframes(
    dataset_uuid=uuid, store=store, dispatch_by=["c"], categoricals={"table": ["b"]}
)
df[0]["table"].dtypes
```

</details>

mlondschien:
Closes #420 

I did not find a cleaner way to get the list of categoricals. Another option would be to do this within `align_categoricals` is `categoricals=None`.

I am not sure whether the
```python
    if len(categoricals) == 0:
        return dfs
```
is required to prevent additional copies in absence of categoricals.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+Version 3.19.2 (2021-xx-xx)
+===========================
+
+* :func:`~kartothek.io.eager.read_dataset_as_dataframes` and
+  :func:`~kartothek.io.iter.read_dataset_as_dataframes__iterator` now correctly return
+  categoricals as requested for misaligned categories.
+
+
 Version 3.19.1 (2021-02-24)
 ===========================
 
--- a/kartothek/io_components/metapartition.py
+++ b/kartothek/io_components/metapartition.py
@@ -51,6 +51,7 @@ from kartothek.core.uuid import gen_uuid
 from kartothek.io_components.utils import (
     InferredIndices,
     _ensure_valid_indices,
+    align_categories,
     combine_metadata,
 )
 from kartothek.serialization import (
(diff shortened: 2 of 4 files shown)


fjetter (approved):
I would think that the `len(categoricals)==0` guard is not required but in terms of code complexity, it is straight forward so I think it is perfectly fine to keep. Thanks for fixing this @mlondschien !

PR #426, opened 2021-03-02
Pyugrade

fjetter:
# Description:

Stumbled upon this in the pandas pre-commit hooks. I think this would've been incredibly useful during (or right after) the py3 migration but it still serves some value to counter deprecated behaviours and/or replaces certain expressions with more modern ones.
I like it but I don't have a strong opinion here.

see https://github.com/asottile/pyupgrade

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,6 +15,13 @@ repos:
         args:
           - --target-version=py36
 
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v2.10.0
+    hooks:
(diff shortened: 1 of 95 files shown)


lr4d:
Very nice

lr4d (approved):

lr4d:
Now we just need to find a pre-commit hook which fixes grammar errors in variables and comments

fjetter:
I'll rerun after #431

xhochy:
@fjetter Can you rerun/rebase? (otherwise happy to take care of that)

PR #429, opened 2021-03-09
[DRAFT] support renaming of datasets while copying a cube

MartinHaffner:
# Description:

Function copy_cube() has a new parameter: datasets_to_rename:

    datasets_to_rename: Optional[Dict[str, str]]
        Optional dict of {old dataset name:  new dataset name} entries. If specified,
        the corresponding datasets will be renamed accordingly. Unknown keys will cause
        an error.

Renaming will also affect the dataset metadata.

### DIFF
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -888,6 +888,28 @@ class DatasetMetadataBuilder(CopyMixin):
         ds_builder.tables = dataset.tables
         return ds_builder
 
+    def modify_dataset_name(self, new_ds_name):
+        """
+        Modify the dataset name and change dependent fields, i.e. UUID and partitioning
+        information.
+
+        TODO: This will atm only work for cube datasets with a UUID like cube++dataset.
+        Which UUIDs are possible for non-cube datasets?
+        """
+        rx_transform_path = re.compile(r"(\w*)\+\+(\w*)/(.*)")
+
+        cube_name, old_ds_name = self.uuid.split("++")
+        self.uuid = cube_name + "++" + new_ds_name
+        for p_key, p in self.partitions.items():
+            pdict = p.to_dict()
+            for table_name, file_name in pdict["files"].items():
+                match = rx_transform_path.match(file_name)
+                if match:
+                    f_cube, f_ds, f_path = match.groups()
+                    pdict["files"][table_name] = f"{f_cube}++{new_ds_name}/{f_path}"
+            self.partitions[p_key] = Partition.from_dict(p_key, pdict)
+        return self
+
     def add_partition(self, name, partition):
         """
         Add an (embedded) Partition.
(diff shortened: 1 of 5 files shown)


kartothek/core/dataset.py
```diff
@@ -888,6 +888,28 @@ def from_dataset(dataset):
...
 
+    def modify_dataset_name(self, new_ds_name):
+        """
+        Modify the dataset name and change dependent fields, i.e. UUID and partitioning
+        information.
+
+        TODO: This will atm only work for cube datasets with a UUID like cube++dataset.
+        Which UUIDs are possible for non-cube datasets?
+        """
+        rx_transform_path = re.compile(r"(\w*)\+\+(\w*)/(.*)")
```
fjetter, line 899: this class should not make *any* assumptions about the UUID. Everything that is allowed by https://github.com/JDASoftwareGroup/kartothek/blob/8848fd43639ea10785a7fe87833a7b6dc687850e/kartothek/core/dataset.py#L36-L37 can be a UUID and this class should handle it this way

fjetter: Not accounting for performance at this moment, you can use https://github.com/JDASoftwareGroup/kartothek/blob/8848fd43639ea10785a7fe87833a7b6dc687850e/kartothek/core/urlencode.py#L25-L39 to split a given key into it's components

kartothek/core/dataset.py
```diff
@@ -888,6 +888,28 @@ def from_dataset(dataset):
...
+
+        cube_name, old_ds_name = self.uuid.split("++")
+        self.uuid = cube_name + "++" + new_ds_name
+        for p_key, p in self.partitions.items():
+            pdict = p.to_dict()
+            for table_name, file_name in pdict["files"].items():
+                match = rx_transform_path.match(file_name)
+                if match:
+                    f_cube, f_ds, f_path = match.groups()
+                    pdict["files"][table_name] = f"{f_cube}++{new_ds_name}/{f_path}"
```
fjetter, line 909: Do not encode the `++` at this level. This method should not even know what a cube is

fjetter (changes requested):
I would suggest to start with a `rename_dataset` or `copy_dataset` method which is not coupled to cubes at all. this way we could write

```python
def copy_cube(...):
    for dataset in cube:
        rename_dataset(...)
```

kartothek/utils/store.py
```diff
@@ -181,12 +187,69 @@ def _copy_azure_cc(keys, src_store, tgt_store):
                 )
 
 
-def _copy_naive(keys, src_store, tgt_store):
-    for k in keys:
-        tgt_store.put(k, src_store.get(k))
+def _copy_naive(key_mappings, src_store, tgt_store, mapped_metadata=None):
```
fjetter, line 190: Please use https://github.com/JDASoftwareGroup/kartothek/blob/8848fd43639ea10785a7fe87833a7b6dc687850e/kartothek/utils/store.py#L189 which includes some specialised code for copy operations. For instance, it uses the copy api from azure directly which is much more efficient since it safes network traffic.

PR #431, opened 2021-03-10
Remove multi table feature

### LINKED ISSUE #427: Removal of complex input types
For mostly historic reasons, kartothek allows input in the form of 

```python
input_data = [
    [
        {
            "data": Dict[str, pandas.DataFrame],
            "secondary_indices": Optional[Dict[str, List[str]]],
            "label": Optional[str]
        },
    ]
] 
```

which allows users a range of features we mostly would like to get rid off
* Multi table writes (deprecated feature in general)
* user calcualted secondary indices (replaced with `secondary_indices` kwarg)
* Explicit labeling of output partitions (should be rather done with partition indices, if needed)

Removal of this in `parse_input_to_metapartition` will allow for significant simplifications in the code base.

There are a few non-trivial consequences by this. In particular this will render pipelines like `merge_datasets_as_delayed` useless since there is no explicit handle for the labels anymore. I firmly believe we should abandon this approach and replace it with a proper join implementation based on indices, which should be a few orders faster anyhow.

@mlondschien you have been involved with these kind of things and I'd like to get some input here. In particular, if we keep the align_dataset and merge pipelines, I would like to know what and how it is actually used for. All workflows I can think of in this area would strongly benefit from proper indexing and/or plain dask usage.

fjetter:
# Description:

As the title state, this will remove any API references to the dataset internal multi table functionality. It is effectively no longer possible to write these.

It goes without saying. This change is in terms of API as breaking as it gets. This is a new major release and I intend to drop many of the deprecated features after merging this as well (effectively everything which raises a warning at the moment)

- [x] Closes #427
- [x] Changelog entry


mlondschien:
Not sure how closes this is to merging, but I thought I would let you know what this breaks for us:

 - `read_table` no longer accepts `table` as a keyword argument.
 - `DatasetMetadata.table_meta["table"]` and `DatasetFactory.table_meta["table"]` raise a `TypeError: an integer is required`. IIUC this is superseeded by `.schema`.
 - `read_dataset_as_dataframes__iterator` no longer takes a `columns = {"table": list-like}` argument and no longer returns `{"table": df}` data frames.

Are you planning on making a deprecating release in between 3.19 and 4.0? The `table` kwarg is no problem since `"table"` is already a default value. However it would be nice to have `schema` available and be able to pass `columns=list-like` to get a data frame before 4.0 or be able to use the current API until 4.1.

fjetter:
If the removal of the `table` is actually a problem we can keep it without effect but I would prefer the clean removal since we've been dragging unused arguments around for years.

All the changes you mentioned were anticipated. We can make the table_meta change backwards compatible. I actually intended to remove it altogether but I don't have a strong opinion on this.

Regarding the columns input, this is actually not that easy to achieve and I'm not sure if it is worth the trouble. The problem here is (and also the reason why this must go) that all pipelines are implementing this somewhat differently

bottom line is that all of these changes feel like breaking but easily dealt with and I'm hesitant to introduce a lot of glue code to ease this migration. I believe these problems are quickly addressed by going over the code base once everywhere.

> DatasetMetadata.table_meta["table"] and DatasetFactory.table_meta["table"] raise a TypeError: an integer is required.

This, I will fix. Cryptic error message are always a problem, even for intended changes.

mlondschien:
Would it be an option to add

```python
@property
def schema(self):
    return self.table_meta["table"]
```

to `DatasetMetadata` and `DatasetFactory` before 4.0?

I guess we will be able to live with the change in `read_dataset_as_dataframes__iterator` API :).

fjetter:
yes, we can add the schema pre 4.X. We have one or two unreleased bugfixes on master which probable makes sense to release anyhow before that

kartothek/core/dataset.py
```diff
@@ -133,13 +117,19 @@ def primary_indices_loaded(self) -> bool:
         return True
 
     @property
-    def tables(self) -> List[str]:
-        if self.table_meta:
-            return list(self.table_meta.keys())
+    def table_name(self) -> str:
+        if self._table_name:
+            return self._table_name
```
bjoern-meier-by, line 122: `self._table_name` is set in __init__ and with a default value which evaluates to True.
Is it really required to keep the 3 cases, I would suggest to drop them.

kartothek/core/dataset.py
```diff
@@ -234,6 +223,42 @@ def to_json(self) -> bytes:
...
+
+        indices = _construct_dynamic_index_from_partitions(
+            partitions=self.partitions,
+            schema=self.schema,
+            default_dtype=pa.string() if self.metadata_version == 3 else None,
+            partition_keys=self.partition_keys,
+        )
+        combined_indices = self.indices.copy()
+        combined_indices.update(indices)
+        return self.copy(indices=combined_indices)
```
bjoern-meier-by, line 260: Maybe I do miss it , but as I currently understand if `self.primary_indices_loaded` is True we return the object itself. If not we load the indices and return a copy. So `self.primary_indices_loaded` is not updated from what I see, so I wonder why not?
Why even the copy?

kartothek/core/dataset.py
```diff
@@ -855,6 +851,7 @@ def create_partition_key(
 
         returns 'my-uuid/testtable/index1=value1/index2=value2/data'
     """
+    # TODO: ensure private
```
bjoern-meier-by, line 854: Still to be done in this PR?

kartothek/io_components/write.py
```diff
@@ -97,49 +88,39 @@ def persist_indices(
...
+    update_dataset: Optional[DatasetFactory],
+    store: KeyValueStore,
+    dataset_uuid: str,
+    table_name: str,
+):
+
+    if not schemas:
+        return None
+    schemas_set = set(schemas)
+    del schemas
```
bjoern-meier-by, line 102: Why this delete?

fjetter: mypy doesn't like to assign the _set_ schemas to a variable which is already typed as a _list_. I introduced another variable and are deleting the old one to ensure it is not used by mistake in the code below. at least in my experience this is a fairly common issue with these kinds of refactoring

kartothek/core/dataset.py
fjetter: I'll remove the TODO. What's public and what isn't was effectively taken care of in #435

fjetter: - [ ] remove todo

kartothek/core/dataset.py
fjetter: The `copy` methods were introduced at the very beginning of `kartothek` and we perceived it as a "clean coding" by not mutating any objects inplace but always return a new instance. This is what the `copy` methods are doing. By now I consider this unnecessary boilerplate but I didn't want to break any patterns here. Therefore if nothing is to be done, the same instance is returned, otherwise a new instance with the mutated state is returned

FYI this method is unchanged but was moved for some reason. I'll move it back to have cleaner diffs / blames. still valid question, of course :)

kartothek/core/dataset.py
fjetter: While the default is a str, it is actually possible to instantiate the class with `None` as a value. 

```python
In [1]: from kartothek.core.dataset import DatasetMetadata

In [2]: ds = DatasetMetadata("uuid")

In [3]: ds.table_name
Out[3]: 'table'

In [4]: ds = DatasetMetadata("uuid", table_name=None)

In [5]: ds
Out[5]: DatasetMetadata(uuid=uuid, table_name=<Unknown Table>, partition_keys=[], metadata_version=4, indices=[], explicit_partitions=True)

In [6]: ds.table_name
Out[6]: '<Unknown Table>'
```

This can happen in the edge case where one tries to read an empty dataset. The problem arises because we do not persist the tables as part of the JSON dataset metadata but is usually inferred by parsing the keys of the partitions. The edge case of an entirely empty dataset (i.e. no files) was not anticipated back then and we have multiple places in the test suite which test explicitly this case of not having any files. I do not recall the actual application for this, though. In favour of not breaking too much I would keep this as is, for now.

kartothek/core/dataset.py
bjoern-meier-by: Ah okay got it. I understand the arguments of preventing mutation but also that not everything can and should be folloewed 100%  and here I might argue against it but as it is pre-change code I am totally fine :)

bjoern-meier-by (approved):

bjoern-meier-by:
Thanks @fjetter for the cleanup and refactoring. 🙇

fjetter:
@bjoern-meier-by thanks for reviewing!

PR #434, opened 2021-03-11
Remove deprecated kwargs

fjetter:
# Description:

Depends on #431 and removes all deprecated arguments which raised `DeprecationWarnings`


fjetter:
will merge after #431

bjoern-meier-by (approved):

PR #435, opened 2021-03-11
Redefine public API

fjetter:
This is a proposal on how to make it more explicit on what we consider public and what not. After this change we will instruct all users to use the imports via `kartothek.api.*` and we will consider everything else as private and potential subject to change.
I did not introduce any noticable structure under `api` since we've had a lot of attempts but never really a successful one. I consider the separation between `dataset` and `cube` still relevant since they sometimes observe different behaviour and work on different entities. otherwise I think it should remain rather flat, opinions welcome


kartothek/api/utils.py
```diff
@@ -0,0 +1 @@
+from kartothek.core.utils import *  # noqa: F401, F403
```
stephan-hesselmann-by, line 1: Do we really need `api.utils`? This is now a subset of `api.dataset`, isn't it?

fjetter: good point

docs/versioning.rst
```diff
@@ -9,38 +9,22 @@ Specification
...
-on an individual case basis if an increase of major version is justified to
-signal the user potential breakage.
-Under any circumstances, breaking interface changes will be documented to the
-best of our knowledge.
+We are aware that the API not only consists of the signatures of functions and
+types of objects exposed by a library but as well by its behaviour. Many changes
+involve judgement calls about the compatibility of a given change and it might
+be possible that we'll introduce behavioural changes which you would not
+perceive compatible with semantic versioning. We will adhere as best as possible
+to this convention. In and case we intend to document all intended changes,
```
stephan-hesselmann-by, line 29: `In any case ...`

stephan-hesselmann-by (changes requested):
Thanks for taking this over, the changes are in line with what I had in mind. But I think we should also update (or maybe remove completely?) `docs/api.rst`.

stephan-hesselmann-by (approved):

PR #443, opened 2021-03-17
Lml 98/support renaming

MartinHaffner:
# Description:

- implements eager copying of datasets, with optional renaming

### DIFF
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -907,15 +907,15 @@ class DatasetMetadataBuilder(CopyMixin):
         ds_builder.partitions = dataset.partitions
         return ds_builder
 
-    def modify_uuid(self, tgt_uuid):
+    def modify_uuid(self, tgt_uuid: str):
         """
         Modify the dataset uuid and depending metadata:
         - paths to partitioning files
         - path to index files
 
-        Parameters:
-            tgt_uuid: str
-                Modified dataset UUID.
+        :param tgt_uuid:
+            Modified dataset UUID.
+        :return modified builder object
         """
 
         # modify file names in partition metadata
@@ -930,9 +930,8 @@ class DatasetMetadataBuilder(CopyMixin):
 
         for i_key, i in self.indices.items():
             if hasattr(i, "index_storage_key"):
-                i.index_storage_key = i.index_storage_key.replace(
-                    self.uuid, tgt_uuid, 1
-                )
+                i_mod = getattr(i, "index_storage_key").replace(self.uuid, tgt_uuid, 1)
+                i.index_storage_key = i_mod  # type: ignore
         self.uuid = tgt_uuid
         return self
 
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -930,8 +930,11 @@ class DatasetMetadataBuilder(CopyMixin):
 
         for i_key, i in self.indices.items():
             if hasattr(i, "index_storage_key"):
-                i_mod = getattr(i, "index_storage_key").replace(self.uuid, tgt_uuid, 1)
-                i.index_storage_key = i_mod  # type: ignore
+                setattr(
+                    i,
+                    "index_storage_key",
+                    getattr(i, "index_storage_key").replace(self.uuid, tgt_uuid, 1),
+                )
         self.uuid = tgt_uuid
         return self
 
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -907,26 +907,33 @@ class DatasetMetadataBuilder(CopyMixin):
         ds_builder.partitions = dataset.partitions
         return ds_builder
 
-    def modify_dataset_name(self, new_ds_name):
+    def modify_uuid(self, tgt_uuid):
         """
-        Modify the dataset name and change dependent fields, i.e. UUID and partitioning
-        information.
+        Modify the dataset uuid and depending metadata:
+        - paths to partitioning files
+        - path to index files
 
-        TODO: This will atm only work for cube datasets with a UUID like cube++dataset.
-        Which UUIDs are possible for non-cube datasets?
+        Parameters:
+            tgt_uuid: str
+                Modified dataset UUID.
         """
-        rx_transform_path = re.compile(r"(\w*)\+\+(\w*)/(.*)")
 
-        cube_name, old_ds_name = self.uuid.split("++")
-        self.uuid = cube_name + "++" + new_ds_name
+        # modify file names in partition metadata
         for p_key, p in self.partitions.items():
             pdict = p.to_dict()
-            for table_name, file_name in pdict["files"].items():
-                match = rx_transform_path.match(file_name)
-                if match:
-                    f_cube, f_ds, f_path = match.groups()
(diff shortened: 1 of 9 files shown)


kartothek/core/dataset.py
```diff
@@ -907,6 +907,37 @@ def from_dataset(dataset):
         ds_builder.partitions = dataset.partitions
         return ds_builder
 
+    def modify_uuid(self, tgt_uuid: str):
+        """
+        Modify the dataset uuid and depending metadata:
+        - paths to partitioning files
+        - path to index files
+
+        :param tgt_uuid:
```
stephan-hesselmann-by, line 916: The rest of the code bases uses numpy style docstrings. Can you change them to be consistent?

kartothek/core/dataset.py
```diff
@@ -907,6 +907,37 @@ def from_dataset(dataset):
...
+            Modified dataset UUID.
+        :return modified builder object
+        """
+
+        # modify file names in partition metadata
```
stephan-hesselmann-by, line 921: This method is not atomic, i.e. if it fails in the middle of operation it leaves the partitions half modified. That seems a bit fragile to me, perhaps we could build the new partitions and then switch them at the end rather than modifying in place in a loop?

Also I think this method should probably be unit tested.

kartothek/io/eager.py
```diff
@@ -738,3 +742,60 @@ def garbage_collect_dataset(dataset_uuid=None, store=None, factory=None):
...
+        Target Store. May be the same as store, if src_dataset_uuid and
+        target_dataset_uuid are different. If empty, value from parameter store is
+        used
+    """
+    if target_dataset_uuid is None:
+        target_dataset_uuid = src_dataset_uuid
+    if tgt_store is None:
+        tgt_store = store
+
+    if (src_dataset_uuid == target_dataset_uuid) & (store == tgt_store):
```
stephan-hesselmann-by, line 774: Shouldn't this just be a no op rather than a ValueError?

kartothek/core/dataset.py
MartinHaffner: OK, changed the comment style. (Not sure if the doctrings I've seen so far were consistent throughout the whole project, btw...)

kartothek/io/eager.py
MartinHaffner: Both are fine for me. However, in similar situations the KTK API raises a ValueError, e.g. https://github.com/JDASoftwareGroup/kartothek/blob/08a809429cba9f46c4c0d09853eb476021764708/kartothek/io/eager.py#L362.
So leaving the error here as is would be at least consistent with the rest of the API.

kartothek/core/dataset.py
MartinHaffner: atomicity: OK, changed this. I don't think btw that non-atomicity would be a problem here, since this is a Metadata**Builder** object with a very limited life span, not a long-living Metadata object. But it's not a big change anyway.

unit tests: ok, going to add this. The method is already included in the higher-level unit tests, but dedicated tests for this method make of course sense.

kartothek/core/dataset.py
stephan-hesselmann-by: They are not consistent unfortunately, but at least the intention is to stick with numpy...

kartothek/core/dataset.py
```diff
@@ -907,6 +907,37 @@ def from_dataset(dataset):
         ds_builder.partitions = dataset.partitions
         return ds_builder
 
+    def modify_uuid(self, tgt_uuid: str):
```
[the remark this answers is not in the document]
MartinHaffner: Fine for me :-) (will commit it with the rest of the changes)

kartothek/io/eager.py
```diff
@@ -738,3 +742,60 @@ def garbage_collect_dataset(dataset_uuid=None, store=None, factory=None):
     # Given that `nested_files` is a generator with a single element, just
     # return the output of `delete_files` on that element.
     return delete_files(next(nested_files), store_factory=ds_factory.store_factory)
+
+
+def copy_dataset(
+    src_dataset_uuid: str,
```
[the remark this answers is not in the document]
MartinHaffner: Done.

kartothek/io/eager.py
```diff
@@ -738,3 +742,60 @@ def garbage_collect_dataset(dataset_uuid=None, store=None, factory=None):
...
+    store: KeyValueStore,
+    target_dataset_uuid: Optional[str] = None,
+    tgt_store: Optional[KeyValueStore] = None,
```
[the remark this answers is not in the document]
MartinHaffner: Done.

kartothek/io/eager_cube.py
```diff
@@ -445,14 +542,32 @@ def copy_cube(cube, src_store, tgt_store, overwrite=False, datasets=None):
...
     )
 
+    # get the keys to copy for the cube
     keys = get_copy_keys(
         cube=cube,
         src_store=src_store,
         tgt_store=tgt_store,
         overwrite=overwrite,
         datasets=datasets,
     )
```
bjoern-meier-by, line 552: Can getting the keys be part of the if clause?

kartothek/io/eager_cube.py
```diff
@@ -419,7 +421,95 @@ def delete_cube(cube, store, datasets=None):
...
+        )
+        md_ds_key = f"{cube_prefix}{KTK_CUBE_UUID_SEPARATOR}{ds}{METADATA_BASE_SUFFIX}{METADATA_FORMAT_JSON}"
+        md_transformed[md_ds_key] = md_ds_transformed
+    return md_transformed
+
+
+def copy_cube(
+    cube: Cube,
+    src_store: Union[KeyValueStore, Callable[[], KeyValueStore]],
+    tgt_store: Union[KeyValueStore, Callable[[], KeyValueStore]],
```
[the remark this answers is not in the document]
MartinHaffner: Done.

kartothek/io/eager_cube.py
MartinHaffner: No: The original keys are needed in both cases, copy without rename (if clause) and copy with rename (else clause).

kartothek/io/eager_cube.py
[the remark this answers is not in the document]
MartinHaffner: On second thought: undone again. This function did already exist with these parameter names. Adopting them to the other schema (though better) would be a breaking change. Sure that we want to do this?

kartothek/io/eager_cube.py
bjoern-meier-by: Args of course. You are right, keep it.

bjoern-meier-by (approved):

stephan-hesselmann-by (approved):
Can you please add a Changelog entry?

stephan-hesselmann-by:
@MartinHaffner can you please get this into a mergable state by resolving the conflicts and getting a green build?

NeroCorleone:
Could we close this one here in favor of https://github.com/JDASoftwareGroup/kartothek/pull/450? Conflicts have been resolved a tiny little thing with the docs has been fixed, additionally.

fjetter:
Closed in favour of #450

PR #450, opened 2021-04-07
Add eager copy for datasets

NeroCorleone:
# Description:

This PR contains work that has already been done [here](https://github.com/JDASoftwareGroup/kartothek/pull/443) plus minor things to get it into a mergeable state. Opening a new one because @MartinHaffner is on vacation and I unfortunately don't have access rights on his fork.

We implement the eager copying of ktk datasets and offer the option to rename copied ktk cubes.

### DIFF
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -938,13 +938,12 @@ class DatasetMetadataBuilder(CopyMixin):
         self.partitions = modified_partitions
 
         for i_key, i in self.indices.items():
-            if (hasattr(i, "index_storage_key")) and (
-                getattr(i, "index_storage_key") is not None
+            if (
+                isinstance(i, ExplicitSecondaryIndex)
+                and i.index_storage_key is not None
             ):
-                setattr(
-                    i,
-                    "index_storage_key",
-                    getattr(i, "index_storage_key").replace(self.uuid, target_uuid, 1),
+                i.index_storage_key = i.index_storage_key.replace(
+                    self.uuid, target_uuid, 1
                 )
         self.uuid = target_uuid
         return self
--- a/kartothek/core/dataset.py
+++ b/kartothek/core/dataset.py
@@ -907,15 +907,15 @@ class DatasetMetadataBuilder(CopyMixin):
         ds_builder.partitions = dataset.partitions
         return ds_builder
 
-    def modify_uuid(self, tgt_uuid):
+    def modify_uuid(self, tgt_uuid: str):
         """
         Modify the dataset uuid and depending metadata:
         - paths to partitioning files
         - path to index files
 
-        Parameters:
-            tgt_uuid: str
-                Modified dataset UUID.
+        :param tgt_uuid:
+            Modified dataset UUID.
+        :return modified builder object
         """
 
         # modify file names in partition metadata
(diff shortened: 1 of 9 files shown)


NeroCorleone:
Looks like the one failing CI build is also on master atm.

bjoern-meier-by (approved):

kartothek/core/dataset.py
```diff
@@ -907,6 +907,48 @@ def from_dataset(dataset):
...
+
+        for i_key, i in self.indices.items():
+            if (hasattr(i, "index_storage_key")) and (
+                getattr(i, "index_storage_key") is not None
+            ):
+                setattr(
+                    i,
+                    "index_storage_key",
+                    getattr(i, "index_storage_key").replace(self.uuid, target_uuid, 1),
+                )
```
fjetter, line 948: Instead of checking for the attribute, I would suggest to check for the type. imo, there is no need for duck-typing here

kartothek/io/eager_cube.py
```diff
@@ -419,7 +421,95 @@ def delete_cube(cube, store, datasets=None):
...
+        else:
+            tgt_uuid = f"{tgt_cube_prefix}{KTK_CUBE_UUID_SEPARATOR}{ds}"
+
+        # transform metadata for this dataset
+        md_ds_transformed = (
+            DatasetMetadataBuilder.from_dataset(ds_meta)
+            .modify_uuid(tgt_uuid)
+            .to_json()[1]
+        )
+        md_ds_key = f"{cube_prefix}{KTK_CUBE_UUID_SEPARATOR}{ds}{METADATA_BASE_SUFFIX}{METADATA_FORMAT_JSON}"
```
fjetter, line 499: This looks suspicious. At this place, there should not be the need to specify a `KTK_CUBE_UUID_SEPARATOR`. The cube separator is part of the dataset UUID and I would not introduce a special treatment like this here

fjetter:
I was raising this issue in the past already but I would propose to implement this as a `rename_dataset` function which takes care of a generic dataset and implement the cube operation as something like

```python
try:
    for dataset in cube:
        rename_dataset(dataset)
except:
    rollback_cube_transaction()
```

instead of implementing this directly on the cube

kartothek/io/eager_cube.py
NeroCorleone: Fixed.

NeroCorleone:
> I was raising this issue in the past already but I would propose to implement this as a `rename_dataset` function which takes care of a generic dataset and implement the cube operation as something like
> 
> ```python
> try:
>     for dataset in cube:
>         rename_dataset(dataset)
> except:
>     rollback_cube_transaction()
> ```
> 
> instead of implementing this directly on the cube

I think this makes sense: I implemented a naive roll back by tracking the already copied datasets and deleting those from the target store, if something goes wrong during copy. Not sure, if this is the best way to do it, though. Would be happy to hear opinions, here.

fjetter:
> I think this makes sense: I implemented a naive roll back by tracking the already copied datasets

I'm not even sure if we absolutely need the rollback. If we do, I would reuse existing functionality. Have a look at https://github.com/JDASoftwareGroup/kartothek/blob/b349ceed6a27539110860b67b6f1f2285436d9b2/kartothek/io_components/cube/write.py#L409-L460

NeroCorleone:
What do you think? Is there anything missing to get this merged?

Issue #445, opened 2021-03-31
kartothek 4.0 update_dataset_from_ddf corrupts datasets if a different table name is used


peter-hoffmann-by:
### Problem description

Kartothek 4.0 breaks updating existing datasets that use a different table name then the default.

### Expected Behaviour

the table name keyword is respected in updating datasets. 

### Example code 

```python
import dask
dask.config.set(scheduler='synchronous')
import dask.dataframe as dd

ddf = dask.datasets.timeseries()
ddf = ddf.reset_index()
print(ddf.head())
ddf['date'] = ddf.timestamp.dt.date
ddf = ddf.loc[:10]
store_url = f"hfs://testdata"

store = get_store_from_url(store_url)
delayed = update_dataset_from_ddf(ddf, store, "testdata", table='predictions', partition_on=['date'], shuffle=True, num_buckets=5)
res = delayed.compute()
```

Output:
 
```
$ tree testdata/testdata 
testdata/testdata 
└── table 
    ├── _common_metadata 
    ├── date=2000-01-01 
    │   ├── 414bb87213014db0b5d34a68e4a03edb.parquet 
    │   ├── 76ab5551c3dc4929b2676b04f2f972ac.parquet
    │   ├── b0614814e3814aecb76705a0f8e5b238.parquet 
    │   ├── eb430028a14644ac845c036b619fff47.parquet 
    │   └── f886ae72fa824827ad850f4af941b8b0.parquet 
```

Expected Output:

```
testdata/testdata 
  └── predictions 
```

The example above is just a minimal example to show that the the table name is not used in update_dataset_from_ddf. The corruption of dataset occurs if you update an existing (pre 4.0 kartothek dataset) with the new release and you suddenly  have two table name in your dataset ('predictions' and 'table').

```
import dask
import pandas as pd
import dask.dataframe as dd
from datetime import date
from storefact import get_store_from_url
from kartothek.io.dask.dataframe import update_dataset_from_ddf, read_dataset_as_ddf

dask.config.set(scheduler='synchronous')
store_url = f"hfs://testdata"
store = get_store_from_url(store_url)


def create():
    df = pd.DataFrame({"date": [date(2021,1,x) for x in range(1,6)], 'value': range(5)})
    ddf = dask.dataframe.from_pandas(df, npartitions=1)
    delayed = update_dataset_from_ddf(ddf, store, dataset_uuid, table='predictions', partition_on=['date'])
    res = delayed.compute()

def update():
[...]

PR #451, opened 2021-04-08
Fix dataset corruption after updates (#445)

stephan-hesselmann-by:
When updating a dataset with a table name other than 'table', an additional table named
'table' is erroneously created. This corrupts the dataset. The issue was introduced after
deprecating the table name feature in the 4.0.0 release. The root cause is not passing the
table name as an argument within `partition_on` and `add_metapartition`, which leads to the
default table name "table" being used.

# Description:

- [ ] Closes #xxxx
- [ ] Changelog entry

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Kartothek 4.0.1 (2021-04-XX)
+============================
+
+* Fixed dataset corruption after updates when table names other than "table" are used (#445).
+
+
 Kartothek 4.0.0 (2021-03-17)
 ============================
 
--- a/kartothek/io/dask/delayed.py
+++ b/kartothek/io/dask/delayed.py
@@ -264,6 +264,7 @@ def update_dataset_from_delayed(
     sort_partitions_by=None,
     secondary_indices=None,
     factory=None,
+    table_name=SINGLE_TABLE,
 ):
     """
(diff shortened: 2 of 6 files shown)


bjoern-meier-by (approved):

fjetter:
Can we put an end-to-end test in, as well? create a dataset w/ custom table name, update it, ensure it is not corrupt? from this unit test it is not immediately clear that this fixes the problem

hoffmann (approved):

stephan-hesselmann-by:
I believe the issue described in #445 should be fixed with these changes (scenario where the table name is non-default but equal for create and update). However, I foresee problems when the table name diverges between creation and update, i.e.
```python
    # Create with table name "predictions"
    delayed = update_dataset_from_ddf(ddf, store, dataset_uuid, table='predictions', partition_on=['date'])

    # Create with default table name
    delayed = update_dataset_from_ddf(ddf, store, dataset_uuid, partition_on=['date'])
```

What whould the expected behavior be in such a cause? Should the table name be inferred as "predictions"? I believe that will require more modifications to the code.

stehessel:
> What whould the expected behavior be in such a case? Should the table name be inferred as "predictions"? I believe that will require more modifications to the code.

I will create a follow up issue for that. Merging this for now as the build failure is already on master and unrelated to these changes.

PR #456, opened 2021-04-16
Testing code moved from kartothek.io.testing to tests.*

MartinHaffner:
Description:

The module kartothek.io.testing contained lots of unit tests. This is grown historically (support for kartothek-yamal).

This change moves the tests to the tests folder, where one would expect to find them. The new test structure is as follows:

* `tests.io.eager`, `tests.io.iter`, `tests.io.dask.bag`, `tests.io.dask.dataframe`, `tests.io.dask.delayed`: These are tests which are specific for one Kartothek implementation backend. Other than before, the test files contained in these modules will execute only the tests contained directly in the test file, and import no more tests from other packages.
* `tests.io.common`: These are tests which span multiple backend implementations. That is, each one of these tests is executed for more than one backend implementation - however, not all tests make sense for every backend implementation.
The existing backend implementations are controlled using the fixture `backend_identifier`. As it was before, the tests rely on a `bound_xxx` fixture which encapsulates the actual implementation of the functionality to be tested. I.e. a `bound_read` fixture may encapsulate an eager read, a dask dataframe read, etc., depending on the current value of the `backend_identifier` fixture.
If a specific implementation does not exist for a specific backend identifier, the test is skipped using `pytest.skip()` in the `bound_xxx` fixture. This leads in fact to a relatively large number of tests to be skipped, since pytest tries to execute the tests for all possible combinations of backend identifiers.
* `tests.io.cube`: Tests for cube functionalities. These tests contain (as before) a fixture called `driver` to encapsulate the different backend implementations. When a specific functionality is not implemented for one driver, the test is skipped.

BY internal: Refers to JIRA task LDE-714.


xhochy (changes requested):
The also includes the `copy_dataset` PR #450, please remove that or state that this is a requirement for this PR to be reviewed.

MartinHaffner:
> The also includes the `copy_dataset` PR #450, please remove that or state that this is a requirement for this PR to be reviewed.

Added PR 450 as a requirement in the comment. (makes things easier...)

PR #457, opened 2021-04-16
Fix mistakes in documentation

janjagusch:
# Description:

I fixed some mistakes in the documentation:

- In the "Principle in-memory representation" section, the documentation refered to a partition `A=42`, which wasn't introduced before.
- In the "Secondary indices" section, the last sentence wasn't complete. I checked the git log and the file was created like that (https://github.com/JDASoftwareGroup/kartothek/commit/28939c7a56802409f8fbc5f2880796269d8fc84a#diff-5c15d229441c71aed6d9d3f371ab1c6498955d4faa0aa7304e399c1cc7009a2f). Maybe @fjetter still knows what he wanted to write? It's probably something about that the interaction is similar to primary indices. But I didn't want to guess, soI removed it for now.

### DIFF
--- a/docs/spec/indexing.rst
+++ b/docs/spec/indexing.rst
@@ -15,7 +15,7 @@ All currently supported kartothek index types are inverted indices and are mappi
 
     index_dct = {1: ["table/12345"], 2: ["table/12345", "table/6789"]}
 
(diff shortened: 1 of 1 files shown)


stephan-hesselmann-by (approved):
Thanks for the fixes

PR #463, opened 2021-04-26
contribution / Add predicates to conditions converter

infehm:
# Description:

Adding a helper module in utils to convert a kartothek read dataset predicate to a query cube predicate.
This functionalities makes integrations from querying datasets to querying cubes easier, and also will be handy for libraries loading both data from datasets and cubes.

- [ ] Closes #xxxx
- [ ] Changelog entry

### DIFF
--- a/kartothek/utils/predicate_converter.py
+++ b/kartothek/utils/predicate_converter.py
@@ -61,7 +61,7 @@ def write_predicate_as_cube_condition(predicate: Tuple[str, str, Any]) -> Condit
 
 
 def convert_predicates_to_cube_conditions(
-    predicates: List[List],
+    predicates: List[List[Tuple[str, str, Any]]],
 ) -> Sequence[Condition]:
     """
     Converts a io.dask.dataset 'read_dataset_as_ddf' predicate to a cube condition
--- a/kartothek/utils/predicate_converter.py
+++ b/kartothek/utils/predicate_converter.py
@@ -1,0 +1,80 @@
+"""
+Helper module to convert kartothek dataset load predicates into cube conditions.
+"""
+
+from __future__ import absolute_import
+
+import pandas as pd
+import pyarrow as pa
+
+from kartothek.core.cube.conditions import Condition
+
+
+def write_predicate_as_cube_condition(predicate):
+    """
+    Rewrites a single io.dask.dataset 'read_dataset_as_ddf' predicate condition as cube condition
+
+    Parameters
+    ----------
+    predicate: list
+        list containing single predicate definition
+
+    Returns
+    -------
+    condition: cube condition object
+        cube condition containing the predicate definition
+    """
+    condition_string = None
+    parameter_format_dict = {}
+    if type(predicate[2]) == int:
+        condition_string = "{} {} {}".format(
+            predicate[0], predicate[1], str(predicate[2])
+        )
+        parameter_format_dict[predicate[0]] = pa.int16()
+    if type(predicate[2]) == str:
+        condition_string = "{} {} {}".format(predicate[0], predicate[1], predicate[2])
+        parameter_format_dict[predicate[0]] = pa.string()
+    if type(predicate[2]) == pd._libs.tslibs.timestamps.Timestamp:
+        condition_string = "{} {} {}".format(
+            predicate[0], predicate[1], predicate[2].strftime("%Y-%m-%d")
+        )
+        parameter_format_dict[predicate[0]] = pa.timestamp("s")
+    if type(predicate[2]) == bool:
+        condition_string = "{} {} {}".format(
+            predicate[0], predicate[1], str(predicate[2])
+        )
+        parameter_format_dict[predicate[0]] = pa.bool_()
+    if type(predicate[2]) == float:
+        condition_string = "{} {} {}".format(
+            predicate[0], predicate[1], str(predicate[2])
+        )
+        parameter_format_dict[predicate[0]] = pa.float64()
+
+    if condition_string is not None:
+        condition = Condition.from_string(condition_string, parameter_format_dict)
+    else:
+        raise TypeError(
+            "Please only enter predicates for parameter values of the following type:"
+            " str, int, float, bool or pandas timestamp, "
+        )
+    return condition
+
+
+def convert_predicates_to_cube_conditions(predicates):
+    """
+    Converts a io.dask.dataset 'read_dataset_as_ddf' predicate to a cube condition
+
+    Parameters
+    ----------
+    predicates: list
+        list containing a list of single predicates
+
+    Returns
+    -------
+    condition: cube condition object
+        cube condition containing the combined predicate definitions
+    """
+    condition = ()
+    for predicate in predicates[0]:
+        condition = condition + (write_predicate_as_cube_condition(predicate),)
+    return condition
--- a/kartothek/utils/predicate_converter.py
+++ b/kartothek/utils/predicate_converter.py
@@ -2,7 +2,7 @@
 Helper module to convert kartothek dataset load predicates into cube conditions.
 """
 
-from typing import Any, List, Tuple
+from typing import Any, List, Sequence, Tuple
 
 import pandas as pd
 import pyarrow as pa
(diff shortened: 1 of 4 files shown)


kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,80 @@
+"""
+Helper module to convert kartothek dataset load predicates into cube conditions.
+"""
+
+from __future__ import absolute_import
```
xhochy, line 5: Kartothek is Python 3+ only, so this should not be needed.
```suggestion
```

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,80 @@
...
+    condition: cube condition object
+        cube condition containing the predicate definition
+    """
+    condition_string = None
+    parameter_format_dict = {}
+    if type(predicate[2]) == int:
+        condition_string = "{} {} {}".format(
+            predicate[0], predicate[1], str(predicate[2])
+        )
+        parameter_format_dict[predicate[0]] = pa.int16()
```
xhochy, line 33: Could you use `pa.types.is_int*` for this and the following?

xhochy (changes requested):

kartothek/utils/predicate_converter.py
infehm: would that make sense in this context? It should not check the type, but rather define it for the the "to_string" conversion.  
If converting to pa.types.is_int(t), what would I need to pass here as "t"?

kartothek/utils/predicate_converter.py
infehm: changed

kartothek/utils/predicate_converter.py
xhochy: Ah, sorry, I saw a `==` here. Than this is OK but I would be curious why you are defaulting to `int16` and not `int64` here?

infehm: no particular reason for that, I saw this was used in some modules (but 32 and 64 as well). I will change it to int64

xhochy:
Looks OK from my side now but my cube knowledge is limited thus waiting for someone else to pick up review from here. Probably this would also be better placed in `kartothek.core.cube.utils` instead of the global `utils` module.

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,78 @@
...
+Helper module to convert kartothek dataset load predicates into cube conditions.
+"""
+
+import pandas as pd
+import pyarrow as pa
+
+from kartothek.core.cube.conditions import Condition
+
+
+def write_predicate_as_cube_condition(predicate):
```
stephan-hesselmann-by, line 11: Can you add typehints?

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,80 @@
...
+
+from typing import Any, List, Tuple
+
+import pandas as pd
+import pyarrow as pa
+
+from kartothek.core.cube.conditions import Condition
+
+
+def write_predicate_as_cube_condition(predicate: Tuple[str, str, Any]) -> Any:
```
xhochy, line 13:
```suggestion
def write_predicate_as_cube_condition(predicate: Tuple[str, str, Any]) -> Condition:
```

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,80 @@
...
+        condition = Condition.from_string(condition_string, parameter_format_dict)
+    else:
+        raise TypeError(
+            "Please only enter predicates for parameter values of the following type:"
+            " str, int, float, bool or pandas timestamp, "
+        )
+    return condition
+
+
+def convert_predicates_to_cube_conditions(predicates: List[List],) -> Any:
```
xhochy, line 63:
```suggestion
def convert_predicates_to_cube_conditions(predicates: Sequence[Tuple[str, str, Any]]) -> Sequence[Condition]:
```

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Returns
+    -------
+    condition: cube condition object
+        cube condition containing the combined predicate definitions
+    """
+    condition: Any = ()
+    for predicate in predicates[0]:
+        condition = condition + (write_predicate_as_cube_condition(predicate),)
+    return condition
```
xhochy, line 80: Probably simpler as
```suggestion
    return [write_predicate_as_cube_condition(predicate) for predicate in predicates]
```

infehm: This does not work as it looses functionality, but I found a working solution now for the condition type

kartothek/utils/predicate_converter.py
```diff
@@ -60,7 +60,9 @@ def write_predicate_as_cube_condition(predicate: Tuple[str, str, Any]) -> Any:
     return condition
 
 
-def convert_predicates_to_cube_conditions(predicates: List[List],) -> Any:
+def convert_predicates_to_cube_conditions(
+    predicates: List[List],
```
xhochy, line 64: The interesting part here is the typing of the inner list. Reading the docstring I'm confused whether the inner but should be a list as it does contain exactly one element and thus should rather be this one element directly.

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,82 @@
...
+    predicates: list
+        list containing a list of single predicates
+
+    Returns
+    -------
+    condition: cube condition object
+        cube condition containing the combined predicate definitions
+    """
+    condition: Any = ()
+    for predicate in predicates[0]:
```
stephan-hesselmann-by, line 80: Predicates are defined as [[cond1 AND cond2] OR [cond3 AND cond4]]. They are implemented as a List of Lists, where the inner list denotes AND and the outer OR conjunctions.`predicates[0]` would therefore only take the first AND condition block and convert it. Is that intentional? I don't think that is clear from the function definition or docstring. I think you should either implement this by taking only a `List` of AND conditions in the first place (instead of `List[List[...]`), or implement the conversion of all predicates.

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Returns
+    -------
+    condition: cube condition object
+        cube condition containing the predicate definition
+    """
+    condition_string = None
+    parameter_format_dict = {}
+    if type(predicate[2]) == int:
+        condition_string = "{} {} {}".format(
```
stephan-hesselmann-by, line 30: I would suggest to use Python f-strings here to be more in line with the rest of the code base

tests/utils/test_predicate_converter.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+
+from kartothek.core.cube.conditions import C
+from kartothek.utils.predicate_converter import (
+    convert_predicates_to_cube_conditions,
+    write_predicate_as_cube_condition,
+)
+
+
+def test_write_predicate_as_cube_condition():
```
stephan-hesselmann-by, line 11: You can use parametrized tests here to reduce the asserts per test. I would also suggest to separate the `test_raises_when...` into separate tests.

tests/utils/test_predicate_converter.py
```diff
@@ -0,0 +1,65 @@
...
+    ):
+        write_predicate_as_cube_condition(("date", "=="))
+
+    with pytest.raises(
+        ValueError, match="Please use predicates consisting of exactly 3 entries"
+    ):
+        write_predicate_as_cube_condition(("date", "==", "date", "=="))
+
+
+def test_convert_predicates_to_cube_conditions():
```
stephan-hesselmann-by, line 50: Same here

stephan-hesselmann-by (approved):
Thanks Inga looks good now. Will merge when the master has been sorted out

kartothek/utils/predicate_converter.py
```diff
@@ -0,0 +1,88 @@
...
+    and InIntervalCondition and will throw an error if conditions of those types are passed.
+
+    Parameters
+    ----------
+    predicate: list
+        list containing single predicate definition
+
+    Returns
+    -------
+    condition: cube condition object
```
stephan-hesselmann-by, line 27:
```suggestion
    condition: Condition
```

stephan-hesselmann-by: @infehm this is ready to be merged now. Can you add a changelog entry for this change? I think the failing doc build can be fixed the suggestion

infehm: I guess I also need to change it for the below function, right?  (convert_predicates_to_cube_conditions)

Issue #475, opened 2021-06-07
Test failures in the Arrow nightly integration builds


jorisvandenbossche:
We started to see some test failures in the nightly integration tests we run in Arrow (running the kartothek tests against the latest pyarrow). Specifically, the `test_update_dataset_from_ddf_empty` test is failing:

```
=================================== FAILURES ===================================
___________________ test_update_dataset_from_ddf_empty[True] ___________________

store_factory = functools.partial(<function get_store_from_url at 0x7f1434733050>, 'hfs:///tmp/pytest-of-root/pytest-0/test_update_dataset_from_ddf_e0/store')
shuffle = True

    @pytest.mark.parametrize("shuffle", [True, False])
    def test_update_dataset_from_ddf_empty(store_factory, shuffle):
        with pytest.raises(ValueError, match="Cannot store empty datasets"):
            update_dataset_from_ddf(
>               dask.dataframe.from_delayed([], meta=(("a", int),)),
                store_factory,
                dataset_uuid="output_dataset_uuid",
                table="core",
                shuffle=shuffle,
                partition_on=["a"],
            ).compute()

tests/io/dask/dataframe/test_update.py:57: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

dfs = [], meta = (('a', <class 'int'>),), divisions = None
prefix = 'from-delayed', verify_meta = True

    @insert_meta_param_description
    def from_delayed(
        dfs, meta=None, divisions=None, prefix="from-delayed", verify_meta=True
    ):
        """Create Dask DataFrame from many Dask Delayed objects
    
        Parameters
        ----------
        dfs : list of Delayed
            An iterable of ``dask.delayed.Delayed`` objects, such as come from
            ``dask.delayed`` These comprise the individual partitions of the
            resulting dataframe.
        $META
        divisions : tuple, str, optional
            Partition boundaries along the index.
            For tuple, see https://docs.dask.org/en/latest/dataframe-design.html#partitions
            For string 'sorted' will compute the delayed values to find index
            values.  Assumes that the indexes are mutually sorted.
            If None, then won't use index information
        prefix : str, optional
            Prefix to prepend to the keys.
        verify_meta : bool, optional
            If True check that the partitions have consistent metadata, defaults to True.
        """
        from dask.delayed import Delayed
    
[...]

PR #482, opened 2021-06-22
Fix test by matching error messages appearing from dask > 2021.5.0, fix documentation build and prepare release 5.0

steffen-schroeder-by:
# Description:
With #475 we pinned dask do not use 2021.5.1 and 2021.6.0 since they contained known issues (dask/dask#7586).
That issue should have been fixed by https://github.com/dask/dask/pull/7769 which should have been included in >2021.6.0.
However, the the release of dask 2021.6.1, the original issue is gone, but there is another (possibly different) issue.
The message text for empty Dask Data Frames changed in a way that it's dependent on a shuffle operation. See https://github.com/JDASoftwareGroup/kartothek/runs/2880306923?check_suite_focus=true for the exact error messages.

This PR proposes to check for all 3 kind of error message to make sure the test runs with all dask versions and indepedently of the `shuffle` parameter. I decided against checking only for `ValueError` because then we might miss some other `ValueErrors` which actually are bugs.

@fjetter fyi, since this could be of interest  for you because it might be a follow up for https://github.com/dask/dask/pull/7769

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-Version 5.0.0 (2021-05-xx)
+Version 5.0.0 (2021-06-23)
 ==========================
 
 This release rolls all the changes introduced with 4.x back to 3.20.0.
--- a/docs/environment-docs.yml
+++ b/docs/environment-docs.yml
@@ -35,7 +35,7 @@ dependencies:
   - setuptools_scm
 
   # Documentation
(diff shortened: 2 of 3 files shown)


fjetter:
There is an issue with the documentation. it appears master is also broken. no idea where/why this exception message changed but the test is fine for me

stephan-hesselmann-by (approved):

PR #487, opened 2021-07-22
Maintenance/additional tests

martin-haffner-by:
# Description:

Added additional tests to test the behavior of read / update operations with non-standard table names (i.e., tables which are not named "table") within datasets. These tests can be used as regression tests, when the handling of table names changes.

- BY internal: Refers to #LDE-1325

### DIFF
--- a/kartothek/io/testing/read.py
+++ b/kartothek/io/testing/read.py
@@ -638,11 +638,17 @@ def test_read_dataset_alternative_table_name(
         store_factory = None
         ds_factory = dataset_factory_alternative_table_name
 
+    # the table to be read must be passed either as string or list
     if isinstance(bound_load_dataframes, functools.partial):
+        # iter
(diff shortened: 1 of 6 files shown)


kartothek/io/testing/read.py
```diff
@@ -31,6 +31,7 @@
 """
 
 import datetime
+import functools
```
steffen-schroeder-by, line 34: partial is already importet 2 lines below. That appers to be the only usage of this new import

steffen-schroeder-by (approved):
Despite of one small nitpick, this looks good. Thanks for contributing.

kartothek/io/testing/read.py
martin-haffner-by: OK. Changed the test to use `partial` instead of `functools.partial` and removed the import.

PR #489, opened 2021-08-03
Building indices removes user defined metadata

pavankumar-jamanjyothi-by:
# Description:

When building indices for an existing dataset via build_dataset_indices methods, user-defined metadata (the "metadata" key in the deserialized by-metadata.json file) is removed.  The reason is that build_dataset_indices functions pass `load_dataset_metadata=False` to the DatasetFactory, e.g. here:
https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/io/eager.py#L817-L822
This has the effect of actively removing user-defined metadata:
https://github.com/JDASoftwareGroup/kartothek/blob/master/kartothek/core/factory.py#L99-L100
so no metadata is written when the by-metadata.json file is written in the end.

Fix is to pass `load_dataset_metadata=True` to the DatasetFactory.


steffen-schroeder-by (changes requested):
Thanks, all in all, looks good to me. There are 2 things, I'd like to see in addition:
1. this is worth an entry in the changelog
2. we should understand why `load_dataset_metadata` was always set to `False` and what implication it has to set it to `True` as default now. (Functionality/Performance/...). Maybe @fjetter has an idea.

kartothek/io_components/gc.py
```diff
@@ -10,7 +10,7 @@ def dispatch_files_to_gc(dataset_uuid, store_factory, chunk_size, factory):
         dataset_uuid=dataset_uuid,
         store=store_factory,
         factory=factory,
-        load_dataset_metadata=False,
```
pavankumar-jamanjyothi-by, line 13: @jochen-ott-by As this is gc, I think we should set it to `False` here.

jochen-ott-by: I think it does not really matter and we can use `load_dataset_meadata=True` everywhere.

kartothek/io/testing/index.py
```diff
@@ -183,3 +185,34 @@ def test_empty_partitions(store_factory, metadata_version, bound_build_dataset_i
...
+
+def test_build_indices_doesnot_drop_metadata(
+    store_factory, bound_build_dataset_indices
+):
+    dataset_uuid = "dataset_uuid"
+    partitions = [
+        {"label": "cluster_1", "data": [("core", pd.DataFrame({"p": [1, 2]}))]},
+        {"label": "cluster_2", "data": [("core", pd.DataFrame({"p": [2, 3]}))]},
+    ]
+    with freeze_time(TIME_TO_FREEZE_ISO):
```
jochen-ott-by, line 198: IIRC, we had some issues with the `freeze_time` approach in the past, which is why almost no test nowadays uses it. I think this test can be re-written without using `freeze_time`, simply by not checking a value for `metadata["creation_time"]`. This would not only drop the dependency on `freezegun` here, but also make the test clearer.

jochen-ott-by (changes requested):

kartothek/io/testing/index.py
pavankumar-jamanjyothi-by: Makes sense. Removed `freeze_time` and pushed the changes.

PR #491, opened 2021-08-11
Add decorators deprecating individual function parameters, unifying t…

jakob-ernst-by:
Adds deprecation decorators in order to streamline migration efforts.

# Description:

Add decorators, allowing to deprecate individual function parameters in order to facilitate the the current and future migration efforts.

### DIFF
--- a/kartothek/utils/migration_helpers.py
+++ b/kartothek/utils/migration_helpers.py
@@ -1,0 +1,198 @@
+import inspect
+import traceback
+import warnings
+from functools import partial, wraps
+from typing import Callable, Tuple
+
+# This constant string is supposed to be passed as message to the decorators in the current migration effort and to be
+# subsequently removed.
+MULTI_TABLE_FEATURE_DEPRECATION_WARNING = (
+    "The `{parameter}` keyword is deprecated and will be removed in the next major"
+    "release in an effort to remove the multi table feature."
+)
+
+
+def _check_params(func: Callable, params: Tuple[str, ...]) -> None:
+    """
+    Applies validators, verifying the declared decorator parameters.
+
+    Validates:
+
+    #. At least one parameter specified;
+    #. No duplicate param declaration;
+    #. Declared params match underlying function signature.
+
+    Parameters
+    ----------
+    params: Tuple[str, ...]
+        Declared decorator parameters.
+
+    Raises
+    ------
+    ValueError
+        If validation fails.
+    """
+    if len(params) < 1:
+        raise ValueError(
+            "At least one parameter must be specified when using this decorator!"
+        )
+
+    if len(params) > len(set(params)):  # check for duplicates
+        raise ValueError("Duplicate parameter assignment in decorator definition!")
+
+    arg_spec = inspect.getfullargspec(func)
+    if not (
+        arg_spec.varargs or arg_spec.varkw
+    ):  # deactivate check for decorated functions with *args or **kwargs
+        func_args = arg_spec.args
+        if not all([parameter in func_args for parameter in params]):
+            raise ValueError(
+                "Invalid parameter in decorator definition: "
+                + ", ".join(sorted(set(params) - set(func_args)))
+                + "!"
+            )
+
+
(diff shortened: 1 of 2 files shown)


kartothek/utils/migration_helpers.py
```diff
@@ -0,0 +1,196 @@
+import inspect
+import traceback
+import warnings
+from functools import partial, wraps
+from typing import Callable, Tuple
+
+multi_table_feature_deprecation_warning = (
+    "The `{parameter}` keyword is deprecated and will be removed in the next major"
+    "release in an effort to remove the multi table feature."
+)
```
ilia-zaitcev-by, line 10: Do we use this variable somewhere?

tests/utils/test_migration_helpers.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import Tuple
+
+import pytest
+
+from kartothek.utils.migration_helpers import (
+    deprecate_parameters,
+    deprecate_parameters_if_set,
+)
+
+message = "Prameter {parameter} is deprecated!"
```
ilia-zaitcev-by, line 11: A couple of minor changes:
1) typo: `Prameter -> Parameter`
2) should we call it `MESSAGE` instead of `message` as it is a global variable? Otherwise, we could create a fixture for this like:
```python
@pytest.fixture
def message() -> str:
    return "Parameter {parameter} is deprecated!"
    
def test_deprecate_optional_parameter_if_set_multi_table(message: str) -> None:
    ...
```

ilia-zaitcev-by (approved):
@jakob-ernst-by a nice implementation! Should help to reduce the number of warnings scattered here in there in the code.

ilia-zaitcev-by:
Added a couple of notes, but overall looks great!

kartothek/utils/migration_helpers.py
jakob-ernst-by: This variable is supposed to be passed as first decorator parameter for the current migration effort.

It is not very clean without use right now, but I think it is not located too badly once in use.

ilia-zaitcev-by: Ah, ok. I think we could call it `MULTI_TABLE_FEATURE_DEPRECATION_WARNING` to highlight that it is a global constant. Or maybe even better, create a small utility:
```python
def warning_message(parameter: str) -> str:
    return MULTI_TABLE_FEATURE_DEPRECATION_WARNING.format(parameter=parameter)
```

tests/utils/test_migration_helpers.py
jakob-ernst-by: I will capitalize it as a constant, you are right.
We can not use fixtures, because the decorators are constructed when the file is parsed, before the actual function calls are done.

kartothek/utils/migration_helpers.py
ilia-zaitcev-by: I think having a function like that could better communicate the rationale why this variable exists to a user.

tests/utils/test_migration_helpers.py
ilia-zaitcev-by: Yeah, right, the decorates created in the test file itself.

kartothek/utils/migration_helpers.py
jakob-ernst-by: I will however capitalize it as constant.

tests/utils/test_migration_helpers.py
jakob-ernst-by: (FYI: This is also the reason, why I test the error cases with nested functions)

kartothek/utils/migration_helpers.py
jakob-ernst-by: Sidenote because I did not address the second option: The proposed small utility is unsuitable because the decorator has to keep the authority on what parameter to raise a warning with and this selection is dynamic, based on the individual function calls for the deprecate_parameters_if_set decorator.

PR #531, opened 2022-11-04
Move to minimalkv

timostrunk:
# Description:
Some time ago in #462 it was discussed that a move to minimalkv is not required as repo access to simplekv will be transfered. As of today the simplekv repository is dead. This is therefore another attempt to move kartothek to minimalkv. I recreated the previous PR as the codebase has changed a bit since the last PR.

Still, this PR should also be credited to @xhochy , because he basically did the same work some time ago.

Functionally nothing should change as minimalkv stores and storefact api are unchanged between the two packages.

Consumers of kartothek will be able to have both minimalkv and storefact and simplekv in the same python environment and use them interchangeably. This PR allows those projects to switch to minimalkv and abandon simplekv gracefully.

The two additional commits part of this PR should be reviewed very carefully: I added them to fix unit tests before starting development. Especially the added type compatibility in serialization/_generic should be reviewed very carefully.

- [ ] Closes #462
- [x] Changelog entry
- [ ] Somebody not @timostrunk checked type compatibility in serialization/_generic
- [ ] Somebody not @timostrunk checked the metadata fix in test_dataset_dyn_part
- [ ] Depends on #532

### DIFF
--- a/.github/workflows/ci-pre-commit.yml
+++ b/.github/workflows/ci-pre-commit.yml
@@ -6,6 +6,8 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
-    - uses: pre-commit/action@v2.0.3
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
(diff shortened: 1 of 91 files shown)


xhochy:
We implemented this over in https://github.com/data-engineering-collective/plateau/pull/51 as we didn't get anyone to respond here anymore. Happy to collaborate with you there.

PR #532, opened 2022-11-04
Fix GitHub actions

timostrunk:
# Description:

PR #531 showed that all github actions (CI, docs and pre-commit) are currently broken.
* Docs are broken, because the environment is pulling in a broken jinja version incompatible with the ancient sphinx version
  Solution -> switch to a mamba action and pin jinja
* pre-commit is broken out of many reasons, mainly due to the conda setup
  Solution -> update github actions, run isort with black profile, pray
* CI is broken because of python 3.7 and old pyarrow versions.
  Solution -> Remove python 3.7 and old pyarrow tests
The unit test changes are also included in this PR. I will rebase #531, once this is merged to get the actions green in #531.

This PR is more a bandaid than anything else. A good fix would be to rebuild the github actions, update sphinx, update dependabot and update all environments.

Todo:
- [ ] Check the two commits actually changing funtionality
- [ ] Confirm that dropping Py3.7 and old pyarrow from CI is ok (was broken before anyways)


timostrunk:
Still fighting with the CI tests. After removing every CI test with pyarrow < 2.0.0, failing tests are pyarrow==5.0.0 and pyarrow==2.0.0 and numfocus nightly.

